PR #3, opened 2015-04-02
Support for monotonic clock

mbodmer:
Hi Trent
Here is my patch as pull request. Maybe this helps as starting point.
Cheers
Marc

### DIFF
--- a/avahi-autoipd/main.c
+++ b/avahi-autoipd/main.c
@@ -791,7 +791,7 @@ int is_ll_address(uint32_t addr) {
 static struct timeval *elapse_time(struct timeval *tv, unsigned msec, unsigned jitter) {
     assert(tv);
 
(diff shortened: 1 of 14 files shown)


lathiat:
Thanks for the pull request, I think before we merge this that we need to check clock usage in a few other places.

In particular the gobject code uses in some places, the gobject functions to get time interchangeable with the avahi ones, and this may cause issues.

So I think we need to audit all time sources as part of this.

lathiat:
I wonder if CLOCK_BOOTTIME would be a better source where available, to ensure things expire correctly after resume from suspend.

Particularly since that would happen right now, but will not happen with monotonic time.

mbodmer:
This PR is superseeded by PR #96 in the meantime ... will close it now

PR #11, opened 2015-10-10
Fix stuck registering

tmds:
Decrement n_probing when removing probing announcer so group doesn't get stuck registering because n_probing can no longer decrement to zero.

Fixes: http://avahi.org/ticket/201, https://github.com/lathiat/avahi/issues/7
Patch was sent to avahi mailing list but was not yet applied: http://lists.freedesktop.org/archives/avahi/2014-March/002291.html

### DIFF
--- a/avahi-core/announce.c
+++ b/avahi-core/announce.c
@@ -35,8 +35,14 @@
 #define AVAHI_PROBE_INTERVAL_MSEC 250
 
 static void remove_announcer(AvahiServer *s, AvahiAnnouncer *a) {
+    AvahiEntry *e;
+
     assert(s);
     assert(a);
+    e = a->entry;
+
+    if (a->state == AVAHI_PROBING && e->group)
+        e->group->n_probing--;
 
     if (a->time_event)
         avahi_time_event_free(a->time_event);


ChrisLesiak:
I have a system that suffers from the "stuck registering" problem.  With 0.6.32-rc , the problem occurs on more than 10% of warm boots.  With this commit, I haven't seen the problem after 260 warm boots.

chhitz:
We have this patch applied in our embedded system since January and it solved the locking without any side effects.

lathiat:
Thanks for the feedback, I really appreciate it.  I will likely merge this patch for 0.32

Batchyx:
What if n_probing is decremented to 0 ? I don't think someone will ever call avahi_s_entry_group_check_probed() on the entry group afterward.

My version of the patch handles this case: https://github.com/Batchyx/avahi/commit/cb15988c4c676410549ef1d04487107cd484b353

tmds:
@Batchyx I can't confirm the call is needed or not. After applying this patch 20 months ago, I never had any issues finding devices via avahi, but that doesn't mean all cases are covered. Did you have a case where you got stuck registering that was fixed by the call? Or do you have a good understanding of how the code is working? On the latter point, I can say that my understanding of the code is limited. The fix I made was by looking at how the registration got stuck rather than trying to understand how the code works.

Batchyx:
@tmds I figured that by understanding the code. There is no magic that makes something happen when n_probing becomes zero, one has to check for it. And that check only happen in avahi_s_entry_group_check_probed() which is called at two places:
- next_state(), right after decrementing n_probing. next_state() advances the announcer state machine, which is called only when needed (e.g. when a probing or announce timer expires)
- entry_group_commit_real(), right after creating an announcer for each interface. Because if there are no usable interfaces, no announcers are created and n_probing is already zero.

So it all depends on your specific setup. i think that starting avahi with one interface and destroying it during the probe should be enough to trigger the bug again. Maybe i will try it soon.

Batchyx:
I have successfully reproduced the problem with your patch applied, by using one interface just like i described above. I'm creating a pull request with my patch.

tmds:
ok, great! Thanks @Batchyx

ChrisLesiak:
I appear to be have a problem with the patch from @Batchyx applied to avahi-0.6.32-rc.

I have two networks, one has a dhcp server and the other is zeroconf only.

After swapping between the two networks, I will find that the avahi-daemon process has exited.  After this, I can switch back and forth between the two networks and the combination of netplugd, avahi-autoipd, and udhcpc will still be getting the addressing right, but avahi-daemon will not be running.

The problem doesn't always occur, but I can trigger it after a few tries.

It appears that is happens only when swapping to the network with the dhcp server.

Batchyx:
Do you have a log output from Avahi when it happens ? or, at least, the output of "ip monitor" while you swap the networks.

ChrisLesiak:
On 01/25/2016 06:40 AM, Batchyx wrote:
>
> Do you have a log output from Avahi when it happens ? or, at least, 
> the output of "ip monitor" while you swap the networks.
>
>

Here is the last "cycle" of network swaps.  avahi-daemon was running at 
the start, as well as half-way through with the 169.254.5.240 IPv4 
address, but not at the end.


172.24.40.4 dev eth0 lladdr 14:fe:b5:26:32:a1 REACHABLE
Deleted 2: eth0    inet 172.24.80.205/16 brd 172.24.255.255 scope global 
eth0
        valid_lft forever preferred_lft forever
Deleted 172.24.0.0/16 dev eth0  proto kernel  scope link  src 172.24.80.205
Deleted broadcast 172.24.255.255 dev eth0  table local  proto kernel  
scope link  src 172.24.80.205
Deleted broadcast 172.24.0.0 dev eth0  table local  proto kernel scope 
link  src 172.24.80.205
Deleted local 172.24.80.205 dev eth0  table local  proto kernel scope 
host  src 172.24.80.205
delete 172.24.40.4 dev eth0 lladdr 14:fe:b5:26:32:a1 REACHABLE
delete 224.0.0.251 dev eth0 lladdr 01:00:5e:00:00:fb NOARP
delete 224.0.0.22 dev eth0 lladdr 01:00:5e:00:00:16 NOARP
Deleted 2: eth0    inet6 fe80::201:29ff:fe20:ce34/64 scope link
        valid_lft forever preferred_lft forever
Deleted local fe80::201:29ff:fe20:ce34 dev lo  proto none  metric 0
Deleted fe80::/64 dev eth0  proto kernel  metric 256
2: eth0: <BROADCAST,MULTICAST> mtu 1500 qdisc pfifo_fast state DOWN 
group default
     link/ether 00:01:29:20:ce:34 brd ff:ff:ff:ff:ff:ff
delete 224.0.0.22 dev eth0 lladdr 01:00:5e:00:00:16 NOARP
delete ff02::1:ff20:ce34 dev eth0 lladdr 33:33:ff:20:ce:34 NOARP
delete ff02::2 dev eth0 lladdr 33:33:00:00:00:02 NOARP
delete ff02::fb dev eth0 lladdr 33:33:00:00:00:fb NOARP
delete ff02::16 dev eth0 lladdr 33:33:00:00:00:16 NOARP
Deleted ff00::/8 dev eth0  metric 256
2: eth0: <BROADCAST,MULTICAST,UP> mtu 1500 qdisc pfifo_fast state UP 
group default
     link/ether 00:01:29:20:ce:34 brd ff:ff:ff:ff:ff:ff
2: eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc pfifo_fast 
state DOWN group default
     link/ether 00:01:29:20:ce:34 brd ff:ff:ff:ff:ff:ff
ff00::/8 dev eth0  metric 256
fe80::/64 dev eth0  proto kernel  metric 256
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast 
state UP group default
     link/ether 00:01:29:20:ce:34 brd ff:ff:ff:ff:ff:ff
2: eth0    inet6 fe80::201:29ff:fe20:ce34/64 scope link
        valid_lft forever preferred_lft forever
local fe80::201:29ff:fe20:ce34 dev lo  proto none  metric 0
2: eth0    inet 169.254.5.240/16 brd 169.254.255.255 scope link eth0:avahi
        valid_lft forever preferred_lft forever
local 169.254.5.240 dev eth0  table local  proto kernel  scope host src 
169.254.5.240
broadcast 169.254.255.255 dev eth0  table local  proto kernel  scope 
link  src 169.254.5.240
169.254.0.0/16 dev eth0  proto kernel  scope link  src 169.254.5.240
broadcast 169.254.0.0 dev eth0  table local  proto kernel  scope link  
src 169.254.5.240
default dev eth0  scope link  metric 1002
Deleted default dev eth0  scope link  metric 1002
Deleted 2: eth0    inet 169.254.5.240/16 brd 169.254.255.255 scope link 
eth0:avahi
        valid_lft forever preferred_lft forever
Deleted 169.254.0.0/16 dev eth0  proto kernel  scope link  src 169.254.5.240
Deleted broadcast 169.254.255.255 dev eth0  table local  proto kernel  
scope link  src 169.254.5.240
Deleted broadcast 169.254.0.0 dev eth0  table local  proto kernel scope 
link  src 169.254.5.240
Deleted local 169.254.5.240 dev eth0  table local  proto kernel scope 
host  src 169.254.5.240
2: eth0    inet 172.24.80.205/16 brd 172.24.255.255 scope global eth0
        valid_lft forever preferred_lft forever
local 172.24.80.205 dev eth0  table local  proto kernel  scope host src 
172.24.80.205
broadcast 172.24.255.255 dev eth0  table local  proto kernel  scope 
link  src 172.24.80.205
172.24.0.0/16 dev eth0  proto kernel  scope link  src 172.24.80.205
broadcast 172.24.0.0 dev eth0  table local  proto kernel  scope link  
src 172.24.80.205
default via 172.24.1.1 dev eth0
172.24.41.38 dev eth0 lladdr bc:ae:c5:07:02:fd STALE
172.24.41.38 dev eth0 lladdr bc:ae:c5:07:02:fd REACHABLE

tmds:
Guys, can we get this moving again?

These are the observations (correct me if I am wrong):
- patch provided in this PR is working fine for me and @ChrisLesiak 
- patch at https://github.com/Batchyx/avahi/commit/cb15988c4c676410549ef1d04487107cd484b353 which also handles the case where n_probing becomes 0, is not working fine for @ChrisLesiak 

As far as I am concerned, we can merge this PR. An additional PR can be sent by @Batchyx to improve further. What do you think?

tmds:
*bump*

Batchyx:
I can't reproduce @ChrisLesiak's crashes in my setup.  Having the output of avahi before the crash would help.  The output of ip monitor does not tell me where avahi crashes.

ChrisLesiak:
I've looked at this again.  I am still able to reproduce the stuck registering problem with 0.6.32.  Applying [Batchyx@cb15988](https://github.com/Batchyx/avahi/commit/cb15988c4c676410549ef1d04487107cd484b353) does fix that.

But I am unable to reproduce the problem I reported on Jan 8.  I'm starting to think it may have been unrelated to the patch.

lathiat:
Thanks for the feedback and testing, much appreciated!

I'll check over the patch and comments and run through it logically to make sure it makes sense.

neophob:
+1

pointsphere:
While I don't understand the code, I can confirm that building v0.6.32 with this PR now makes my Raspberry Pi register reliably. Before that it often got stuck on registering.

jitendrajagasia:
Hello lathiat,

Is the fix valid as per you? & in which avahi version will it be available

Thanks
Jitendra

tmds:
Closing due to age.

PR #13, opened 2015-10-10
update translations

rbuj:
mkdir .tx
cat << EOF > .tx/config
[main]
host = https://www.transifex.com

[avahi.master-tx-po-avahi-pot]
file_filter = po/<lang>.po
source_file = po/avahi.pot
source_lang = en
EOF
tx pull -fas
sed -i 's/An unexpected D-Bus error occured/An unexpected D-Bus error occurred/g' po/*.po po/*.pot
cd po
ls | sed -n 's/\.po$//p' > LINGUAS
intltool-update --pot --gettext-package=avahi
for i in `cat LINGUAS`; do intltool-update --dist $i; done


piotrdrag:
@rbuj This is great! Thanks for doing this.

@lathiat are you planning on using the old Transifex project again to handle translations (https://www.transifex.com/lennart/avahi/), or https://translate.zanata.org, or some other service, or would you prefer getting pull requests? I'd like to update https://l10n.gnome.org/module/avahi/ with proper instructions.

lathiat:
Thanks for the script, and PR.  Truthfully I am not very familiar with localisation workflows.. do you have any advice, is transifex still OK or is there a better choice?

I think Launchpad also has a bunch of stuff in there from Ubuntu translations.

I will merge your PR for now but I welcome input into the process going forward.

rbuj:
I read in somewhere [1] that Fedora Project uses Zanata [2] now because Transiflex decided to use closed source.

[1] https://fedoraproject.org/wiki/L10N_Move_To_Zanata

[2] https://fedora.zanata.org/

piotrdrag:
Transifex is now closed source on the server side, but most people are OK with that. Fedora's problem was more with self-sustainability, which doesn't apply here. I think using distro-specific services like Launchpad or fedora.zanata limits potential translators. Many freedesktop.org projects use Transifex, and I'd say it's still a good fit for Avahi.

rbuj:
I still remember when OpenSolaris community (NetBeans IDE, openSolaris OS, OpenDS, ...) which used openCTI platform for managing their translations, and openCTI decided to use some closed components. I wish this never happens again. In this regard I want to really convey to freedesktop.org my best wishes.

PR #32, opened 2015-10-20
initscript: Remove obsolete archlinux scripts

heftig:
If Arch Linux is detected, these should not be installed. The
systemd detection installs the right files.


lathiat:
Thanks for your contribution.

PR #36, opened 2015-11-13
Update avahi-discover code to Gtk 3 and Python 3

takluyver:
The code can still run with Python 2 as well.

I didn't work out all the autotools stuff, so I just tweaked the `@substitution@` bits by hand so I could run the code to test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ script: ./bootstrap.sh && make -j3
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get build-dep avahi avahi-sharp
+  - sudo apt-get install python-gi gir1.2-gtk-3.0
   - sudo apt-get install mono-mcs
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ script: ./bootstrap.sh && make -j3
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get build-dep avahi avahi-sharp
-  - sudo apt-get install python-gi gir1.2-gtk-3.0
+  - sudo apt-get install python-gi-dev gir1.2-gtk-3.0
   - sudo apt-get install mono-mcs
--- a/avahi-python/avahi-discover/Makefile.am
+++ b/avahi-python/avahi-discover/Makefile.am
@@ -24,7 +24,7 @@ EXTRA_DIST = \
 
 if HAVE_PYTHON
 if HAVE_PYTHON_DBUS
-if HAVE_PYGTK
+if HAVE_PYGOBJECT
 
 pythonscripts =
 
--- a/avahi-python/avahi-discover/avahi-discover.py
+++ b/avahi-python/avahi-discover/avahi-discover.py
@@ -17,18 +17,21 @@
 # Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
 # USA.
 
+from __future__ import print_function
+
 import os, sys
 
 try:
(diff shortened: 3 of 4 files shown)


lathiat:
Thanks so much for the pull request and work! I'll review this and get back to you :-) Look forward to getting it merged.

takluyver:
You're welcome. I was browsing through the [Fedora Python 3 porting database](http://portingdb-encukou.rhcloud.com/), and since it's alphabetical, avahi was near the top. :-)

takluyver:
Ping - I think this might have dropped off the radar ;-)

goetzc:
@lathiat can this PR get merged?

andhe:
Pushed a couple of extra fixes to https://github.com/andhe/avahi/commits/pygobject on top of this PR.

takluyver:
Looks good to me, thanks. Unfortunately, I can't easily pull it into my branch for this PR, because it looks like you've redone my commits rather than starting from my branch. :-S Maybe it's easiest to open a new PR with your branch and we can close this one?

andhe:
@takluyver oh sorry. I did 'git am' to apply your pr downloaded as a patch from here rather than fork your branch. If the shas don't match up perfectly you should still simply be able to pull in my commits by adding my repo as a remote and then cherry-picking my commits into your branch.

> git remote add andhe https://github.com/andhe/avahi.git
> git fetch andhe
> git cherry-pick f780c0fcaaf90512bda15027eb6e4fefc32f6ce0
> git cherry-pick b0b953ba21b03b8945532dd11e59a39b373a80ad
> git cherry-pick edbe7de84d7573836e66f7b5e7259cd226bf78d7

(Not sure why/if a simple git merge wouldn't do the right thing though.)

takluyver:
I've got around to integrating @andhe's changes, and fixing the Travis builds to work with the new autoconf stuff.

lathiat:
This looks great, but I found one main issue:
 (1) I can't Control-C the program from the terminal and hitting the close button doesn't actually exit.  This does work with the current version.

As an aside a couple of things it'd be great to also improve are:
 (2) I think we need to somehow update avahi-python itself to install for both python2 and python3 - I'm not sure how to best handle that do you have any idea?
 (3) I merged a patch to import gnu.dbm in avahi.ServiceTypeDatabase; this should probably be revised though due to the patching of .firstkey() etc from Makefile.am.  Happy to patch that in a separate PR.

takluyver:
1. I found [a workaround](http://stackoverflow.com/a/16486080/434217) for the issue with Ctrl-C, and I've pushed to this branch.
2. I agree that it would be good to make avahi-python available for both Python versions, but I haven't sussed out the autotools build machinery here. I'm still editing the `@subsitution@` bits out of the Python file by hand so I can test it. So I probably can't help you with this!
3. Sorry, I don't follow how that's related to this PR?

lathiat:
With regards to (3), avahi-discover improts avahi.ServiceTypeDatabase which requires gdbm and that doesn't work in python3.  Surprised you could test the tool without making the same change, guessing you linked python3 to python in which case the existing code to detect and use 'dbm' (for solaris) instead might work.  In any case it's fixed enough for now, I mostly mentioned it since when I first wrote (3) I hadn't pushed that change but decided to do that while investigating it.

Thanks for the fixes!

takluyver:
Aha, gotcha. Yes, I wrote a `gdbm.py` containing `from dbm.gnu import *`, which was enough to let me test the script. In Python code, I'd normally do the checks and fallbacks at runtime, because there may be different libraries available at runtime to build time.

PR #43, opened 2016-01-27
Do not create a directory below /usr/share which does not have content

DimStar77:
pkgdata_DATA is never used - it's pkglibdata_DATA in the other cases


DimStar77:
ping

PR #44, opened 2016-01-27
avahi-sharp: multi arch distributions have mono stuff generally not i…

DimStar77:
…n libdir

On typical multiarch systems, libdir maps to /usr/lib64, the entire mono stack (*.dll) though
is not 64 bit and all mono assemblies are installed in /usr/lib/...

### DIFF
--- a/avahi-sharp/Makefile.am
+++ b/avahi-sharp/Makefile.am
@@ -73,10 +73,10 @@ monodoc_DATA = avahi-sharp-docs.zip avahi-sharp-docs.tree avahi-sharp-docs.sourc
 endif
 
 install-data-hook: $(ASSEMBLY)
-	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /i $(ASSEMBLY) /package avahi-sharp /gacdir $(libdir) /root $(DESTDIR)$(libdir)
+	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /i $(ASSEMBLY) /package avahi-sharp /gacdir $(prefix)/lib /root $(DESTDIR)$(prefix)/lib
 
 uninstall-hook: $(ASSEMBLY)
-	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /u avahi-sharp /package avahi-sharp /gacdir $(libdir) /root $(DESTDIR)$(libdir)
+	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /u avahi-sharp /package avahi-sharp /gacdir $(prefix)/lib /root $(DESTDIR)$(prefix)/lib
 
 endif
 endif
--- a/avahi-sharp.pc.in
+++ b/avahi-sharp.pc.in
@@ -1,6 +1,6 @@
 prefix=@prefix@
 exec_prefix=@prefix@
-libdir=@libdir@
+libdir=@prefix@/lib
 
 Name: avahi-sharp
 Description: Mono bindings for the Avahi mDNS/DNS-SD stack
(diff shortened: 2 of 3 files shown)


DimStar77:
ping

DimStar77:
ping ?

DimStar77:
ping?

DimStar77:
ping?

DimStar77:
ping?

avahi-sharp/Makefile.am
```diff
@@ -73,10 +73,10 @@ monodoc_DATA = avahi-sharp-docs.zip avahi-sharp-docs.tree avahi-sharp-docs.sourc
 endif
 
 install-data-hook: $(ASSEMBLY)
-	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /i $(ASSEMBLY) /package avahi-sharp /gacdir $(libdir) /root $(DESTDIR)$(libdir)
+	$(AM_V_GEN)MONO_SHARED_DIR=. $(GACUTIL) /i $(ASSEMBLY) /package avahi-sharp /gacdir $(prefix)/lib /root $(DESTDIR)$(prefix)/lib
```
pemensik, line 76: Is this path the same on Debian/Ubuntu and Fedora/RHEL families? Wouldn't configure time flag to override this path make more sense?

pemensik: I would use variable defined at the top, ``sharplib=$(prefix)/lib, then use it multiple times. If anyone needs to override it, it would make it easier.

pemensik: Verified both Fedora and Debian use the same path, it should work for both. I think it is reasonable to expect it would work for most other distributions as well.

pemensik (approved):

evverx:
Looks like Fedora and SUSE are OK with this. @mbiebl I wonder if Debian is OK with this?

(Generally this part of avahi is effectively unmaintained so I'm not sure it should be moved around any more)

mbiebl:
hard-coding the directory feels wrong. With `--libdir` you at least have a chance to configure the directory according to your distro policy.
That said, I'm not really familiar with mono and the Debian avahi package no longer builds the mono bindings.

avahi-sharp/Makefile.am
mbiebl: How did you check that? The Debian avahi package does not build the mono bindings?

mbiebl:
I *assume* the motivation for the PR is, that `$libdir` is currently used within avahi for both its C libraries and mono libraries/bindings and those directories are not necessarily related. So, a dedicated `$monolibdir, --with-monolibdir` would make some sense to me.
Fwiw, attached is the output of `apt-file search /usr/lib | grep -E "\.dll$"`
[mono.txt](https://github.com/lathiat/avahi/files/13170863/mono.txt)
It appears, most packages (on Debian) install to /usr/lib/cli, some to /usr/lib/$pkg, mono itself seems to use /usr/lib/mono and very few actually do you multiarch diretories.

mbiebl:
One more/last note: It appears the mono eco system in Debian is not in a very healthy state:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010445
So I'm not sure, who I could ask more about packaging of mono stuff in Debian.

evverx:
> I assume the motivation for the PR is, that $libdir is currently used within avahi for both its C libraries and mono libraries/bindings and those directories are not necessarily related. So, a dedicated $monolibdir, --with-monolibdir would make some sense to me.

Looking at https://src.fedoraproject.org/rpms/avahi/blob/rawhide/f/avahi-0.6.30-mono-libdir.patch (where `libdir` is replaced with `$(prefix)/lib`) it seems `--with-monolibdir` would be the right thing to do.

> So I'm not sure, who I could ask more about packaging of mono stuff in Debian.

I just wanted to make sure that this PR doesn't break the Debian package or cause any unnecessary packaging issues there. Looks like it doesn't so I guess it should be left to distributions where it matters (as long as they don't break each other's packages with their patches).

evverx:
I think it can be closed. I think this whole thing should be frozen but other than that I agree with @mbiebl that it shouldn't be hard-coded.

Looking at https://build.opensuse.org/request/show/956509 it seems openSUSE no longer ships this stuff.

@DimStar77 if you're involved in packaging avahi on openSUSE it would be great if you could take a look at https://github.com/lathiat/avahi/issues/516.

PR #45, opened 2016-01-27
g-i: reference the build shared object from the .gir file

DimStar77:
openSUSE uses a dependency extractor for g-i typelibs that lists which shared libraries are referenced.

It was observed if the linking is not done the right way, generally we end up with weird dependencies (which g-i internally seems to somehow guess until it finds something that works).

Putting the right reference in place makes those things work (mageia uses the same extractor, fedora was discussing it)


DimStar77:
ping

lathiat:
Thanks!

PR #46, opened 2016-01-27
Move to /run

DimStar77:
on modern, systemd based systems, the default directory is /run, no longer /var/run


DimStar77:
ping

dovchinnikov:
Trying to `avahi-daemon` installed from Entware:
```
$ avahi-daemon
mkdir("/run/avahi-daemon/"): No such file or directory
```
I cannot create `/run` directory, because `/` is mounted in read only mode by router OS.

PR #51, opened 2016-03-06
configure.ac: Fix libavahi-client test

heftig:
This only affected the report at the end of configure.


lathiat:
Good spot, thanks.

PR #56, opened 2016-05-24
Include <poll.h> instead of <sys/poll.h>

ncopa:
Both POSIX and the manpage for poll(2) says that `poll.h` should be
included, not `sys/poll.h`

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/poll.h.html
http://man7.org/linux/man-pages/man2/poll.2.html

This fixes compile warning when building with musl libc.

### DIFF
--- a/avahi-common/simple-watch.c
+++ b/avahi-common/simple-watch.c
@@ -21,7 +21,7 @@
 #include <config.h>
 #endif
 
-#include <sys/poll.h>
+#include <poll.h>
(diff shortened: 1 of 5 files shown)


lathiat:
Thanks for your contribution, I just merged the same fix not realizing it was a duplicate:
https://github.com/lathiat/avahi/pull/58

Thus this is effectively merged, but will not merge this actual request.

PR #58, opened 2016-06-18
Fix warnings on musl libc

somasis:
poll.h is defined in POSIX and has existed since at least SUSv2.
Usage of sys/poll.h is a glibc-ism and produces warnings on musl, even
though they are the same header; the only difference is that poll.h
is guaranteed portable due to its standardization.

### DIFF
--- a/avahi-common/simple-watch.c
+++ b/avahi-common/simple-watch.c
@@ -21,7 +21,7 @@
 #include <config.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


joerg-krause:
Furthermore, `sys/poll.h` is silently redirected to `poll.h` on Glibc.

lathiat:
Thanks for your contribution!

PR #59, opened 2016-06-23
avahi-daemon.service.in: start after network-online

mgorse:
We are seeing this in a log while beta testing SLE 12 SP2
May 31 11:16:35 linux avahi-daemon[801]: Failed to open /etc/resolv.conf: No such file or directory

So it appears that avahi-daemon wants resolv.conf to be set up correctly; maybe it makes sense to ensure that NetworkManager or wicked have started first.

### DIFF
--- a/avahi-daemon/avahi-daemon.service.in
+++ b/avahi-daemon/avahi-daemon.service.in
@@ -18,6 +18,8 @@
 [Unit]
 Description=Avahi mDNS/DNS-SD Stack
 Requires=avahi-daemon.socket
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=dbus


lathiat:
Thanks for your contribution.

This functionality is used mainly for the seldom-used publish-dns-servers, publish-resolv-conf-dns-servers and publish-domain options which advertise your DNS services on the LAN for avahi-dnsconfd to consume.  These options are also re-read when reload config is called, which can also happen on inotify - though we currently don't inotify on resolv.conf (and I'm not sure how that would work on the non existant file anyway)

avahi-dnsconfd and these options have long outlived their life, they have since been replaced by extensions to the radvd protocol and optionally DHCPv6.  I am not really aware of people actually using them in production and I do not believe any other implementation has implemented using these.    They will likely be removed in some future release where we have a good reason to break backwards compatibility at the same time.

From my understanding of systemd units, it appears that the "Wants" and "After" combination makes it a total loose coupling, so this wouldn't cause any problems but it also seems an unnecessary dependency.  Instead of this change, I propose that we squelch this resolv.conf error unless the options that use it are actually configured.

Do you see any problem with that?

mgorse:
Thanks for the review. I think that what you are proposing would be fine.

PR #63, opened 2016-07-11
Only warn on missing resolv.conf if the options that use it are enabled

mgorse:
If /etc/resolv.conf is missing (possibly because avahi started before wicked/NetworkManager), then we get a warning, but usually this warning is harmless.

Follow-up to https://github.com/lathiat/avahi/pull/59

### DIFF
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -172,7 +172,11 @@ static int load_resolv_conf(void) {
 #endif
 
     if (!f) {
-        avahi_log_warn("Failed to open "RESOLV_CONF": %s", strerror(errno));
+        if ((config.publish_dns_servers && config.publish_dns_servers[0]) ||
+             config.publish_resolv_conf ||
+             config.server_config.publish_domain) {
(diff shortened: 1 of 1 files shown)


rojkov:
If chroot is enabled then the log contains

```
Jul 29 13:03:23 intel-corei7-64 avahi-daemon[628]: ../../avahi-0.6.32/avahi-daemon/chroot.c: open() failed: No such file or directory
Jul 29 13:03:23 intel-corei7-64 avahi-daemon[624]: Failed to open /etc/resolv.conf: Invalid argument
```

This patch removes only the second line from the logs.

lathiat:
As a review for this merge generally, we only need to check config.publish_resolv_conf (publish_domain and publish_dns_servers do not rely on the resolv.conf load).

However, I realised the further problem that resolv.conf is used for the wide area server list as well.  For that reason, the only real fix for this is to monitor and reload that file as it changes.  I have created #118 to track that work and will not merge this even though wide area is a low usage feature at the moment.

Thanks for your contribution and apologies I am not able to merge it immediately.

Issue #64, opened 2016-07-25
Become definitive upstream for nss-mdns


agoode:
For context, see
* https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=211345
* https://bugs.chromium.org/p/chromium/issues/detail?id=626377

I would like to propose merging the nss-mdns package into this one, making avahi the definitive upstream for that package. nss-mdns has a hard requirement on Avahi in its recommended configuration, and there are outstanding patches from FreeBSD that have not been accepted upstream.

The existing upstream is at http://0pointer.de/lennart/projects/nss-mdns/ and http://git.0pointer.net/nss-mdns.git/ but has received no updates since 2007.

I can prepare the patches if this idea is acceptable. My further plans include:
* unifdef many of the compile options, specifically the non-Avahi support which has been long deprecated
* fix up and merge the outstanding FreeBSD ports patches
* then finally look into adding some new functionality to nss-mdns

My ultimate goal here is to allow Avahi + nss-mdns to coexist with DNS servers that provide .local resolution, using the SOA algorithm described here: https://support.apple.com/en-us/HT201275. This would allow the elimination of the problematic scripts currently used by distributions that hard-stop Avahi when such a configuration is detected.

lathiat:
+1 this sounds great. I also have code to support interface indexes/scopes so link local IPv6 will work properly which means IPv6 can be enabled by default and generally work properly.

I suspect However I also need to pay attention to systemd-resolved. Seems they implemented a mini mdns stack for some strange reason already but the good news is that Avahi support could use dbus rather than the UNIX socket and thus also easy to add interface index/scopes for that.

I'll set it up and happily take your patches.

agoode:
Yes, I looked at systemd-resolved and thought "one problem at a time" 😺

agoode:
@lathiat Can you take a look at the pull request and let me know if it's what you're looking for?

agoode:
See https://github.com/lathiat/nss-mdns.

lemenkov:
Cool! It's great to see that nss-mdns receives an attention after a long hiatus! Thanks @agoode !

PR #65, opened 2016-07-25
Merge entire history of nss-mdns into a subdirectory of this repository

agoode:
Commands run:
```sh
git clone git://git.0pointer.net/nss-mdns.git
git clone https://github.com/lathiat/avahi.git
cd nss-mdns
git filter-branch --index-filter 'git ls-files -s |
  sed "s-\t\"*-&nss\-mdns/-" |
  GIT_INDEX_FILE=$GIT_INDEX_FILE.new git update-index --index-info &&
  mv "$GIT_INDEX_FILE.new" "$GIT_INDEX_FILE" ||:' HEAD
cd ../avahi
git remote add nss ../nss-mdns
git remote update
git merge nss/master

```

For issue #64.


agoode:
@lathiat Any thoughts on this? I have a few more pull requests queued up (including autoconf merges) but want to make sure this is the right approach.

lathiat:
Apologies for the delay.  I put some thought into it and I think we are better maintaining this as a separate repository.  However I would like to take over maintenance and take your contributions.

The reasons I feel this is the best course of action is:
 (1) There's no real benefit to combining them, nss-mdns is likely to change less often and tying it to avahi releases is unlikely to have much advantage
 (2) It will create headache for the distro packagers to try and combine  them into the same source package -- and equally update another package unnecessarily with each update
 (3) It's likely to be less relevant in many use-cases with systemd-resolved

I'm still open to discussion on this part if you strongly believe this is the wrong way to go, then please go ahead and comment as such.  Otherwise, I've created a new repository and imported the history here, if you can go ahead and propose your merges there that would be great:
https://github.com/lathiat/nss-mdns

agoode:
Sounds good. I'll prepare some patches and send them your way. Thanks for taking care of this.

PR #69, opened 2016-08-09
avahi-daemon: Start the daemon after network is up

rojkov:
In case of ConnMan used as a network manager the file
/etc/resolve.conf isn't available for reading before
connmand is up and running. This precludes Avahi from
loading /etc/resolv.conf.

The patch makes avahi-daemon start after ConnMan is fully
initialized.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>

### DIFF
--- a/avahi-daemon/avahi-daemon.service.in
+++ b/avahi-daemon/avahi-daemon.service.in
@@ -18,6 +18,8 @@
 [Unit]
 Description=Avahi mDNS/DNS-SD Stack
 Requires=avahi-daemon.socket
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=dbus


lathiat:
Thanks for the contribution, likely I will not merge this for the reasons given in the discussion here:
https://github.com/lathiat/avahi/pull/59

Can you tell me if this is actually breaking a use-case for you, and specifically what that use case is? Or were you just trying to fix the error.

rojkov:
Nothing really breaks. Just the messages keep confusing people looking for failures in system logs. I  guess a tree or two would be saved if the guys stopped to file bugs about the messages.

But I'd be happy if the outdated options using resolv.conf were dropped in a future release.

lathiat:
Thanks for the note, what we'll do here is merge a fix to no longer error about resolv.conf which was submitted already:
https://github.com/lathiat/avahi/pull/63

That will fix the immediate concern.. and look at also deprecating later.

Thanks for your contribution and ideas!

Issue #70, opened 2016-08-09
Support user-specified MTU


martinoneutrino:
I would like to request support for a user-configured MTU. In the case of a WLAN mixed with a jumbo-frames configured LAN, Multicast DNS packets that exceed the WLAN MTU are silently dropped.
I would see this as being a simple "mtulimit" setting, which would replace the discovered hardware MTU if the setting is lower. This would allow the multicast packets to be sent with the correct maximum size.

lathiat:
What MTU sizes do you have set on your WLAN and LAN, and what packet sizes are you seeing avahi generate that are greater?

martinoneutrino:
LAN is 7000 bytes, and WLAN is standard 1500. avahi generated (on the LAN) a packet of 1772 bytes, which was never received by devices on the WLAN.

evverx:
Closing because I think it's a niche feature and there has been no response: https://github.com/avahi/avahi/pull/71#issuecomment-1781284132.

pemensik:
I think what would be needed more is self-detection of bridged networks. If avahi would be aware eth0 and wlan0 interfaces are bridged, it might lower MTU as long as it knew they are interconnected.

But anyway. Avahi should send announcements both over eth0 and wlan0 with MTU obtained from those interfaces. It should be okay to send bigger packet than wlan0 is using. The problem would be if we did not announce the same thing also on wlan0 with correct size, making this announcement lost. Only then hosts on wlan0 might not receive announcement properly.

PR #71, opened 2016-08-11
Support mtulimit setting

martinoneutrino:
This will fulfill the enhancement request in #70

### DIFF
--- a/avahi-core/core.h
+++ b/avahi-core/core.h
@@ -53,6 +53,7 @@ typedef struct AvahiServerConfig {
     int publish_workstation;          /**< Register a _workstation._tcp service */
     int publish_domain;               /**< Announce the local domain for browsing */
     int check_response_ttl;           /**< If enabled the server ignores all incoming responses with IP TTL != 255. Newer versions of the RFC do no longer contain this check, so it is disabled by default. */
+    unsigned mtulimit;                /**< If set, the server limits outgoing packets to the lower of this size or the actual MTU. Do not set unless you have a multicast network using different frame sizes. */
     int use_iff_running;              /**< Require IFF_RUNNING on local network interfaces. This is the official way to check for link beat. Unfortunately this doesn't work with all drivers. So bettere leave this off. */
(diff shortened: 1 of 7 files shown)


lathiat:
I'm a bit here and there about merging this.  Generally speaking, i think it's bad network configuration to have devices with differing MTUs on the same network.  This will generally cause problems and just isn't all that well handled in general and not a good idea in practice.  It mostly works with TCP because it negotiates the MTU but obviously we're not doing that here and it's not really possible to negotiate an MTU sensibly.

Having said that, RFC6762 does have this to say:
>    Note that 9000 bytes is also the maximum payload size of an Ethernet
>    "Jumbo" packet [Jumbo].  However, in practice Ethernet "Jumbo"
>    packets are not widely used, so it is advantageous to keep packets
>    under 1500 bytes whenever possible.  Even on hosts that normally
>    handle Ethernet "Jumbo" packets and IP fragment reassembly, it is
>    becoming more common for these hosts to implement power-saving modes
>    where the main CPU goes to sleep and hands off packet reception tasks
>    to a more limited processor in the network interface hardware, which
>    may not support Ethernet "Jumbo" packets or IP fragment reassembly.

Mostly it's talking about the fragmentation side never the less.

Looking at the Bonjour source, they seem to have a concept of NormalMaxDNSMessageData and it looks like they may normally limit the max packet size from a quick read.

If that's the case and given it's just an option we may not enable by default then there may be no harm in merging... will think about it.

martinoneutrino:
I think having a setting for MTU limit will give an admin more flexibility in their network setup. There aren't many multicast services in use, and without MTU negotiation available, it causes some interoperability issues.
This change is a fairly trivial one, but it provides a necessary option which fulfills the "advantageous to keep packets under 1500 bytes whenever possible" as you quoted.

evverx:
> I think having a setting for MTU limit will give an admin more flexibility in their network setup

This PR has been open since 2016 and there don't seem to be any thumb-ups, +1s and so on. It makes me think that this feature is kind of niche and I'm not sure it should be kept upstream. I'll add the "needs-discussion" label.

PR #73, opened 2016-08-29
avahi-daemon: fix SIGHUP/SIGUSR1 race condition at startup

jacmet:
We may get signalled to reload / dump data before signal handlers are set up
(in run_server()) and instead get killed as these signals default to
terminate the process.  Instead ignore them until we are ready

Signed-off-by: Peter Korsgaard peter@korsgaard.com

### DIFF
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -1646,6 +1646,13 @@ int main(int argc, char *argv[]) {
 #endif
         }
 
+        /* We may get signalled to reload / dump data before signal
+         * handlers are set up (in run_server()) and instead get
+         * killed as these signals default to terminate the process.
+         * Instead ignore them until we are ready */
+        ignore_signal(SIGHUP);
+        ignore_signal(SIGUSR1);
+
         if (daemon_pid_file_create() < 0) {
             avahi_log_error("Failed to create PID file: %s", strerror(errno));
 


jacmet:
@lathiat ping?

avahi-daemon/main.c
```diff
@@ -1646,6 +1646,13 @@ int main(int argc, char *argv[]) {
 #endif
         }
 
+        /* We may get signalled to reload / dump data before signal
+         * handlers are set up (in run_server()) and instead get
+         * killed as these signals default to terminate the process.
+         * Instead ignore them until we are ready */
+        ignore_signal(SIGHUP);
+        ignore_signal(SIGUSR1);
```
CameronNemo: SIGUSR1 defaults to being ignored.

jacmet: According to http://man7.org/linux/man-pages/man7/signal.7.html, both default to term:

SIGHUP        1       Term    Hangup detected on controlling terminal
                                     or death of controlling process
..
SIGUSR1   30,10,16    Term    User-defined signal 1

pemensik: Both defaults to Term according to my page.

pemensik (approved):

jacmet:
@pemensik  Great, better late than never ;)

Issue #77, opened 2016-09-14
Allow for binary TXT record values in XML


edmont:
According to https://tools.ietf.org/html/rfc6763#section-6.5 :

> The value is opaque binary data.  Often the value for a particular attribute will be US-ASCII [RFC20] or UTF-8 [RFC3629] text, but it is legal for a value to be any binary data.

However, the current static XML service definition only allows printable characters to be included as TXT record value (http://linux.die.net/man/5/avahi.service).

From an old @poettering message to the list (https://lists.freedesktop.org/archives/avahi/2010-August/001915.html):

> The C API allows you to create binary TXT fields very easily. However, this is not supported in the XML definition for static services, simply because there is no nice way to include binary data in XML files.

What about an attribute in the XML 'text-record' tag?
`<txt-record value="binary">"val=0xf0a9"</txt-record>
`

edmont:
Another option: https://github.com/lathiat/avahi/pull/116

lathiat:
Planning to merge #116, so closing this out.  Thanks.

PR #84, opened 2016-10-03
Added support for binary TXT records in XML service definition

edmont:
Solves #77.

Using the new **`<txt-binary>`** tag in the XML service definition, the value after '=' char is read as an hex binary and accordingly published. The code checks for the presence of '=' and for an even number of characters in the value part, but does not check for valid base 16 digits. The key and equals sign remain human readable.

Valid examples:
`<txt-binary>key=37ab</txt-binary>`
`<txt-binary>foo=abcDEF</txt-binary>` <-- Indistinct use of uppercase and lowercase

Invalid examples:
`<txt-binary>key37ab</txt-binary>` <-- Lack of equals sign
`<txt-binary>foo=abcDEFF</txt-binary>` <-- Odd number of value digits

Wrong behavior example:
`<txt-binary>key=12mb8</txt-binary>` <-- Non base 16 digit.

### DIFF
--- a/avahi-daemon/avahi-service.dtd
+++ b/avahi-daemon/avahi-service.dtd
@@ -16,3 +16,5 @@
 <!ATTLIST port>
 <!ELEMENT txt-record (#PCDATA)>
 <!ATTLIST txt-record>
+<!ELEMENT txt-binary (#PCDATA)>
+<!ATTLIST txt-binary>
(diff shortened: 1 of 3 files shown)


lathiat:
Closing in preference of #116, but thanks for your contribution.

PR #85, opened 2016-10-12
add Xpra service type

totaam:
http://xpra.org/trac/ticket/731
The original ticket I had created for avahi is now 404:
http://avahi.org/ticket/368


lathiat:
Thanks!

PR #101, opened 2017-02-16
Fix watch cleanup issue in watch_free.

cbchan:
When an AvahiWatch instance `w` is freed by watch_free(), watch_free()
is supposed to set `w->glib_poll->watch_req_cleanup = TRUE` to trigger
cleanup_watches() during prepare_func(), but instead it incorrectly sets
`w->glib_poll->timeout_req_cleanup = TRUE`. Without invoking
cleanup_watches(), the AvahiWatch instance isn't properly destroyed and
its associated GSource instance isn't removed. The watch callback could
potentially be invoked inappropriately on the supposedly dead watch.

### DIFF
--- a/avahi-glib/glib-watch.c
+++ b/avahi-glib/glib-watch.c
@@ -87,7 +87,7 @@ static void cleanup_watches(AvahiGLibPoll *g, int all) {
             destroy_watch(w);
     }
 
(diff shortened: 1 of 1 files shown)


lathiat:
This looks great, thanks!

PR #108, opened 2017-03-23
spelling fixes

ka7:
patch contains some spelling fixes ( just in comments ) as found by a bot ( https://github.com/ka7/misspell_fixer ). Any upcoming License changes (e.g. GPL2 to GPL3+) are hereby granted.


evverx:
Thanks for the PR!

Looking at the latest codespell report at https://fossies.org/linux/test/avahi-master.tar.gz/codespell.html it seems all these typos have been fixed since then. I think this PR can be closed.

(I might have missed something here but either way I think it would be better to open a new PR based on top of the master branch).

PR #112, opened 2017-03-31
avahi-autoipd: clear previously set address before binding a new one

chombourger:
Delete any IP addresses set to the avahi alias/label before setting a new
so we do not end-up with multiple IP addresses

See https://github.com/lathiat/avahi/issues/111

Signed-off-by: Cedric Hombourger <Cedric_Hombourger@mentor.com>


lathiat:
Looks good to me, thanks!

PR #115, opened 2017-04-19
filter denied virtual interfaces when adding addresses

### LINKED ISSUE #103: Avahi registering new address on wrong interface when using ip aliasing
Having the following hardware interface, Avahi registers `192.168.42.1` on `eth0`, which is what I want:
```
eth0      Link encap:Ethernet  HWaddr A0:14:3D:C5:B3:6F  
          inet addr:192.168.42.1  Bcast:192.168.42.255  Mask:255.255.255.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:126 errors:0 dropped:3 overruns:0 frame:3013
          TX packets:362 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:19377 (18.9 KiB)  TX bytes:41751 (40.7 KiB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:16436  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```

The problem is when I add an ip address using ip aliasing with `ifconfig eth0:0 192.168.42.2` (see #99 for details on why I need to do that):

ifconfig:
```
...
eth0:0    Link encap:Ethernet  HWaddr A0:14:3D:C5:B3:6F  
          inet addr:192.168.42.2  Bcast:192.168.42.255  Mask:255.255.255.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
...
```
Avahi:
```
<system startup>
...
Joining mDNS multicast group on interface eth0.IPv4 with address 192.168.42.1.
New relevant interface eth0.IPv4 for mDNS.
...
Registering new address record for 192.168.42.1 on eth0.*.
...
<somewhere here I add the interface>
...
Registering new address record for 192.168.42.2 on eth0.*.
...

```
Avahi registers the new address on `eth0` instead of `eth0:0`, which makes it answer discovery requests with the wrong ip (192.168.42.2) and is the cause of all of my problems :confused:

As I also noted, I didn't see any trace of a new relevant interface `eth0:0` being added.

I also know that `eth0:0`, as a virtual interface, will have the same index as `eth0` but a different label and its own message in the `netlink` response to a `RTM_GETADDR` query.

I am currently investigating the [code](https://github.com/lathiat/avahi/blob/master/avahi-core/iface.c#L76) to find a workaround or a proper fix to this issue, as I strongly need this setup to be working.

If you had any advice, help, or heads start to give me that could help with this issue it would be strongly needed and gladly welcomed :smile_cat:

deb0ch:
netlink does not send an interface change notification when setting up a virtual
interface such as eth0:0: that information is only available in RTM_NEWADDR
messages.

It is therefore needed to check if a new address does belong to an allowed
interface label before actually adding it to the hardware interface.

Fixes #103

### DIFF
--- a/avahi-core/iface-linux.c
+++ b/avahi-core/iface-linux.c
@@ -183,6 +183,7 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
         size_t l;
         AvahiAddress raddr, rlocal, *r;
         int raddr_valid = 0, rlocal_valid = 0;
+        char label[IFNAMSIZ + 1];
 
         /* We are only interested in IPv4 and IPv6 */
         if (ifaddrmsg->ifa_family != AF_INET && ifaddrmsg->ifa_family != AF_INET6)
@@ -193,6 +194,8 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
          * for this interface, we ignore this address. */
         if (!(i = avahi_interface_monitor_get_interface(m, (AvahiIfIndex) ifaddrmsg->ifa_index, avahi_af_to_proto(ifaddrmsg->ifa_family))))
             return;
+
+        memset(label, 0, IFNAMSIZ + 1);
 
         /* Fill in address family for our new address */
         rlocal.proto = raddr.proto = avahi_af_to_proto(ifaddrmsg->ifa_family);
@@ -229,15 +232,31 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
 
                     memcpy(raddr.data.data, RTA_DATA(a), RTA_PAYLOAD(a));
                     raddr_valid = 1;
-
                     break;
+
+                case IFA_LABEL: {
+                    size_t label_len;
+
+                    if (RTA_PAYLOAD(a) > IFNAMSIZ) {
+                        label_len = IFNAMSIZ;
+                        avahi_log_warn("interface label longer than IFNAMSIZ !");
+                    }
+                    else
+                        label_len = RTA_PAYLOAD(a);
+                    memcpy(label, RTA_DATA(a), label_len);
+                    label[label_len] = '\0';
+                }
+                break;
 
                 default:
                     ;
             }
-
             a = RTA_NEXT(a, l);
         }
+
+        /* Ignore if label denied in avahi-daemon.conf. */
+        if (strlen(label) != 0 && !avahi_interface_label_is_relevant(i, label))
+            return;
 
         /* If there was no adress attached to this message, let's quit. */
         if (rlocal_valid)
--- a/avahi-core/iface.c
+++ b/avahi-core/iface.c
@@ -675,25 +675,30 @@ static int avahi_interface_is_relevant_internal(AvahiInterface *i) {
     return 0;
 }
 
+int avahi_interface_label_is_relevant(AvahiInterface *i, const char *label) {
+    AvahiStringList *l;
+    assert(i);
+    assert(label);
+
+    for (l = i->monitor->server->config.deny_interfaces; l; l = l->next)
+        if (strcasecmp((char*) l->text, label) == 0)
+            return 0;
+    if (!i->monitor->server->config.allow_interfaces)
+        return 1;
+    for (l = i->monitor->server->config.allow_interfaces; l; l = l->next)
+        if (strcasecmp((char*)l->text, label) == 0)
+            return 1;
+    return 0;
+}
+
 int avahi_interface_is_relevant(AvahiInterface *i) {
     AvahiStringList *l;
     assert(i);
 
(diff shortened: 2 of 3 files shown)


avahi-core/iface-linux.c
```diff
@@ -194,6 +195,8 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
         if (!(i = avahi_interface_monitor_get_interface(m, (AvahiIfIndex) ifaddrmsg->ifa_index, avahi_af_to_proto(ifaddrmsg->ifa_family))))
             return;
 
+        memset(label, 0, IFNAMSIZ + 1);
```
kalvdans: It is enough to do
<pre>
label[0] = '\0';
</pre>

deb0ch: Never too sure, right ? It doesn't cost much and makes buffers cleaner.

kalvdans: Yes, it is not a big deal. <pre>char label[IFNAMSIZ + 1] = "";</pre> have also the same effect of zeroing out the whole array without using a separate memset.

deb0ch:
Hi, any news on this ? @lathiat

lathiat:
Thanks for the PR. I'd like to merge this, however with two fixes

 (1) Currently it only works with deny-interfaces and not allow-interfaces. I guess to make it work with allow-interfaces we need to expand allow-interfaces to filter out the part before the colon when deciding if interfaces are relevant
 (2) Let's use sizeof(label) instead of repeating the size

xandey:
@lathiat Thanks for pointing out that I can use this. It solves my problem.

In case anyone gets here from google: I have a device which uses a hard coded fallback IP. You can put the hard coded IP on an aliased interface then use this PR to deny that virtual interfaces. This way the hard coded IP is skipped from mDNS record since its likely that that won't be routable.

I have avahi-autoipd which assigns eth:avahi the IPv4LL address.
I also manually allocate eth0:fallback in the fallback case.
I use deny-interfaces to specify eth0:fallback.

This way avahi-daemon uses any static IP on eth0, then uses the link-local address as it's fallback.

Jasper-Ben:
This is a feature we are interested in as well. What is the current status here and how can we help to ensure this is part of the next release?

deb0ch:
Nice to see that some people actually need this as well ! ❤️ 

To respond to @lathiat (after all this time) :

> Thanks for the PR. I'd like to merge this, however with two fixes
> 
> (1) Currently it only works with deny-interfaces and not allow-interfaces. I guess to make it work with allow-interfaces we need to expand allow-interfaces to filter out the part before the colon when deciding if interfaces are relevant
> (2) Let's use sizeof(label) instead of repeating the size

I haven't worked on this for years, do not even work in this company anymore, and it was already the case when you first replied. I now have so little time to dedicate to my own coding projects that I can hardly imagine investing more time on this PR.

For (1), reading the commit again it seems to me that it is already working for allow-interfaces : where before it was the `i->hardware->name`, it now the same `label` that is compared as for the denied interfaces. If it was not the case, it could also be the scope of a separate development. 

(2) is so easy to fix that it could be done by anybody who merges this PR.

Jasper-Ben:
Thanks for the reply @deb0ch! :slightly_smiling_face:
We are currently applying your changes as patch to v0.7 and it seems to work just fine so far (has not hit our production yet, though). Having this feature in upstream would be super helpful.

JsBergbau:
Any update when this pull request will be merged? Would be very useful if added soon.

guillerodriguez:
Similar use case here, we have an Ethernet interface eth0 with a primary address, and an alias interface eth0:1 with another IP address that we don't want published by Avahi. We tried to deny-interface eth0:1 but that didn't work, then arrived at this PR which seems to solve exactly the same problem.

I am aware that #103 was closed as "not a bug with avahi-daemon" but I think it would be desirable to support alias interfaces in allow-interfaces / deny-interfaces, as that's what users would probably expect when configuring avahi-daemon.conf.

evverx:
> I think it would be desirable to support alias interfaces in allow-interfaces / deny-interfaces

The problem is that in its current form with this patch applied `allow-interfaces=iface:0` doesn't work: https://github.com/avahi/avahi/issues/528#issuecomment-1818425069 so it's kind of half implemented at this point.

guillerodriguez:
I see. In our case we tested `deny-interfaces=eth0:1` and `allow-interfaces=eth0` and they both worked.

I understand that the problem is that virtual interface names don't seem to work in `allow-interfaces`.

PR #119, opened 2017-05-11
database: fix duplicate service

slokhorst:
"Posta - POP3" seems to be Italian, mark it as such.


slokhorst:
The line was added in 82a95a3679057a8d9e90aba42ffb1bc987573683

lathiat:
Thanks!

PR #120, opened 2017-05-12
Update po/LINGUAS

piotrdrag:
The new Occitan translation is missing from po/LINGUAS, which causes it not to be used at all.


lathiat:
Thanks!

PR #124, opened 2017-05-19
Remove invalid po/sl.si file

piotrdrag:
Looks like it was committed erroneously a long time ago. The real Slovenian translation is in po/sl.po.


lathiat:
Thanks!

PR #133, opened 2017-07-13
python: Added a txt_array_to_dict functions

muelli:
Avahi consumers seem to cargo-cult this functionality
from across projects, e.g. https://dev.gajim.org/gajim/gajim/blob/master/gajim/common/zeroconf/zeroconf_avahi.py#L131
It does indeed seem reasonably to have Avahi ship this function,
so this patch includes it.

### DIFF
--- a/avahi-python/avahi/__init__.py
+++ b/avahi-python/avahi/__init__.py
@@ -110,3 +110,22 @@ def dict_to_txt_array(txt_dict):
         l.append(string_to_byte_array("%s=%s" % (k,v)))
 
     return l
+
+
+def txt_array_to_dict(self, txt_array):
+    txt_dict = {}
+    for els in txt_array:
+        key, val = '', None
+        for c in els:
+            c = chr(c)
+            if val is None:
+                if c == '=':
+                    val = ''
+                else:
+                    key += c
+            else:
+                val += c
+        if val is None: # missing '='
+            val = ''
+        txt_dict[key] = val
+    return txt_dict


avahi-python/avahi/__init__.py
```diff
@@ -110,3 +110,22 @@ def dict_to_txt_array(txt_dict):
...
 
     return l
+
+
+def txt_array_to_dict(self, txt_array):
+    txt_dict = {}
+    for els in txt_array:
+        key, val = '', None
+        for c in els:
+            c = chr(c)
```
smcv: In Python 3 this would have to be something like `bytes((c,))` to build up a bytes object: TXT records aren't guaranteed to be in any particular encoding, so (Unicode) strings aren't a suitable rerpresentation.

muelli: do you know any polyglot version?  I assume this module to be used by a lot of python2 projects. In fact, Debian hasn't managed yet to ship this trivially compatible module for python3. Yet, it'd be nice to have something that works on both Pythons, I guess.

smcv: The best I can come up with is:

```
def _bytechr(c):
    """
    if str is bytes:
        return chr(c)
    else:
        return bytes((c,))
```

and use that instead of `chr`.

I think this function really needs some unit tests that can be run on both Python 2 and Python 3, with assertions like:

```
txt_array_to_dict(self, [map(ord, 'foo')]) == {b'foo': b''}
txt_array_to_dict(self, [map(ord, 'foo=')]) == {b'foo': b''}
txt_array_to_dict(self, [map(ord, 'foo=bar')]) == {b'foo': b'bar'}
txt_array_to_dict(self, [[0, 1, 2, 0xff, 0x99]]) == {b'\x00\x01\x02\xff\x99': b''}
```

Happily, it doesn't do any I/O or networking, so it should be trivial to test.

avahi-python/avahi/__init__.py
```diff
@@ -110,3 +110,22 @@ def dict_to_txt_array(txt_dict):
         l.append(string_to_byte_array("%s=%s" % (k,v)))
 
     return l
+
+
+def txt_array_to_dict(self, txt_array):
```
smcv: What's `self` here? It doesn't seem to be used or useful?

avahi-python/avahi/__init__.py
```diff
@@ -110,3 +110,22 @@ def dict_to_txt_array(txt_dict):
...
+            c = chr(c)
+            if val is None:
+                if c == '=':
+                    val = ''
+                else:
+                    key += c
+            else:
+                val += c
+        if val is None: # missing '='
+            val = ''
```
smcv: Is it intentional that this makes it impossible to distinguish between `foo` and `foo=`?

I would have expected the TXT records

```
no_value
empty_value
has_value=foo
```

to come out in Python as, perhaps, something like

```
{
    b'no_value': None,
    b'empty_value': b'',
    b'has_value': b'foo',
}
```

(On one hand, `dict_to_txt_array` doesn't handle that either; on the other hand, perhaps it should, since https://tools.ietf.org/html/rfc6763 does specifically distinguish between an empty value and the absence of a value.)

evverx:
I think it can be closed. I agree with Simon that it should be possible to tell `key` and `key=` apart and `self` shouldn't be there. There are no tests either.

PR #134, opened 2017-07-13
avahi-python: Use the agnostic DBM interface

heftig:
Also fixes configure failing if Python 3 is the build python and GDBM is
enabled, since Py3 only has anydbm under the name of 'dbm'.

Not enough to make ServiceTypeDatabase.py compatible with Py3, but it's
a start.

### DIFF
--- a/service-type-database/build-db
+++ b/service-type-database/build-db
@@ -1,4 +1,4 @@
-#!@PYTHON@
+#!/usr/bin/env python
 # -*-python-*-
 # This file is part of avahi.
 #
@@ -17,7 +17,12 @@
 # Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
 # USA.
 
-import @DBM@, sys
+try:
+    import anydbm as dbm
+except ImportError:
+    import dbm
+
+import sys
 
 if len(sys.argv) > 1:
     infn = sys.argv[1]
@@ -29,9 +34,9 @@ if len(sys.argv) > 2:
 else:
     outfn = infn + ".db"
 
-db = @DBM@.open(outfn, "n")
+db = dbm.open(outfn, "n")
 
-for ln in file(infn, "r"):
+for ln in open(infn, "r"):
     ln = ln.strip(" \r\n\t")
     
     if ln == "" or ln.startswith("#"):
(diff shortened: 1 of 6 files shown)


service-type-database/build-db
```diff
@@ -1,4 +1,4 @@
-#!@PYTHON@
+#!/usr/bin/env python
```
andhe: python3 ?!

mbiebl: The use of /usr/bin/env is an unrelated change and should be dropped from this pull request or at least be moved into a separate commit with a proper justification for this change. Distro packages are stronly discouraged to use /usr/bin/env, as this can cause them to pick up a "random" python version. So, instead of changing this to "/usr/bin/env python3", please drop this hunk completely.

mbiebl: https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html#s-interpreter

troycurtisjr: For what it is worth, Fedora has similar guidelines: https://fedoraproject.org/wiki/Packaging:Python#Multiple_Python_Runtimes

lathiat:
Thanks! With the unicode change in #178 you seem to be able to launch avahi-discover with either python2 or python3 now.

service-type-database/build-db
smcv: Distributions can edit the `#!` line according to their preferred policies at install time if they need to (in Debian, `dh-python` does that), so this is not necessarily a guideline that upstreams need to follow.

(But yes, it would have been better if this change was in a separate commit where it could be justified and merged (or not) on its own merits.)

andhe:
@heftig I tried to "make dist" and build avahi from the generated (snapshot) tarball, but the build-db file was missing and make failed because of it. Likely because of this PR. Do you have any idea what the problem is and how to fix it?

heftig:
@andhe
`noinst_SCRIPTS` in `service-type-database/Makefile.am` needs to be changed to `dist_noinst_SCRIPTS`.

pyhalov:
There's issue in this change - Python 3 will prever gdbm by default. So when  --disable-gdbm, --enable-dbm is given, service-type-database build fails.

Issue #135, opened 2017-07-17
examples/core-publish-service.c: server_callback() doesn't re-register services


jfether:
In the example for Avahi Core publish service, the server_callback() routine has a logic error. 

```C
        case AVAHI_SERVER_RUNNING:
            /* The serve has startup successfully and registered its host
             * name on the network, so it's time to create our services */

            if (!group)
                create_services(s);

            break;

        case AVAHI_SERVER_COLLISION: {
            char *n;
            int r;

            /* A host name collision happened. Let's pick a new name for the server */
            n = avahi_alternative_host_name(avahi_server_get_host_name(s));
            fprintf(stderr, "Host name collision, retrying with '%s'\n", n);
            r = avahi_server_set_host_name(s, n);
            avahi_free(n);

            if (r < 0) {
                fprintf(stderr, "Failed to set new host name: %s\n", avahi_strerror(r));

                avahi_simple_poll_quit(simple_poll);
                return;
            }

        }

            /* Fall through */

        case AVAHI_SERVER_REGISTERING:

	    /* Let's drop our registered services. When the server is back
             * in AVAHI_SERVER_RUNNING state we will register them
             * again with the new host name. */
            if (group)
                avahi_s_entry_group_reset(group);

            break;
```
The condition `!group` means compares against the value of group being NULL, but the reset code does not change the pointer to NULL, therefore the behaviour does not work as expected.

I will post a pull request shortly to address the issue.

PR #136, opened 2017-07-17
Fix logic error in server_callback()

jfether:
server_callback() follows logic to handle AVAHI_SERVER_COLLISION and AVAHI_SERVER_REGISTERING states with an unregister of the service group, but does not reset the pointer to the group to NULL. Added a avahi_s_entry_group_free(group) and reset of the pointer to NULL to ensure the re-registration succeeds.

This is to resolve issue #135.

### DIFF
--- a/examples/core-publish-service.c
+++ b/examples/core-publish-service.c
@@ -168,12 +168,14 @@ static void server_callback(AvahiServer *s, AvahiServerState state, AVAHI_GCC_UN
 
         case AVAHI_SERVER_REGISTERING:
 
-	    /* Let's drop our registered services. When the server is back
+            /* Let's drop our registered services. When the server is back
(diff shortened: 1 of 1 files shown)


evverx:
Thanks for the PR! As far as I can tell it should make it publish the service when conflicts occur but setting `group` to `NULL` leads to creating a new group every time conflicts are detected and it would probably be better to reuse the same group again instead. It seems to me that instead of `(!group)` the example should use something like
```
(!group || avahi_s_entry_group_is_empty(group))
```
With that exactly one group is created. I get that it's just an example but still.

evverx:
That being said I think this PR should be good to go. I replaced the tabs, bumped the bracket (to make it consistent with avahi) and force-pushed the code to trigger the CI.

PR #138, opened 2017-08-02
avahi-daemon: fix segfault when parsing static service XML with TXT r…

BrsBrd:
…ecords key with empty value

Adding support of TXT record value-format introduced regression causing the daemon to segfault when .service files contained key with empty value.
ex:
`<txt-record>keyWithEmptyValue=</txt-record>`

Fix tested with TXT record:
 - without value,
 - with empty value,
 - with value
Each time without value-format and with value format set to "text", "binary-hex" and "binary-base64".

Also updating example.service.

### DIFF
--- a/avahi-daemon/example.service
+++ b/avahi-daemon/example.service
@@ -29,6 +29,24 @@
   <service>
     <type>_printer._tcp</type>
     <port>515</port>
+    
+    <!-- TXT records without value-->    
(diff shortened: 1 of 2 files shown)


pemensik (approved):

evverx:
I think this issue was fixed in https://github.com/lathiat/avahi/pull/226.

evverx:
Since the issue is already fixed I think this PR can be closed.

PR #146, opened 2017-09-13
Update avahi-discover.py

yingchaolu:
changes due to python3: has_key() is deprecated in favor of  key in d.

### DIFF
--- a/avahi-python/avahi-discover/avahi-discover.py
+++ b/avahi-python/avahi-discover/avahi-discover.py
@@ -209,10 +209,10 @@ class Main_window:
             self.new_service_type(interface, protocol, self.stype, domain)
 
     def new_domain(self,interface, protocol, domain, flags):
-        if self.zc_ifaces.has_key((interface,protocol)) == False:
+        if ((interface,protocol) in self.zc_ifaces) == False:
             ifn = self.get_interface_name(interface, protocol)
             self.zc_ifaces[(interface,protocol)] = self.insert_row(self.treemodel, None, ifn,None,interface,protocol,None,domain)
-        if self.zc_domains.has_key((interface,protocol,domain)) == False:
+        if ((interface,protocol,domain) in self.zc_domains) == False:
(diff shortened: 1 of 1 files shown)


avahi-python/avahi-discover/avahi-discover.py
```diff
@@ -209,10 +209,10 @@ def browse_domain(self, interface, protocol, domain):
             self.new_service_type(interface, protocol, self.stype, domain)
 
     def new_domain(self,interface, protocol, domain, flags):
-        if self.zc_ifaces.has_key((interface,protocol)) == False:
+        if ((interface,protocol) in self.zc_ifaces) == False:
```
BirdAPI: shouldn't this use `not in` like so:

```python
if (interface,protocol) not in self.zc_ifaces:
```

avahi-python/avahi-discover/avahi-discover.py
```diff
@@ -209,10 +209,10 @@ def browse_domain(self, interface, protocol, domain):
...
             ifn = self.get_interface_name(interface, protocol)
             self.zc_ifaces[(interface,protocol)] = self.insert_row(self.treemodel, None, ifn,None,interface,protocol,None,domain)
-        if self.zc_domains.has_key((interface,protocol,domain)) == False:
+        if ((interface,protocol,domain) in self.zc_domains) == False:
```
BirdAPI: use `not in` like above

avahi-python/avahi-discover/avahi-discover.py
yingchaolu: Yes. This looks better.

avahi-python/avahi-discover/avahi-discover.py
yingchaolu: I think they do the same thing but your suggestion is better.

rmathot:
I also had tracebacks with this when trying to run avahi-discover with Python3 (while it is avertised to be P3-compatible). Please merge this PR :) 👍

PR #152, opened 2017-09-27
Add a shell completion function for zsh

okapia:
This adds a shell auto-completion function for use with zsh.

In the source tree, I've put this in a top-level `completion` directory but how that is organised is not something I'm especially concerned by.

Ideally, it should be installed to `${prefix}/share/zsh/site-functions` to have a good chance of being picked up by zsh at runtime. I'm no automake expert but I think that ends up being something like:
```
zshcompdir = $(datadir)/zsh/site-functions
dist_zshcomp_DATA = _avahi
EXTRA_DIST = _avahi
```
It'd be great if this could be included.

### DIFF
--- a/completion/_avahi
+++ b/completion/_avahi
@@ -1,0 +1,54 @@
+#compdef avahi-browse avahi-browse-domains avahi-resolve avahi-resolve-address avahi-resolve-host-name
+
+# Completion function for zsh
+
+local curcontext="$curcontext" pri ret=1
+local -A opt_args
+local -a state line args serv desc disp
+
+args=(
+  '(- 1)'{-h,--help}'[display usage information]'
+  '(- 1)'{-V,--version}'[display version information]'
+  '(-v --verbose)'{-v,--verbose}'[enable verbose mode]'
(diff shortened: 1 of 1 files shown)


g0tar:
[Upstreamed in zsh](https://github.com/zsh-users/zsh/commits/master/Completion/Unix/Command/_avahi), can be closed here.

evverx:
Thanks for looking into this. Closing.

okapia:
While committing to zsh somewhat solves the problem, in general it is better for upstream projects to carry completion functions. There tends to be a better chance that they get updated as and when new options are added. Hence I contributed it here first and only pushed the zsh repository after this PR was ignored for a couple of years. Please consider including the function within avahi. I would then take care of removing it from zsh.

evverx:
> in general it is better for upstream projects to carry completion functions

Agreed but currently there is no bandwidth (upstream) to get this PR to the point where it's fully integrated into the build system. If you are interested in updating the PR and integrating it into the build system I think it can be considered of course.

PR #154, opened 2017-10-14
avahi-autoipd.action: Avoid some stat calls

scop:


lathiat:
Thanks for the pull request, in reality we should probably replace this manual check with a check that actually looks in $PATH, e.g. "where" or similar -- however that would almost certainly increase the number of stat calls and would probably rarely matter.

Is there a specific reason you wanted to reduce these?

evverx:
This PR fixes two shellcheck warnings. That being said currently shellcheck reports 184 issues five of which violate the same "SC2166 rule" so if it was about fixing shellcheck warnings it would probably make sense to fix them all at once. Also I agree with Trent that this custom check should probably be replaced with a better way to locate binaries in the first place but looks like it works as is and isn't a hot path anyway. I think it can be closed. (there are already quite a few one-line patches making it hard to cleanly rebase other PRs on top of the master branch to trigger the CI so I'd avoid merging another one. Once those PRs are handled I guess it could be revisited).

Issue #125, opened 2017-05-20
Advertising services on the local machine


lathiat:
It is desirable to support advertising services but only to the local machine, and not to advertise them to the network.

The specific use case is the IPPUSBXD project (https://github.com/tillkamppeter/ippusbxd/)

Currently we disable the loopback interface unless no other interfaces are activated.  We should consider just enabling it at all times - there also seems to be a problem with the lo.IPv6 interface at least in avahi-discover it does not appear which is surprising.

Review the usage of loopback and figure out how to handle this nicely.  One drawback to always advertising it is that local services will always appear 'twice' from 2 interfaces even on single interface machines.  We could potentially not advertise services on loopback unless specifically requested, or not worry about it since we need to improve duplicate names on multiple interfaces anyway since we enabled IPv6 by default.

tillkamppeter:
Duplicate appearing of services I have already without "lo" support. My laptop runs the Virtual Machine Manager (libvirt-based) and this creates a bridge network interface ("virbr") to provide network for the VMs by a kind of virtual NAT router. This makes all local CUPS queues (and other services from my laptop) already appearing under two interfaces. The same happens if you connect wired and wireless network simultaneously.
So for me it looks common having a service appearing in avahi-discover more than once. So I assume that clients must be able to handle it.

tillkamppeter:
Attaching the patch which makes entries on the "lo" interface being advertised, as suggested in the documentation of ippusbxd. Note that more work on Avahi is needed.

[list-lo-entries.diff.txt](https://github.com/lathiat/avahi/files/1016328/list-lo-entries.diff.txt)

tillkamppeter:
My problem is that if I apply my patch, the IPP-over-USB printer gets advertised (on the "lo" interface) and one can see it in the output of avahi-discover. The record shows its address with the correct IP, 127.0.0.1, but not with the hostname "localhost". Instead, the network host name of my laptop is shown, something like "XXX.local". So CUPS or cups-browsed (when not configured to use only IP addresses and no host names) try to access the printer via the URI "ipp://XXX.local:60000/ipp/print" (Note that ippusbxd uses port 60000 here as the standard IPP port 631 is already occupied by CUPS) and fail, as XXX.local is resolved to the laptop's network IP address and not to 127.0.0.1. Configuring cups-browsed to use only IP addresses makes it successfully using the URI "ipp://127.0.0.1:60000/ipp/print".
[...]

Issue #10, opened 2015-07-30
Debian #751288: Avahi-daemon is noisy


lathiat:
Package: avahi-daemon
Version: 0.6.31-4

Avahi-daemon fills my logs with

 Jun 11 07:38:57 lanthane avahi-daemon[2502]: Invalid response packet from host 172.23.36.97.
 Jun 11 07:39:56 lanthane avahi-daemon[2502]: Invalid response packet from host 172.23.43.78.
 Jun 11 07:39:56 lanthane avahi-daemon[2502]: Invalid response packet from host 172.23.36.97.

This will become a big problem as apparently Windows 10 causes this.

See also https://bugs.launchpad.net/ubuntu/+source/avahi/+bug/1342400
which references a patch that would help (haven't tested it myself yet,
may do today).

Maybe this bug needs to be merged with #639163
(which is already merged with #688061, although IMHO that's a different
problem).

Paul

jmaggard10:
The patch in that bug deals with OSX mDNS response packets that include Additional RRs, but do not include Answer RRs.

Windows 10 does something different.  It sends responses to *all* mDNS queries, whether or not it has any records for those requests.  That is, it responds with a mDNS packet containing 0 Answer RRs and 0 Additional RRs.  This appears to me to be in violation of RFC 6762 (Multicast DNS) Section 6, which states "A Multicast DNS responder MUST only respond when it has a positive, non-null response to send, or it authoritatively knows that a particular record does not exist."  I don't know how to file a bug with Microsoft, but IMHO this should probably be addressed on their side.

jmaggard10:
Maybe something like this makes sense?
```
diff --git a/avahi-core/server.c b/avahi-core/server.c
index e87f035..18b1c84 100644
--- a/avahi-core/server.c
+++ b/avahi-core/server.c
@@ -972,8 +972,14 @@ static void dispatch_packet(AvahiServer *s, AvahiDnsPacket *p, const AvahiAddres
             return;
         }
 
+        if (avahi_dns_packet_get_field(p, AVAHI_DNS_FIELD_ANCOUNT) == 0 &&
+            avahi_dns_packet_get_field(p, AVAHI_DNS_FIELD_ARCOUNT) == 0) {
+
+            avahi_log_debug("Received empty response packet from host %s.", avahi_address_snprint(t, sizeof(t), src_address));
+            return;
+        }
+
[...]

PR #161, opened 2017-12-15
Add support to advertise local services ("localhost") on the local machine only

tillkamppeter:
This is the patch attached to Issue #125 (on Dec 6, 2017) and also
shown in the readme.md of ippusbxd
(https://github.com/OpenPrinting/ippusbxd).

It makes also services on the loopback ("lo") interface being
advertised and these records use "localhost" instead of the network
host name of the machine as server host name. This way clients, like
for example CUPS or cups-browsed will find these local services and be
able to work with them as they were network services.

### DIFF
--- a/avahi-core/iface-linux.c
+++ b/avahi-core/iface-linux.c
@@ -104,8 +104,8 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
         hw->flags_ok =
             (ifinfomsg->ifi_flags & IFF_UP) &&
             (!m->server->config.use_iff_running || (ifinfomsg->ifi_flags & IFF_RUNNING)) &&
-            !(ifinfomsg->ifi_flags & IFF_LOOPBACK) &&
-            (ifinfomsg->ifi_flags & IFF_MULTICAST) &&
+            ((ifinfomsg->ifi_flags & IFF_LOOPBACK) ||
+             (ifinfomsg->ifi_flags & IFF_MULTICAST)) &&
             (m->server->config.allow_point_to_point || !(ifinfomsg->ifi_flags & IFF_POINTOPOINT));
 
         /* Handle interface attributes */
--- a/avahi-core/iface-pfroute.c
+++ b/avahi-core/iface-pfroute.c
@@ -80,8 +80,8 @@ static void rtm_info(struct rt_msghdr *rtm, AvahiInterfaceMonitor *m)
   hw->flags_ok =
     (ifm->ifm_flags & IFF_UP) &&
     (!m->server->config.use_iff_running || (ifm->ifm_flags & IFF_RUNNING)) &&
-    !(ifm->ifm_flags & IFF_LOOPBACK) &&
-    (ifm->ifm_flags & IFF_MULTICAST) &&
+    ((ifm->ifm_flags & IFF_LOOPBACK) ||
+     (ifm->ifm_flags & IFF_MULTICAST)) &&
     (m->server->config.allow_point_to_point || !(ifm->ifm_flags & IFF_POINTOPOINT));
 
   avahi_free(hw->name);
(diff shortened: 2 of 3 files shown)


lathiat:
Note for later; the patch in it's current form (including as shipping on Ubuntu) causes messages like this:
sendmsg() to 0:0:ff02:: failed: Network is unreachable

This seems to happen becuase the AvahiInterface protocol is set to PF_LOCAL instead of PF_INET or PF_INET6 as we assume elsewhere. This causes some if/else statements not to run to send data including the mcast address. 


(gdb) bt
#0  sendmsg_loop (fd=12, msg=0x7fffffffe190, flags=0) at socket.c:460
#1  0x00007ffff7f902a8 in avahi_send_dns_packet_ipv6 (fd=12, interface=1, p=0x5555555ad4c0, 
    src_address=0x55555558bd9c, dst_address=0x0, dst_port=0) at socket.c:624
#2  0x00007ffff7f814e8 in avahi_interface_send_packet_unicast (i=0x55555558bd60, p=0x5555555ad4c0, a=0x0, port=0)
    at iface.c:601
#3  0x00007ffff7f8156a in avahi_interface_send_packet (i=0x55555558bd60, p=0x5555555ad4c0) at iface.c:608
#4  0x00007ffff7f9374f in append_known_answers_and_send (s=0x55555558c660, p=0x5555555ad4c0) at query-sched.c:269
#5  0x00007ffff7f938c4 in elapse_callback (e=0x5555555aa670, data=0x5555555aa620) at query-sched.c:310
#6  0x00007ffff7f7ee77 in expiration_event (timeout=0x55555558db00, userdata=0x555555588b00) at timeeventq.c:94
#7  0x00007ffff7fc34fa in timeout_callback (t=0x55555558db00) at simple-watch.c:447
#8  0x00007ffff7fc38e1 in avahi_simple_poll_dispatch (s=0x555555584ac0) at simple-watch.c:570
#9  0x00007ffff7fc3a4a in avahi_simple_poll_iterate (s=0x555555584ac0, timeout=-1) at simple-watch.c:605
#10 0x000055555555ef2f in run_server (c=0x55555557e200 <config>) at main.c:1256
#11 0x000055555555fc38 in main (argc=1, argv=0x7fffffffe528) at main.c:1674

tillkamppeter:
@rithvikp1998, could you have a look into this issue which @lathiat is mentioning?

rithvikp1998:
Yeah, I'll take a look at it and will give an update by the end of the week

rithvikp1998:
I wanted to clarify one thing though. I just reread the discussion on the issue this patch addresses (#125) and I wanted to confirm if we are going ahead with this implementation at all? It seems that there was no conclusion to the discussion on the issue. @lathiat? @tillkamppeter?

lathiat:
I'm working on the CUPS issue this week. My current plan is to allow lo as an interface, but not to replace the hostname with 'localhost'. And to fix the issue with localhost/the hostname inside CUPS.

Optionally I am considering a way to make it so that localhost is a special interface and only services created specifically on localhost appear there but I'm not too worried about that for the immediate case.

lathiat:
Will update the CUPS bug in the next couple of days. I have an IPPUSB printer arriving today so I can finish debugging a patch for CUPS.

tillkamppeter:
As discussion has moved over from issue #125 to here, I am taking Mike Sweet back into the loop.
@michaelrsweet, @lathiat is working on the Avahi localhost issue now. Please could you chime in here as we (me and you) agreed on that Avahi already should advertise the local service with the "localhost" host name.

michaelrsweet:
From #125:

> If Avahi returns 127.0.0.1 as one of the addresses for a .local lookup, that will cause some serious security problems when machine A (a.local.) looks up machine B ("b.local.") and gets its own loopback address. By returning localhost ("localhost.") that security issue is avoided.
>
> Keep in mind as well that when CUPS tries to connect to a printer/server, it tries all of the addresses returned by a lookup in parallel until one of the connections succeeds. Since CUPS also validates the Host: header in requests (and block any attempt to communicate with cupsd over the loopback interface if the hostname is not "localhost" or "localhost."), this will result in a successful connection but a failed request, breaking printing.
>
> So you really do need to return "localhost" for services registered on the loopback interface.

PR #163, opened 2018-01-12
Avahi: Protected strcmp against NULL pointers in multithreaded apps

MarquisdeGeek:
A bug is triggered when a dbus message arrives, but the Avahi library
hasn't completed initialisation of a browser object (AvahiRecordBrowser
and potentially AvahiServiceTypeBrowser). Until initialisation has
completed, any dbus message that attempts to interrogate a browser
linked list (domain_browsers/service_type_browsers) will issue a segfault.

At various places in the code, b->path is involved in a string compare. By
initialising path to NULL, and adding it to the linked list immediately,
it is possible that strcmp uses a NULL ptr for its comparison (by code on
other threads) before the path is properly assigned by the avahi_strdup
call.

This fix checks for NULL before it is used.

Signed-off-by: Steven Goodwin <marquisdegeek@gmail.com>

### DIFF
--- a/avahi-client/browser.c
+++ b/avahi-client/browser.c
@@ -281,7 +281,7 @@ DBusHandlerResult avahi_domain_browser_event (AvahiClient *client, AvahiBrowserE
         goto fail;
 
     for (db = client->domain_browsers; db; db = db->domain_browsers_next)
-        if (strcmp (db->path, path) == 0)
+        if (db->path && strcmp (db->path, path) == 0)
             break;
 
     if (!db)
@@ -502,7 +502,7 @@ DBusHandlerResult avahi_service_type_browser_event (AvahiClient *client, AvahiBr
         goto fail;
 
     for (b = client->service_type_browsers; b; b = b->service_type_browsers_next)
-        if (strcmp (b->path, path) == 0)
+        if (b->path && strcmp (b->path, path) == 0)
             break;
 
     if (!b)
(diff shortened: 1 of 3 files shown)


lathiat:
@MarquisdeGeek while this change looks OK, I am curious in what specific case you hit this issue and which AvahiPoll implementation are you using?

Generally right now avahi-client is not thread safe and should not be used from multiple threads out of the box.  You can use it in multiple threads by using AvahiThreadedPoll and taking a lock on all access.

There was documentation on how to do this in the old trac website, I've just committed that to git into the files docs/THREADS (1cc2b8e8d62e939b8bd683f795794878863931af)

MarquisdeGeek:
IIRC, it was triggered by dbus, and about 1 in every 100-200 messages.

pemensik (commented):
It seems to me this is fixing the issue on the wrong place. ``avahi_service_resolver_new()`` appends new service_resolver to the list on the start. Then it starts actually initializing that structure. I think more effective would be moving resolver appending to the end of structure initialization. Add it to the list *only* when all structure members have been obtained. If there is a hidden reason for current style, I do not see it.

evverx:
I think this PR can be closed.

https://github.com/lathiat/avahi/commit/1cc2b8e8d62e939b8bd683f795794878863931af should hopefully help to figure out how to use it with more than one thread. If it segfaults when that documentation is followed could you open a new issue and attach the code that can trigger those crashes?

PR #173, opened 2018-03-26
[Closed] D-Bus API Race

simoninator:
This is a possible fix to the issue #9 .

### DIFF
--- a/avahi-core/browse-domain.c
+++ b/avahi-core/browse-domain.c
@@ -233,3 +233,10 @@ void avahi_s_domain_browser_free(AvahiSDomainBrowser *b) {
 
     avahi_free(b);
 }
+
+void avahi_s_domain_browser_repeat_items(AvahiSDomainBrowser *b) {
+    assert(b);
+
+    if(b->record_browser)
+        avahi_s_record_browser_restart(b->record_browser);
+}
--- a/avahi-core/browse-service-type.c
+++ b/avahi-core/browse-service-type.c
@@ -154,4 +154,11 @@ void avahi_s_service_type_browser_free(AvahiSServiceTypeBrowser *b) {
     avahi_free(b);
 }
 
+void avahi_s_service_type_browser_repeat_items(AvahiSServiceTypeBrowser *b) {
+	assert(b);
+
+	if(b->record_browser)
+		avahi_s_record_browser_restart(b->record_browser);
+}
+
 
--- a/avahi-core/browse-service.c
+++ b/avahi-core/browse-service.c
@@ -165,3 +165,11 @@ void avahi_s_service_browser_free(AvahiSServiceBrowser *b) {
     avahi_free(b->service_type);
     avahi_free(b);
 }
+
+void avahi_s_service_browser_repeat_items(AvahiSServiceBrowser *b)
+{
+    assert(b);
+
+	if(b->record_browser)
+		avahi_s_record_browser_restart(b->record_browser);
+}
--- a/avahi-core/browse.c
+++ b/avahi-core/browse.c
@@ -577,6 +577,12 @@ void avahi_s_record_browser_free(AvahiSRecordBrowser *b) {
     browser_cancel(b);
 }
 
+void avahi_s_record_browser_repeat_items(AvahiSRecordBrowser *b) {
+    assert(b);
+
+    avahi_s_record_browser_restart(b);
+}
+
 void avahi_s_record_browser_destroy(AvahiSRecordBrowser *b) {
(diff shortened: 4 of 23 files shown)


r-mike:
Please merge, really need this!

simoninator:
@lathiat  could you please review this pull request. Is there any chance this will be merged?

lathiat:
Repository is not abandoned, but I have not committed time to merge things recently.  I am planning to spend some time getting things merged and resolved in the next couple of weeks.

Thanks for creating a pull request to solve this issue I really appreciate it!

I am thinking that adding a repeat option is not the best fix since it is likely to be vulnerable to race conditions where the same service is repeated based on the timing of when the repeat command is sent versus when the objects finish streaming in.

My thoughts for a proper fix are to either

 (a) change the API version and add an explicit 'start' command rather than a repeat command (or perhaps, we would first create a service resolver without any parameters and then start it and provide the parameters then).  Obviously the old api version needs to stay in place for existing clients and automatically send items.

 (b) I need to look at the d-bus docs but I do wonder if it is possible to detect once the client has bound to the object and then start sending things automatically.


Do you or anyone else have any thoughts about any of the 3 methods (yours and whether my race condition concern is valid, and the two I suggested above).

simoninator:
It is true, there would be an evil race condition. You would receive two AllForNow signals without being able to assign them to the initial object notification or to the repeat command.  I will also spend some time looking at the dbus docs, maybe there is an easy solution.

smcv:
> (a) change the API version and add an explicit 'start' command rather than a repeat command (or perhaps, we would first create a service resolver without any parameters and then start it and provide the parameters then). Obviously the old api version needs to stay in place for existing clients and automatically send items.

This is viable, and is probably the best option if each object (e.g. service resolver) is only expected to be used by a single client. The "modern" API pattern would look like:

```
     CLIENT                      DBUS-DAEMON       AVAHI

CreateWhateverBrowser2 ------------------------->\
                      <--------------------------/ OK, created /browsers/b123
AddMatch(... /browsers/b123 ... ) ->\
                      <-------------/ OK, added match rule
Start()               -------------------------->\
                      <--------------------------/ OK
                                             <---- ItemNew(alpha) ---->
                                             <---- ItemNew(beta) ----->
                                             <---- ItemNew(gamma) ---->
                                             <---- AllForNow() ------->
                                                ... time passes ...
                                             <---- ItemNew(delta) ---->
```

> (b) I need to look at the d-bus docs but I do wonder if it is possible to detect once the client has bound to the object and then start sending things automatically.

This is not possible. A D-Bus client does not send you any messages unless that is done explicitly.

There is an option (c), which is a variation on what @simoninator proposed: provide a method call that lists everything that was already emitted in signals by this same object (the conventional jargon term is "state recovery"), perhaps something like

```
GetItems() -> (
    array of structs each with the same fields as ItemNew: items_emitted_already,
    boolean: has_emitted_all_for_now
)
```

If each object is expected to be used by a single client, then (a) is probably the better option.

If objects are expected to be shared between clients, then (a) can't be used and it'll have to be (c). A method call is a lot more conventional than re-emitting the signals, and IMO less confusing: the client-side subtlety would be that the client needs to connect to signals before it calls the method, and ignore any signals that come in before it receives the method reply (which is normal for D-Bus APIs, in particular the standard o.fd.Properties).

simoninator:
My understanding is that objects won‘t be shared. Whats the boss‘ opinion @lathiat? Tomorrow I‘ll have a look on how to implement both options (a) and the proposed method variation (c)

lathiat:
Yep no sharing

simoninator:
I implemented option (a), like @smcv suggested in the sequence diagram, but I'm not sure yet on what is the best way to change the API version and keep the downwards compatibility?

smcv:
> I'm not sure yet on what is the best way to change the API version and keep the downwards compatibility?

For each D-Bus object that had the design issue of starting to emit signals as soon as it's created:
* Add a new method on the Server, ServiceBrowserNew2 or similar, which creates the object but doesn't start using it
* Add a new method on the object that gets created, Start or similar, which starts doing the work and emitting the signals
* Make the old method, e.g. ServiceBrowserNew, equivalent to calling ServiceBrowserNew2 and immediately calling Start on the result (I expect 90% of the code can be shared)

API users that want to remain backwards-compatible with older Avahi can call ServiceBrowserNew2, if it succeeds carry on with the new code path, and if it fails call ServiceBrowserNew instead.

simoninator:
Done, see new pull request #175

PR #175, opened 2018-04-05
New D-Bus API

simoninator:
This pull request implements the API change mentioned in #173

### DIFF
--- a/avahi-common/dbus.h
+++ b/avahi-common/dbus.h
@@ -29,6 +29,7 @@ AVAHI_C_DECL_BEGIN
 
 #define AVAHI_DBUS_NAME "org.freedesktop.Avahi"
 #define AVAHI_DBUS_INTERFACE_SERVER AVAHI_DBUS_NAME".Server"
+#define AVAHI_DBUS_INTERFACE_SERVER2 AVAHI_DBUS_NAME".Server2"
 #define AVAHI_DBUS_PATH_SERVER "/"
 #define AVAHI_DBUS_INTERFACE_ENTRY_GROUP AVAHI_DBUS_NAME".EntryGroup"
 #define AVAHI_DBUS_INTERFACE_DOMAIN_BROWSER AVAHI_DBUS_NAME".DomainBrowser"
@@ -45,8 +46,9 @@ number is increased the API has been extended but is downwards
 compatible. If the release changes compatibility is lost.
 
 Avahi 0.6 implements API version 0x0201;
-Avahi 0.6.1 implements API version 0x0202 */
-#define AVAHI_DBUS_API_VERSION ((uint32_t) 0x0203)
+Avahi 0.6.1 implements API version 0x0202
+Avahi 0.7 implements API version 0x0203 */
+#define AVAHI_DBUS_API_VERSION ((uint32_t) 0x0204)
 
 #define AVAHI_DBUS_ERR_OK "org.freedesktop.Avahi.Success"
 #define AVAHI_DBUS_ERR_FAILURE "org.freedesktop.Avahi.Failure"
--- a/avahi-core/avahi-test.c
+++ b/avahi-core/avahi-test.c
@@ -366,7 +366,8 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     stb = avahi_s_service_type_browser_prepare(server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, NULL, 0, stb_callback, NULL);
     avahi_s_service_type_browser_start(stb);
 
-    sb = avahi_s_service_browser_new(server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, "_http._tcp", NULL, 0, sb_callback, NULL);
+    sb = avahi_s_service_browser_prepare(server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, "_http._tcp", NULL, 0, sb_callback, NULL);
+    avahi_s_service_browser_start(sb);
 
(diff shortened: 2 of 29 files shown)


r-mike:
nice work @simoninator ! Looking forward to see this getting merged.

simoninator:
Hey @lathiat,
when do you think you'll have the time to merge this pull request?

simoninator:
Hey @lathiat,
any news on merging this pull request?

lathiat:
Planning to get on top of it this week

simoninator:
Probably this week?

simoninator:
Probably this week?

r-mike:
Any news about this?

simoninator:
Well, I think that this pull request is never going to be merged!

SilaUser:
@lathiat any news on finding some time to merge this pull request?

simoninator:
@lathiat, any news on merging this PR?

simoninator:
@lathiat could you find some time to have a look at this PR and maybe merge it?

lathiat:
I'd much prefer if we avoid tacking a "2" on the end of the function names (e.g. ServiceBrowserNew2). I've got a couple of ideas on how we can do that

The first thing that makes sense would be to use a versioned API. The following documents cover this topic:
http://0pointer.de/blog/projects/versioning-dbus.html
https://dbus.freedesktop.org/doc/dbus-api-design.html

Initially I got stuck on this point "Note, however, that supporting multiple interface versions simultaneously requires that object paths are versioned as well, so objects must not be put on the bus at the root path (‘/’). This is for technical reasons: signals sent from a D-Bus service have the originating service name overwritten by its unique name (e.g. com.example.MyService2 is overwritten by :1:15)."

I think? we can get away with this for our limited use case, because the interface name is always sent when calling ServiceBrowserNew and we don't need to persist what interface was used later. So at least for this specific change we can get away with it.

(for the below I specifically reference ServiceBrowserNew since that's the most common however obviously this also applies to a number of other Browsers and Resolvers so assume when I mention ServiceBrowser below that I am talking about all of them - I just found it helpful to give an explicit example) 

If it's too difficult or otherwise problematic to user the interface version; I think instead we can simply change the name instead. Instead of "ServiceBrowserNew2" we can use "ServiceBrowserPrepare" which then requires an explicit call to "Start" where as "ServiceBrowserNew" would retain it's existing behavior and automatically Start as part of the creation. Besides looking less silly having "2" in the new name used permanently, it also more clearly distinguishes the difference in behavior between the two. The only slightly problem is that clients supporting the new and old server APIs would need to handle calling Prepare and then falling back if it gets back a NoMethodError.

Lastly I considered how we can fix existing clients alongside providing a new v2 API for this. We should be able to defer the start of the resolvers on the server-side for a short time to give the client time to setup it's signal receiver. This should make clients that are currently likely to fail, less likely to fail. I did a few quick tests and the best case delay for the signal handlers to be registered seems to be around 0.6 milliseconds so I don't think we'd need to delay it by much - only a few ms to handle most cases. I'd probably want to re-test this on a loaded and slower system to make sure it's not orders of magnitudes out. I think we should do this regardless of whatever other API change we end up making.

If we're worried about issues caused by the reply delays (though if we only need a few ms I am not too worried about it - if we ended up needing 100s of ms it might be more of a concern) - we could probably only do the delay for the Browsers and skip it for the Resolvers since we have directly-reply dbus methods for ServiceResolve/HostNameResolve/AddressResolve and don't need registering to a new object path. We could also obviously just skip the delay when the new version API is used and/or when Start is called - just cancel the delay and start anyway.

Which brings me full circle to what I think might be the most ideal and backwards compatible solution

(1) Create a new Avahi server interface API version, org.freedesktop.Avahi2
(2) Keep the name ServiceBrowserNew on both interfaces.
- If they use org.freedesktop.Avahi then we maintain the existing behavior to automatically start the ServiceBrowser but we delay the start by perhaps 10ms (we can determine the most ideal value) to magically fix most existing clients in most situations - it's not an ideal fix but they were already broken so it's not any worse. To make life easy we can potentially support a specific call to Start to start it early.
- If they use org.freedesktop.Avahi2 then we don't auto-start at all,

Lastly we have a number of avahi-core users outside of avahi-daemon and the current change will silently break those apps in that the API isn't backwards compatible. That's probably not necessary -- similar to the first suggestion above we could add a new _prepare method to go with _start and keep _new with the existing behavior to just call _prepare and then _start. Even if we don't add _prepare to the client/dbus API.

Would love feedback/discussion on each of the ideas.

simoninator:
Thank you @lathiat for your response.

I also think a second interface is the best option we have, as it doesn't break existing behaviour.

In my opinion, I think we should change the method names from xBrowserNew to xBrowserPrepare in the new interface. The API change would then be recognized directly in the new dbus interface and there might be no confusion.

I haven't found the time to look for a suitable timer solution to delay the replies in the old interface. Is there an implementation of a timer available, that can be used to delay the callbacks in the event/poll loop?

simoninator:
@lathiat ping

simoninator:
@lathiat ping

simoninator:
@lathiat ping

simoninator:
@lathiat Could you please have a look at the latest commits of this pull request?

simoninator:
@lathiat Could you please have a look at the latest commits of this pull request?

lathiat:
As I read the current pull request, the general approach is

 - To add new "Prepare" and "Start" methods for most services on both the avahi-core and avahi-client APIs when using the org.freedesktop.Avahi2 interface
 - Re-implement the existing "New" methods on the old org.freedesktop.Avahi interface to call Prepare and then Start immediately
 - Implement a 10ms delay on the Start for DBUS API users to attempt to silently fix old clients

This all looks good to me, I will check a couple extra details and do some testing next week but should more or less be OK.

lathiat:
Will need to bump the minor version numbers for the libraries in configure.ac but I can take care of that. In theory since we keep the old New function we shouldn't have to bump the major number.

simoninator:
@lathiat did you find some time to check the PR?

PR #178, opened 2018-04-27
avahi-python: Encode unicode strings as UTF-8

smcv:
Previously, we would effectively encode anything representable in
Latin-1 as Latin-1, and crash on anything not representable in Latin-1:
    
    >>> import avahi
    >>> avahi.string_to_byte_array(u'©')
    [dbus.Byte(169)]
    >>> avahi.string_to_byte_array(u'\ufeff')
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/dist-packages/avahi/__init__.py", line 94, in string_to_byte_array
        r.append(dbus.Byte(ord(c)))
    ValueError: Integer outside range 0-255
    
This is particularly important for Python 3, where the str type
is a Unicode string.
    
The b'' syntax for bytestrings is supported since at least Python 2.7.
    
These functions now accept either Unicode strings (Python 2 unicode,
Python 3 str), which are encoded in UTF-8, or bytestrings
(Python 2 str, Python 3 bytes) which are taken as-is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ ltmain.sh
 missing
 py-compile
 stamp-h1
+test-driver
--- a/avahi-python/avahi/.gitignore
+++ b/avahi-python/avahi/.gitignore
@@ -1,1 +1,6 @@
+*.log
+*.pyc
+*.pyo
+*.trs
 ServiceTypeDatabase.py
+__pycache__/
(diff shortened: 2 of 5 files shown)


smcv:
> Unicode strings (Python 2 unicode, Python 3 str), which are encoded in UTF-8

Justification for this arbitrary choice of encoding: RFC 6763 says "Often the value for a particular attribute will be US-ASCII [RFC20] or UTF-8 [RFC3629] text".

PR #184, opened 2018-05-31
replaced 2.x has_key with 3.x friendly in

dsnowb:
Howdy! I'm new to contributions, so apologies in advance if I'm missing any appropriate protocol. I ran into this incompatibility issue trying to run avahi-discover on my Arch system, and thought I'd make the change and see if you guys would like it. Anyway, love this tool. Best and thanks!


javiromero:
Thanks for the patch! I'm also having this issue on Arch and looking at the PRs I saw https://github.com/lathiat/avahi/pull/146 from last year addressing this same issue, so it would be nice if any of both could be merged :pray:

albertvaka:
Hi @lathiat, can this get merged or is there any blocker?

This works both in python2 (down to 2.3) and python3, while the current code doesn't work in python3.

PR #185, opened 2018-06-22
build: Attempt GDBM 1.15 compatibility

heftig:
Pretty hacky attempt at fixing the build with GDBM 1.15. Not sure I'm happy with this, but there you go.


heftig:
Now our Python with GDBM 1.16 generates `.db` instead of `.db.{pag,dir}` again; I have no idea what's going on.

PR #190, opened 2018-07-10
avahi_service_resolver_new: more documentation

orbisvicis:
Fully describe the behavior of browser, resolver and associated
callbacks with regard to query and address protocols.


lathiat:
Thanks!

PR #195, opened 2018-08-24
Dot on the wrong side of 'local'

destruktiw:
The correct domain name for local queries is `local` or `local.` (second has probably better performance), but `.local` is incorrect.

Avahi throws the following error when it is used:
`avahi_service_browser_new() failed: Invalid domain name`


lathiat:
Thanks for the suggestion apologies for the delay in merging but I've pushed this change.

PR #198, opened 2018-09-27
Updated Czech translation

Marv-CZ:


pemensik:
Is there a reason, why is this change waiting for 0.9 release? It seems improving current situation.

pemensik:
Would you be able to update this request, @Marv-CZ ? It cannot be merged as it is now.

PR #202, opened 2018-10-25
Two small fixes to issues discovered by static analysis

msekletar:


msekletar:
@lathiat it would be awesome if you could review this. Should be very quick. Thanks!

avahi-daemon/chroot.c
```diff
@@ -188,8 +188,8 @@ static int recv_fd(int fd) {
             return -1;
         }
 
-        assert(h->cmsg_len = CMSG_LEN(sizeof(int)));
-        assert(h->cmsg_level = SOL_SOCKET);
+        assert(h->cmsg_len == CMSG_LEN(sizeof(int)));
+        assert(h->cmsg_level == SOL_SOCKET);
```
pemensik, line 192: This part has been already proposed in other PR. Have to find its ID.

avahi-client/browser.c
```diff
@@ -72,6 +72,8 @@ static void parse_domain_file(AvahiDomainBrowser *b) {
         if (avahi_normalize_name(buf, domain, sizeof(domain)))
             b->static_browse_domains = avahi_string_list_add(b->static_browse_domains, domain);
     }
+
+    fclose(f);
```
pemensik, line 76: Thanks, definitely needed here!

pemensik (approved):

avahi-daemon/chroot.c
pemensik: This is also fixed in PR #372 .

msekletar: So let's close #372 as a duplicate of this one?

pemensik (approved):

evverx:
I think Coverity is right here but to be honest I'm not sure why the assertions are used there in the first place. They can be compiled out when `NDEBUG` is defined and looking at https://github.com/lathiat/avahi/pull/17 it appears `NDEBUG` is actually used to avoid a bunch of debug messages.

Issue #169, opened 2018-03-01
avahi-browse prints invalid UTF-8 for BluetoothAddress property sent by Apple TV 


mpontillo:
A bug was reported against our project (MAAS), which parses the output of `avahi-browse --all --resolve --no-db-lookup --parsable --no-fail`. In the bug, the following error occurred:

    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc8 in position 9122: invalid continuation byte

After triaging the output, I determined that the following portion of the output was the problem:
```

>>> hex(9122)
'0x23a2'


$ xxd avahi.log 
...
00002280: 443a 3432 3a45 333a 3039 220a 3d3b 656e  D:42:E3:09".=;en
00002290: 7333 333b 4950 7636 3b41 7070 6c65 5c30  s33;IPv6;Apple\0
000022a0: 3332 5456 3b5f 6d65 6469 6172 656d 6f74  32TV;_mediaremot
000022b0: 6574 762e 5f74 6370 3b6c 6f63 616c 3b41  etv._tcp;local;A
000022c0: 7070 6c65 2d54 562e 6c6f 6361 6c3b 3139  pple-TV.local;19
000022d0: 322e 3136 382e 312e 3831 3b34 3931 3630  2.168.1.81;49160
000022e0: 3b22 4c6f 6361 6c41 6972 506c 6179 5265  ;"LocalAirPlayRe
000022f0: 6365 6976 6572 5061 6972 696e 6749 6465  ceiverPairingIde
00002300: 6e74 6974 793d 3544 4239 3132 3037 2d44  ntity=5DB91207-D
00002310: 3333 362d 3445 4542 2d41 3534 332d 3538  336-4EEB-A543-58
00002320: 3131 4541 4237 3636 4643 2220 2253 7973  11EAB766FC" "Sys
00002330: 7465 6d42 7569 6c64 5665 7273 696f 6e3d  temBuildVersion=
00002340: 3135 4b36 3030 2220 2255 6e69 7175 6549  15K600" "UniqueI
00002350: 6465 6e74 6966 6965 723d 3744 4633 4241  dentifier=7DF3BA
00002360: 4637 2d33 3238 462d 3431 3645 2d42 4330  F7-328F-416E-BC0
00002370: 322d 4135 3636 3345 3637 3342 4238 2220  2-A5663E673BB8" 
00002380: 224e 616d 653d 4170 706c 6520 5456 2220  "Name=Apple TV" 
[...]

Issue #144, opened 2017-09-03
avahi-browse parsable format does not escape TXT strings


marvelous:
Running the following command containing a newline in arguments:
```
avahi-publish --service myname _mytype._tcp 42 '"
=;virbr0;IPv4;myname;_mytype._tcp;local;myhostname.local;8.8.8.8;42;"'
```
and browsing the result with
```
avahi-browse --resolve --parsable _mytype._tcp
```
shows a spurious line in `avahi-browse` output.

I don't know if the TXT record is non-standard-compliant and should be dropped by the responder and rejected/escaped by avahi-publish, or if the TXT record is standard-compliant and should be escaped by avahi-browse.

lathiat:
Fixed by #206

PR #206, opened 2018-12-02
avahi-common: Repair avahi_string_list_to_string to escape embedded quotes and backslashes.

tonyg:
This PR addresses issues #169 and #144, by causing `avahi_string_list_to_string` to escape double-quotes and backslashes with a leading backslash and to escape control bytes using a decimal representation.

I (quickly) checked callers of `avahi_string_list_to_string`, and it looks like in all contexts producing escaped output is a reasonable thing to do.

I have run `strlst-test` with valgrind, which reported no problems.

The patch uses *decimal* escapes (just like escaped labels), using them for control bytes strictly less than 32. This lets naive parsers of `avahi-browse -p` output work line-by-line, and allows correct representation of e.g. embedded NUL bytes.

| input                           |  buggy current output (unpatched) | output (this patch) |
| ----- | ------------------ | ------------------- |
| `"hello world"` | `"\"hello world\""` | `"\"hello world\""` |
| `"a\"b\nc\0d\ae"` | `"\"a\"b\nc\""` | `"\"a\\\"b\\010c\\000d\\007e\""` |

### DIFF
--- a/avahi-common/strlst-test.c
+++ b/avahi-common/strlst-test.c
@@ -48,6 +48,7 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     a = avahi_string_list_add_pair(a, "blubb", "blaa");
     a = avahi_string_list_add_pair(a, "uxknurz", NULL);
     a = avahi_string_list_add_pair_arbitrary(a, "uxknurz2", (const uint8_t*) "blafasel\0oerks", 14);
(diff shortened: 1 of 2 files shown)


tonyg:
On second thought, line-by-line processing of `avahi-browse` output is so useful that I've revised the patch to escape bytes less than 32, as well, using the decimal escaping convention familiar from escaped labels.

lathiat:
Thankyou!

Issue #167, opened 2018-02-07
s/UNCOMMITED/UNCOMMITTED/g


peternewman:
As found by https://github.com/lucasdemarchi/codespell . I guess this may want backwards compatibility of enums etc?

warmsocks:
I ran codespell on the current master codebase, and it found other typos and misspellings too.  I'll submit a pull request with those fixed.

warmsocks:
#222 fixes this issue and a few dozen other minor English language issues, mostly in comments.

EskoDijk:
I reviewed #222 and provided review comments there.

evverx:
@jschleus sorry to bother you. I wonder if it would be possible to add `avahi` to Fossies to build codespell reports like https://fossies.org/linux/test/systemd-main.e30b4c1.230529.tar.gz/codespell.html? It should hopefully help to prevent PRs fixing the same typos.

evverx:
The https://github.com/lathiat/avahi/tree/master/specs folder should probably be excluded if possible.

I'm not sure what happens when repository moves. In theory GitHub should keep redirecting lathiat/avahi to the new place but I'm not sure if it can affect Fossies somehow.

jschleus:
> @jschleus sorry to bother you. I wonder if it would be possible to add `avahi` to Fossies to build codespell reports like https://fossies.org/linux/test/systemd-main.e30b4c1.230529.tar.gz/codespell.html? It should hopefully help to prevent PRs fixing the same typos.

@evverx No problem. The `Avahi` project contents page  can be found under https://fossies.org/linux/misc/avahi and the according codespell report (if available) under https://fossies.org/linux/misc/avahi/codespell.html.

Both version-independent URLs (the second one is not linked) should redirect always to the last release.

Perhaps more meaningful may be the continuously (!) updated codespell report for the Avahi GitHub "master" version which is available within a special restricted "test" folder that isn't really integrated into the standard Fossies services and should also not be accessible to search engines:

https://fossies.org/linux/test/avahi-master.tar.gz/codespell.html

That version independent URL hopefully always redirects to the report for the latest "master" version identified by the short commit ID and a year-month-day string (YYMMDD) representing the according git pull date (mostly = commit date). The refresh rate is currently half-hourly.

Although after a first review some obviously wrong matches ("False Positives" = FPs) are already filtered out (unsure I was about the variable name `calback``) please inform me if you find more of them so that I can force a new improved check if applicable.

[...]

PR #222, opened 2019-04-17
Correct spelling mistakes found by codespell.  Fixes lathiat/avahi #167

warmsocks:
Fixes issue #167.


specs/draft-cheshire-dnsext-dns-sd-02.txt
```diff
@@ -379,9 +379,9 @@ Internet Draft       DNS-Based Service Discovery      14th February 2004
    with the <Instance> portion, since it is allowed to contain any
    characters, including dots.
 
-   Any dots in the <Instance> portion should be escaped by preceeding
+   Any dots in the <Instance> portion should be escaped by preceding
```
EskoDijk, line 382: we can't go fixing any text in the 'specs' folder - these are external IETF specs for reference and need to be included verbatim.

avahi-gobject/ga-entry-group.h
```diff
@@ -32,7 +32,7 @@ typedef enum {
     GA_ENTRY_GROUP_STATE_UNCOMMITED = AVAHI_ENTRY_GROUP_UNCOMMITED,
     GA_ENTRY_GROUP_STATE_REGISTERING = AVAHI_ENTRY_GROUP_REGISTERING,
     GA_ENTRY_GROUP_STATE_ESTABLISHED = AVAHI_ENTRY_GROUP_ESTABLISHED,
-    GA_ENTRY_GROUP_STATE_COLLISTION = AVAHI_ENTRY_GROUP_COLLISION,
+    GA_ENTRY_GROUP_STATE_COLLISION = AVAHI_ENTRY_GROUP_COLLISION,
```
EskoDijk, line 35: Changing this variable name would potentially change the interface for the GObject binding to Avahi, I think. So it should not be part of this multi-file spelling correction effort - for this change, specifically any binding/side effects needs to be known and discussed prior to changing (potentially) the API here.

avahi-sharp/EntryGroup.cs
```diff
@@ -41,7 +41,7 @@ public enum PublishFlags {
     }
 
     public enum EntryGroupState {
-        Uncommited,
+        Uncommitted,
```
EskoDijk, line 44: (see also other comment on API change) This would also change (potentially) the API of current C# programs using Avahi, so would therefore need to have its own separate PR. (Maybe needs to be included in an API revision or so.)

EskoDijk (changes requested):
Good spelling contributions in here. Needs to be changed for the following:
* don't fix typos in variable names that are (potentially) used to provide an external API to Avahi. Changing would break the binding/interface. Best do this in separate PRs where impact can be considered. See separate comments on code lines.
* don't fix typos in 'specs' folder - these are verbatim included IETF specs.

pemensik:
I have rebase it to current master, which fixed some places. Reverted the specs fixes and added backward compatible definition. @EskoDijk, what do you think about modified version?

EskoDijk:
@pemensik Thanks, good backward compatible solution in the C code.
However there is still a change in the API for C# code (consists of the two changes in the `avahi-sharp` directory).  The enum item name is changed there in C# which causes any build using this library to break.
(C# doesn't have #define to fix that for us - although there are some tricks maybe to make it work see https://stackoverflow.com/questions/8043027/non-unique-enum-values )

So I would not include this C# change as part of a "typo fix clean sweep PR" like this one.

EskoDijk:
@pemensik Thanks, that solves it.

PR #224, opened 2019-05-04
Fix issue 117 "Spurious name conflicts" in case of reflector

Bischoff:
The reflector sends back the whole cache when it receives a query.

As a result, a machine named `somemachine.local` may receive its own cached information, think its name is duplicate (used by another machine), and change it unexpectedly to `somemachine-2.local` (more detail on issue #117).

This PR removes an old workaround supposed to fix known answer suppression (part of commit 4ff0807c04fcc239de52a793bceb88e7f3408f3f from 2005). However, I think the workaround was worse than the problem itself: at my place, this makes the reflector unusable.

It **does not** fix issue 117 in the general case.


Bischoff:
Sorry, it's more complicated than that, this patch is not correct, closing.

PR #226, opened 2019-05-10
Fix a crash when txt records has an empty value

krnowak:
That's a regression that came with 3b067e06b274 ("Add support for
binary values in TXT records in service files"), we dropped the NULL
check for the strlen call. Other txt record types already check for
NULL value in the decode functions.

Reported at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=882386


mtdcr:
This segfault is older than two years now. Is anybody going to merge this PR?

See also: #247

PR #229, opened 2019-05-24
Reconfirm

nkarstens:
Implements support for reconfirm operation in avahi.


nkarstens:
Any chance we can get this accepted into the next release? We've been using it successfully for a number of years. I'm not sure why the Travis CI build did not complete; based on the output it seems like it was unrelated to the change. According to https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit you need write access to the repo to manually restart the build.

gregshpit:
Hi,

Are you going to merge "reconfirm" fix into the master branch and release it ? We really need it asap.
Thanks.

Gregory

nkarstens:
Updated again to latest version of master

pemensik:
This seems to be change too advanced for my current code understanding. I would get back to it once we merged simpler issues first.

nkarstens:
If there is going to be a 0.9 release soon it would be great if we could get this in there as well. Thanks!

pemensik:
Interesting feature, thank you for the contribution.

I lack some explanation, why this change is needed and why it is a good thing. I think the section 10.4 is more about client application trying to use resources it found via DNS-SD. Like it would try to connect to printer queue and found that address does not work. I am not sure the intention of MDNS were to trigger series of 4 multicast queries.

Is there actual use case, which this change is trying to solve? I have not seen it explained.

pemensik:
Do I understand it correctly the core of this change is adding client API call ``avahi_reconfirm_record()``, which allows clients to indicate the record might be not up to date? All other changes are more or less preparation of this new ability?

nkarstens:
Thanks for your questions!

What we have found is that mDNS is good at service discovery, but it is not good at service "undiscovery" -- determining when a service is no longer available. The last paragraph of RFC 6761 section 10.4 describes this by pointing out that one way to do that is to periodically poll the service, or for the service to periodically announce itself, and both are an inefficient use of network resources.

The alternative described in 10.4 is that when a client fails to contact a service it can issue a query to determine if the service is still available. This is the "reconfirm" operation.

The nice thing is that the passive observation of failure (POOF) described in 10.5 means that one client performing a reconfirm can cause all other clients to learn of the unavailability of a given service.

Basically the way that we use this is if a client fails to connect to a service it issues a reconfirm and sets a timer. If the timer expires then the remote device has indicated that the service is still available and we re-attempt the connection. If we receive a callback indicating that the service has been removed before the timer expires, then we stop the client. This behavior matches the reconfirm API provided by the dns-sd API.

I think your assessment is accurate -- the end goal is to provide `avahi_reconfirm_record()` and most other changes support that.

evverx:
I haven't looked at the PR closely but I just built `avahi` with this PR applied and then ran
```sh
busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server ReconfirmRecord "iisqquay" 0 0 "" 0 0 0 0
```
`avahi` crashed with
```sh
0x00007ffff79cfe5c in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install dbus-libs-1.14.6-1.fc37.x86_64 expat-2.5.0-1.fc37.x86_64 glibc-2.36-9.fc37.x86_64 libcap-2.48-5.fc37.x86_64 libdaemon-0.14-24.fc37.x86_64 libgcc-12.3.1-1.fc37.x86_64 libzstd-1.5.5-1.fc37.x86_64 lz4-libs-1.9.4-1.fc37.x86_64 systemd-libs-251.14-2.fc37.x86_64 xz-libs-5.4.1-1.fc37.x86_64
(gdb) bt
#0  0x00007ffff79cfe5c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff797fa76 in raise () from /lib64/libc.so.6
#2  0x00007ffff79697fc in abort () from /lib64/libc.so.6
#3  0x00007ffff796971b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7978656 in __assert_fail () from /lib64/libc.so.6
#5  0x00007ffff7d7df46 in avahi_rdata_parse (record=0x6040000079d0, rdata=0x0, size=0) at dns.c:861
#6  0x000000000056cbaf in dbus_start_reconfirm (c=0x612000000040, m=0x610000000340, error=0x7ffff581b320) at dbus-protocol.c:1032
#7  0x0000000000564f8b in dbus_select_common_methods (c=0x612000000040, m=0x610000000340, userdata=0x0, iface=0x5bed60 <str> "org.freedesktop.Avahi.Server", error=0x7ffff581b320) at dbus-protocol.c:1103
#8  0x0000000000563cea in msg_server_impl (c=0x612000000040, m=0x610000000340, userdata=0x0) at dbus-protocol.c:1231
#9  0x00007ffff7ed0d89 in dbus_connection_dispatch () from /lib64/libdbus-1.so.3
#10 0x000000000059be25 in dispatch_timeout_callback (t=0x6060000006e0, userdata=0x603000000670) at ../avahi-common/dbus-watch-glue.c:105
#11 0x00007ffff7f9dac2 in timeout_callback (t=0x6060000006e0) at simple-watch.c:447
#12 0x00007ffff7f9c807 in avahi_simple_poll_dispatch (s=0x60e000000040) at simple-watch.c:563
#13 0x00007ffff7f9db39 in avahi_simple_poll_iterate (s=0x60e000000040, timeout=-1) at simple-watch.c:605
#14 0x000000000052caa5 in run_server (c=0xf7bd00 <config>) at main.c:1268
#15 0x000000000051ca7a in main (argc=3, argv=0x7fffffffe3c8) at main.c:1693
```

I wonder if this PR has been tested?

nkarstens:
Updated and re-tested on latest master.

Thanks @evverx for noticing that bug. It has been fixed on the latest branch. I added a check to `dbus_start_reconfirm()` to make sure that rdata buffer actually exists before I start to use it.

Do you think we need to do something similar in `avahi_dbus_msg_etnry_group_impl()` for the `AddRecord` method? It also calls `avahi_dbus_read_rdata()` there without validating the buffer afterwards. I tried using `busctl` like you did to get it to crash but I think it will take multiple calls (the first being to use `EntryGroupNew` to add an entry group, but I'm not that good at using `busctl`).

evverx:
> Do you think we need to do something similar in avahi_dbus_msg_etnry_group_impl() for the AddRecord method? It also calls avahi_dbus_read_rdata() there without validating the buffer afterwards. I tried using busctl like you did to get it to crash but I think it will take multiple calls (the first being to use EntryGroupNew to add an entry group, but I'm not that good at using busctl).

I think it should be fixed one way or another. It's tracked in https://github.com/lathiat/avahi/issues/452 (with an easy way to reproduce it locally). I'm trying to figure out how to catch this stuff with the CI. I opened a PR last December where I brought `dfuzzer` but it can't catch https://github.com/lathiat/avahi/issues/452.

nkarstens:
> > Do you think we need to do something similar in avahi_dbus_msg_etnry_group_impl() for the AddRecord method? It also calls avahi_dbus_read_rdata() there without validating the buffer afterwards. I tried using busctl like you did to get it to crash but I think it will take multiple calls (the first being to use EntryGroupNew to add an entry group, but I'm not that good at using busctl).
> 
> I think it should be fixed one way or another. It's tracked in #452 (with an easy way to reproduce it locally). I'm trying to figure out how to catch this stuff with the CI. I opened a PR last December where I brought `dfuzzer` but it can't catch #452.

Nice, it's great you noticed that. I'm really glad there has been so much work put into the project recently, thank you!

evverx:
> there has been so much work put into the project recently

avahi is still kind of dormant unfortunately and I think code reviews aren't as effective as they can be because without the CI (or more precisely with the CI covering only 10% https://coveralls.io/builds/60369370) it's all manual and error-prone (and I still run dfuzzer manually because to get it work here the build system should be fixed first).

That being said I think even though some PRs aren't merged (or reviewed) people backport them anyway and hopefully report any issues they have encountered so those PRs kind of get tested and this is helpful too.

Anyway dfuzzer seems to trigger a memory leak somewhere:
```sh
=================================================================
==23506==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 400 byte(s) in 10 object(s) allocated from:
    #0 0x55bd9c006ece in __interceptor_malloc (/avahi/avahi-daemon/.libs/avahi-daemon+0xf8ece) (BuildId: 7170793980f863c7fdca53757b8ca6f0c488e3ed)
    #1 0x7f4b22094dc3 in xmalloc /avahi/avahi-common/malloc.c:68:15
    #2 0x7f4b22094c0a in avahi_malloc /avahi/avahi-common/malloc.c:107:16
    #3 0x7f4b21f8ffc8 in avahi_new_internal /avahi/avahi-core/../avahi-common/malloc.h:50:12
    #4 0x7f4b21f9087a in avahi_record_new /avahi/avahi-core/rr.c:101:15
    #5 0x7f4b21f90c16 in avahi_record_new_full /avahi/avahi-core/rr.c:127:9
    #6 0x55bd9c090868 in dbus_start_reconfirm /avahi/avahi-daemon/dbus-protocol.c:1029:15
    #7 0x55bd9c08903b in dbus_select_common_methods /avahi/avahi-daemon/dbus-protocol.c:1103:16
    #8 0x55bd9c087f03 in msg_server_impl /avahi/avahi-daemon/dbus-protocol.c:1239:9
    #9 0x7f4b21d8a70b  (/lib/x86_64-linux-gnu/libdbus-1.so.3+0x2870b) (BuildId: c6439c653749c1fa9aa52f01199f5e157e0a5edd)
    #10 0x7ffd3f73b45f  ([stack]+0x1f45f)

Direct leak of 360 byte(s) in 9 object(s) allocated from:
    #0 0x55bd9c006ece in __interceptor_malloc (/avahi/avahi-daemon/.libs/avahi-daemon+0xf8ece) (BuildId: 7170793980f863c7fdca53757b8ca6f0c488e3ed)
    #1 0x7f4b22094dc3 in xmalloc /avahi/avahi-common/malloc.c:68:15
    #2 0x7f4b22094c0a in avahi_malloc /avahi/avahi-common/malloc.c:107:16
    #3 0x7f4b21f8ffc8 in avahi_new_internal /avahi/avahi-core/../avahi-common/malloc.h:50:12
    #4 0x7f4b21f9087a in avahi_record_new /avahi/avahi-core/rr.c:101:15
    #5 0x7f4b21f90c16 in avahi_record_new_full /avahi/avahi-core/rr.c:127:9
    #6 0x55bd9c090868 in dbus_start_reconfirm /avahi/avahi-daemon/dbus-protocol.c:1029:15
    #7 0x55bd9c08903b in dbus_select_common_methods /avahi/avahi-daemon/dbus-protocol.c:1103:16
    #8 0x55bd9c087e5d in msg_server_impl /avahi/avahi-daemon/dbus-protocol.c:1231:9
    #9 0x7f4b21d8a70b  (/lib/x86_64-linux-gnu/libdbus-1.so.3+0x2870b) (BuildId: c6439c653749c1fa9aa52f01199f5e157e0a5edd)
    #10 0x7ffd3f73b45f  ([stack]+0x1f45f)
```

evverx:
FWIW dfuzzer can be found at https://github.com/dbus-fuzzer/dfuzzer (I co-maintain it so if it fails to build or doesn't work just let me know. It was tailored to systemd and its use-after-frees :-) but it can be run against avahi too). The script that can build avahi with ASan can be found in https://github.com/lathiat/avahi/pull/425.

nkarstens:
@evverx This is my first time running LeakSanitizer; I'm having trouble getting it running and was wondering if you'd be able to help. This is what I am doing:

1) Clone https://github.com/nkarstens/avahi.git and check out the reconfirm branch.
2) Add https://github.com/evverx/avahi.git as another remote and merge in werror branch.
3) Edit .github/workflows/build.sh and add the following flags to autogen_args:
   `--disable-libevent --disable-qt5 --disable-mono --disable-monodoc`
4) Still in build.sh, add `strict-prototypes` to the list of warning to ignore for avahi-daemon.
5) Run the following to compile:
   `CC=clang ASAN_UBSAN=true ./.github/workflows/build.sh`
   Build is successful and all tests pass.
6) `sudo ./avahi-daemon/avahi-daemon --file=/etc/avahi/avahi-daemon.conf --no-chroot`
7) In a separate shell, run `./avahi-client/client-test`
8) `sudo ./avahi-daemon/avahi-daemon -k` to shut down the daemon

```
avahi-daemon 0.8 exiting
==90940==LeakSanitizer has encountered a fatal error.
==90940==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==90940==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

9) `sudo -i` to get an interactive shell so I can export vars
10) `export LSAN_OPTIONS=verbosity=1:log_threads=1`
11) Repeat steps 6-8

```
avahi-daemon 0.8 exiting
==91555==Could not attach to thread 91511 (errno 1).
==91555==Failed suspending threads.
==91511==LeakSanitizer has encountered a fatal error.
==91511==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==91511==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

evverx:
> sudo ./avahi-daemon/avahi-daemon --file=/etc/avahi/avahi-daemon.conf --no-chroot

Could you add `--no-drop-root`? I should have mentioned that LSan isn't compatible with those privilege shenanigans. Sorry!

I think it should make the `client-test/avahi-daemon` test work. `dfuzzer` expects `avahi` to be fully installed though because it can't find the `Reconfirm` method if the dbus daemon isn't aware of `org.freedesktop.Avahi.Server.xml`.

nkarstens:
@evverx OK, I think I got it figured out. Here is the diff from the previous version:

```
diff --git a/avahi-core/reconfirm.c b/avahi-core/reconfirm.c
index d7be8d5..7d8de80 100644
--- a/avahi-core/reconfirm.c
+++ b/avahi-core/reconfirm.c
@@ -108,8 +108,10 @@ int avahi_record_reconfirm(
     assert(server);
     assert(r);

-    if (!(avahi_s_record_browser_new(server, interface, protocol, r->key, 0, record_browser_callback, avahi_record_ref(r))))
+    if (!(avahi_s_record_browser_new(server, interface, protocol, r->key, 0, record_browser_callback, avahi_record_ref(r)))) {
+        avahi_record_unref(r);
         return 0;
+    }

     /* If reconfirming an SRV record then also reconfirm the associated PTR record */
     if (r->key->type == AVAHI_DNS_TYPE_SRV) {
```

This has been reflected in 29f2287f899704f1edc4e31fea385b28798b3225.

This was an error case: if `avahi_s_record_browser_new()` failed to create a new `AvahiSRecordBrowser` then `record_browser_callback()` would never execute, so the reference to the record would never decrement and the record would never be freed. Everything checks out now (i.e., both `client-test` and `dfuzzer` execute without any leaks).

For posterity, here is the `dfuzzer` command I used:

`dfuzzer -n org.freedesktop.Avahi -o / -i org.freedesktop.Avahi.Server -t ReconfirmRecord`

What are your thoughts on merging this?

evverx:
> What are your thoughts on merging this?

I think at this point it's safe to say that @pemensik is the most active maintainer so it's up to Petr in the end I believe. https://github.com/lathiat/avahi/milestone/4 is a priority now (though I think for example the valgrind false positive can be removed from there because the only issue there is that `musl` on Linux doesn't work properly with overly long interface names and should be gotten around in avahi). More generally I placed avahi on life support back in December but since the circumstances changed I'm in the process of reevaluating that. I'll either start maintaining it properly in Q3 or leave all that to Petr.

FWIW this PR was built successfully on various arches in https://github.com/evverx/avahi/pull/3.

Before I forget thanks for taking a look at https://github.com/lathiat/avahi/issues/463. It's much appreciated.

avahi-core/reconfirm.c
```diff
@@ -103,6 +103,8 @@ int avahi_record_reconfirm(
     AvahiProtocol protocol,
     AvahiRecord *r) {
 
+    AvahiRecord *r_assoc;
```
ceggers-arri, line 106: The scope of `r_assoc` can be reduced.

avahi-core/reconfirm.c
```diff
@@ -111,5 +113,30 @@ int avahi_record_reconfirm(
         return 0;
     }
 
+    /* If reconfirming an SRV record then also reconfirm the associated PTR record */
+    if (r->key->type == AVAHI_DNS_TYPE_SRV) {
```
ceggers-arri, line 117: For my application it is very useful that the PTR record is automatically removed. Even if I would browse the list with my application's service type manually from my application, I would not know which entries should be reconfirmed (as the record's data with the PTR destination is not exposed in clear text via D-Bus). But the devices on my network also register a service subtype, which is not covered by this.

Example:
- Host: myhost.local
- Type: _http._tcp
- Subtype: _mysub._sub._http._tcp

So after reconfirmation has removed the SRV entry, the second PTR record (linking the subtype to my service) still persists.

myservice._http_tcp.local  IN      SRV 0 0 80 myhost.local  <-- removed by reconfirmation
_http._tcp.local   IN      PTR myservice._http._tcp.local  <-- removed by reconfirmation
_mysub._sub._http._tcp.local IN      PTR myservice._http._tcp.local <-- NOT covered by reconfirmation

ceggers-arri (commented):
Hi Nate,

after reading [10.4 – Cache Flush on Failure Indication](https://datatracker.ietf.org/doc/html/rfc6762#section-10.4), I wondered whether Avahi has such a feature.

While your series in general works as expected, I am unsure how correctly get rid of ALL records (A/AAAA/TXT/SRV/PTR, ...) related to a stale host.

avahi-core/reconfirm.c
evverx: > as the record's data with the PTR destination is not exposed in clear text via D-Bus

It's exposed via D-Bus when clients subscribe to records with `RecordBrowser*`.

FWIW I'm not sure this API should automatically reconfirm RRs unless it's requested explicitly. There should probably be a high-level API that could purge all the RRs related to services.

Either way it's a good point and it would probably make sense to figure out how this API can be used.

avahi-daemon/dbus-protocol.c
```diff
@@ -997,6 +998,52 @@ static DBusHandlerResult dbus_prepare_record_browser_object(RecordBrowserInfo **
     return avahi_dbus_respond_path(c, m, i->path);
 }
 
+static DBusHandlerResult dbus_start_reconfirm(DBusConnection *c, DBusMessage *m, DBusError *error) {
+    int32_t interface, protocol;
+    uint32_t flags, size;
+    char *fullname;
```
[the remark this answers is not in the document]
evverx: Looks like `fullname` and the order of the arguments in general came from https://developer.apple.com/documentation/dnssd/1804726-dnsservicereconfirmrecord. It should be fine to use it in the compat library but the avahi part should be consistent with avahi indeed.

avahi-core/reconfirm.c
nkarstens: This seems to be a matter of preference and Avahi doesn't have consistency here. Going to leave as-is unless we can point to some project-policy on this (or that not fixing this is ultimately preventing the feature from merging).

avahi-core/reconfirm.c
nkarstens: The API is meant to reconfirm individual resource records. Some RRs are tied to services, others to devices. I agree with what @evverx mentioned about not doing this automatically. There could be some high-level API to handle this, but for now applications that take that on themselves. Generally-speaking, though, I think the idea behind DNS-SD is that you are working with _services_, not _devices_.

avahi-daemon/dbus-protocol.c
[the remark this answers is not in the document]
nkarstens: OK, that should be done now.

avahi-common/defs.h
```diff
@@ -309,6 +309,11 @@ typedef enum {
     AVAHI_DOMAIN_BROWSER_MAX
 } AvahiDomainBrowserType;
 
+/** Flags for reconfirming a record */
+typedef enum {
+    AVAHI_RECONFIRM_RECORD_NOFLAGS = 0      /**< Placeholder for future changes */
```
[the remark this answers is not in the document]
nkarstens: Done.

avahi-daemon/dbus-protocol.c
```diff
@@ -997,6 +998,52 @@ static DBusHandlerResult dbus_prepare_record_browser_object(RecordBrowserInfo **
...
+    }
+
+    result = avahi_record_reconfirm(avahi_server, (AvahiIfIndex) interface, (AvahiProtocol) protocol, r);
+
+    avahi_record_unref(r);
+
+    if (!result)
+        return avahi_dbus_respond_error(c, m, AVAHI_ERR_INVALID_RECORD, NULL);
+
+    return avahi_dbus_respond_int32(c, m, AVAHI_OK);
```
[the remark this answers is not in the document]
nkarstens: This now uses avahi_dbus_respond_ok() instead of returning an integer.

evverx:
@nkarstens I kind of started looking at how it's done with `dns-sd` and as far as I can see it comes with the `-C` option:
```sh
dns-sd -C <name> <rrtype> <rrclass>           (Query; reconfirming each result)
```
Looks like it works without `rdata` (and I think it makes sense if the idea is to get rid of stale RRs somewhat unconditionally). I wonder if you know what happens under the hood there? Does it go over the cache and look for the corresponding rdata or does it just fire a new query and use the response to update the cache?

More generally it seems to me it could be implemented by adding a lookup option named, say, `AVAHI_LOOKUP_RECONFIRM` that could be `or`ed with the other lookup flags. Without that flag avahi could work as usual but with that flag it could bypass the cache and update stale RRs along the way.

evverx:
(Just to clarify I'm not saying it should be rewritten. I'm just trying to figure out how this API can cover various use cases. It's all prompted by https://github.com/lathiat/avahi/pull/229#discussion_r1338720642)

ceggers-arri:
> The following session (with a stale A RR) leads to avahi sending four packets corresponding to the first `ReconfirmRecord` call and then it ignores the next `ReconfirmRecord` call. I'd expect it to send a query (or somehow indicate that whatever is reconfirmed isn't in the cache). I'm not sure the first call should end before all the four packets are sent and the stale RR is removed either.

I am unsure whether this is relevant here, but I noticed that reconfirmation doesn't work (in terms of stale entries are not removed) if I flood the Avahi daemon with equal reconfirm requests via D-Bus. After limiting the rate of reconfirmation requests sent by my application, it looks fine for me (tested with the patch series from 2023-11-14).

evverx:
Nice catch! I can confirm that something like
```sh
while :; do
    busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server ReconfirmRecord iisqquay -- 2 0 Evgenys-MacBook-Air.local 1 1 0 4 192 168 56 3
done
```
prevents avahi from ever removing the stale RR from the cache (it doesn't seem to auto-expire either. at least `avahi-resolve` running in parallel kept returning the same IP for more than 120 seconds).

nkarstens:
> @nkarstens I kind of started looking at how it's done with `dns-sd` and as far as I can see it comes with the `-C` option:
> 
> ```shell
> dns-sd -C <name> <rrtype> <rrclass>           (Query; reconfirming each result)
> ```
> 
> Looks like it works without `rdata` (and I think it makes sense if the idea is to get rid of stale RRs somewhat unconditionally). I wonder if you know what happens under the hood there? Does it go over the cache and look for the corresponding rdata or does it just fire a new query and use the response to update the cache?
> 
> More generally it seems to me it could be implemented by adding a lookup option named, say, `AVAHI_LOOKUP_RECONFIRM` that could be `or`ed with the other lookup flags. Without that flag avahi could work as usual but with that flag it could bypass the cache and update stale RRs along the way.

@evverx -- The `-C` option uses `DNSServiceQueryRecord` to query the record first and then calls `DNSServiceReconfirmRecord` with the result. Handling of the `-C` option starts at [dns-sd.c line 2256](https://github.com/apple-oss-distributions/mDNSResponder/blob/d5029b5dff8aa59d1fc07ed796e994106ef58dee/Clients/dns-sd.c#L2256). The callback passed into `DNSServiceQueryRecord` is `qr_reply`. Jump to [line 1253](https://github.com/apple-oss-distributions/mDNSResponder/blob/d5029b5dff8aa59d1fc07ed796e994106ef58dee/Clients/dns-sd.c#L1253) and you can see that it calls `DNSServiceReconfirmRecord` using the `rdata` passed in with the callback.

I don't think you can create a reconfirm API without using rdata because the RR may not be unique (i.e., there may be several RRs with the same name and they only differ in content). For example, if you are reconfirming a `PTR` record for `_http._tcp.local` then you'd need to indicate _which_ `PTR` record you are reconfirming.

evverx:
Thanks for looking into this!

> I don't think you can create a reconfirm API without using rdata because the RR may not be unique 

Agreed. The "low-level" API can't operate without rdata. Looks like the "high-level" API should keep track of rdata itself then. @ceggers-arri I wonder if it's suitable for your use case?

nkarstens:
> Nice catch! I can confirm that something like
> 
> ```shell
> while :; do
>     busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server ReconfirmRecord iisqquay -- 2 0 Evgenys-MacBook-Air.local 1 1 0 4 192 168 56 3
> done
> ```
> 
> prevents avahi from ever removing the stale RR from the cache (it doesn't seem to auto-expire either. at least `avahi-resolve` running in parallel kept returning the same IP for more than 120 seconds).

If this is a concern then it would probably be easy enough to fix by modifying `avahi_cache_entry_reconfirm` to check to see if we aren't already in a reconfirm state. What do you think?

evverx:
> if we aren't already in a reconfirm state. What do you think?

As far as I understand it would prevent all the clients (apart from the first one) from getting a meaningful response via D-Bus. I think all the "reconfirm" calls should coalesce by analogy with the usual queries.

evverx:
Speaking of D-Bus I have to admit I'm not sure what responses should look like. I think when I was experimenting with this patch I expected it to return when the RR was either reconfirmed or the whole call failed. It wasn't based on any actual use cases though so it's hard to tell whether that expectation makes any sense.

ceggers-arri:
> > I don't think you can create a reconfirm API without using rdata because the RR may not be unique
> 
> Agreed. The "low-level" API can't operate without rdata. Looks like the "high-level" API should keep track of rdata itself then. @ceggers-arri I wonder if it's suitable for your use case?

I have currently two use cases where I use reconfirm
1. When I get a timeout error from the ServiceResolver. Some mDNS records have a (quite) short TTL of 120 s. So if a host is removed from a network, the (cyclic) renewal of these records is often the first thing where the absence of a host is recognized by others. I am unsure whether keeping the ServiceResolver(s) active all the time is compliant to the mDNS specification, but this is how our system works currently.
2. When a connection to a (mDNS) device fails (TCP connect timeout).

I guess that the application should normally not be interested in the rdata.  But to get rid of specific (stale) PTR records, I currently have to investigate the rdata of the items returned by the RecordBrowser in order to determine which records may be related to the stale host.

From my perspective, not requiring a RecordBrowser would be nice for a higher level interface. I would prefer calling Reconfirm only with information I already have (returned by the ServiceBrowser/-Resolver). Currently I have to create RecordBrowsers for PTR, TXT and SRV records. Requiring only a single function call to initiate reconfirmation of all records related to a specific host/service would be nice. If such an interface would not fit into Avahi/mDNS semantics, at least some documentation/example code for use cases like mine should be added.

nkarstens:
> Speaking of D-Bus I have to admit I'm not sure what responses should look like. I think when I was experimenting with this patch I expected it to return when the RR was either reconfirmed or the whole call failed. It wasn't based on any actual use cases though so it's hard to tell whether that expectation makes any sense.

The way mDNSResponder works is that you will get a callback if the service is removed (just like if the service was gracefully removed). When calling the reconfirm API you should also set up a timer. If the timer expires before you get a callback then you can try contacting the service again. If you get a callback removing the service then you know the reconfirm operation failed and the service is no longer available.

nkarstens:
> > > I don't think you can create a reconfirm API without using rdata because the RR may not be unique
> > 
> > 
> > Agreed. The "low-level" API can't operate without rdata. Looks like the "high-level" API should keep track of rdata itself then. @ceggers-arri I wonder if it's suitable for your use case?
> 
> I have currently two use cases where I use reconfirm
> 
>     1. When I get a timeout error from the ServiceResolver. Some mDNS records have a (quite) short TTL of 120 s. So if a host is removed from a network, the (cyclic) renewal of these records is often the first thing where the absence of a host is recognized by others. I am unsure whether keeping the ServiceResolver(s) active all the time is compliant to the mDNS specification, but this is how our system works currently.

Keeping the service resolver active all of the time is a "continuous query" and is compatible with the specification. However, I don't think you need to use the reconfirm operation here because the timeout means that the service has already been removed.

>     2. When a connection to a (mDNS) device fails (TCP connect timeout).

This is the traditional use for the reconfirm API.

> I guess that the application should normally not be interested in the rdata. But to get rid of specific (stale) PTR records, I currently have to investigate the rdata of the items returned by the RecordBrowser in order to determine which records may be related to the stale host.
> 
> From my perspective, not requiring a RecordBrowser would be nice for a higher level interface. I would prefer calling Reconfirm only with information I already have (returned by the ServiceBrowser/-Resolver). Currently I have to create RecordBrowsers for PTR, TXT and SRV records. Requiring only a single function call to initiate reconfirmation of all records related to a specific host/service would be nice. If such an interface would not fit into Avahi/mDNS semantics, at least some documentation/example code for use cases like mine should be added.

One thing to keep in mind is that mDNS is meant for _service_ discovery, not _host_ discovery. You shouldn't need to work directly with host RRs like `A` or `AAAA`.

You also shouldn't need to worry about removing _all_ RRs related to a service. When a new service is resolved, just create a record browser for the `SRV` record associated with the service. If there is an issue contacting that service then use the `rdata` to reconfirm that service. No need to worry about `PTR` or `TXT` RRs.

mvduin:
> Keeping the service resolver active all of the time is a "continuous query" and is compatible with the specification. However, I don't think you need to use the reconfirm operation here because the timeout means that the service has already been removed.

You'd want to reconfirm the record that keeps it showing up in the ServiceBrowser reconfirmed, i.e. the PTR from service type to service name (which typically has a long TTL, avahi uses 75 minutes). Otherwise you may be force to keep retrying resolving the service since it continues to be listed by the browser.

mvduin:
> One thing to keep in mind is that mDNS is meant for _service_ discovery, not _host_ discovery. You shouldn't need to work directly with host RRs like `A` or `AAAA`.

I'm not sure what you're trying to say here, obviously it is useless to discover a service unless you can resolve it all the way down to the IP address(es) to connect to.

I don't think anyone specifically _wants_ to work with A/AAAA records manually through RecordBrowsers, that's just to work around API limitations such as lack of reconfirm and/or lack of a continuous version of ResolveService (but the latter is only really needed as workaround for the former I think).

nkarstens:
> > One thing to keep in mind is that mDNS is meant for _service_ discovery, not _host_ discovery. You shouldn't need to work directly with host RRs like `A` or `AAAA`.
> 
> I'm not sure what you're trying to say here, obviously it is useless to discover a service unless you can resolve it all the way down to the IP address(es) to connect to.
> 
> I don't think anyone specifically _wants_ to work with A/AAAA records manually through RecordBrowsers, that's just to work around API limitations such as lack of reconfirm and/or lack of a continuous version of ResolveService (but the latter is only really needed as workaround for the former I think).

Based on [this comment](https://github.com/avahi/avahi/pull/229#pullrequestreview-1646545472) from @ceggers-arri, he was trying to remove all records related to a host. I was pointing out that this shouldn't be necessary, you can just reconfirm the `SRV` record and if the reconfirm fails then it should be removed by the service resolver. Using the `PTR` record would probably work as well.

evverx:
It doesn't cover subtypes as far as I can remember and that's where that discussion started I think. The question boils down to what avahi should reconfirm automatically. It it was really "low-level" it wouldn't touch PTR records and let clients deal with that but it covers that so it's kind of reasonable to assume that it can provide some knobs to purge more than just one RR.

ceggers-arri:
> > > One thing to keep in mind is that mDNS is meant for _service_ discovery, not _host_ discovery. You shouldn't need to work directly with host RRs like `A` or `AAAA`.
> > 
> > 
> > I'm not sure what you're trying to say here, obviously it is useless to discover a service unless you can resolve it all the way down to the IP address(es) to connect to.

I confirm that I directly use the IP addresses reported by the ServiceResolver for contacting other hosts (in contrast to using name resolution by the C library).

> Based on [this comment](https://github.com/avahi/avahi/pull/229#pullrequestreview-1646545472) from @ceggers-arri, he was trying to remove all records related to a host. I was pointing out that this shouldn't be necessary, you can just reconfirm the `SRV` record and if the reconfirm fails then it should be removed by the service resolver. Using the `PTR` record would probably work as well.

Sorry, I forgot to mention that my service(s) announce a "service subtype". On the "browser side", I have one ServiceBrowser for each _service subtype_ I am interested in (not for _services_). The subtypes are expressed as additional `PTR` records which are NOT automatically reconfirmed in the current implementation. In order to get rid of stale service (subtypes) in my ServiceBrowser, I have to "manually" reconfirm them (only `PTR` records of _services_ are automatically reconfirmed, not for _service subtypes_). This is also the reason why I have to filter the RRs returned by the RecordBrowser by their rdata. I only want to reconfirm `PTR` records pointing to the service/host in question.

Sorry for (possibly) not using the correct wording.

ceggers-arri:
> You also shouldn't need to worry about removing _all_ RRs related to a service. When a new service is resolved, just create a record browser for the `SRV` record associated with the service. If there is an issue contacting that service then use the `rdata` to reconfirm that service. No need to worry about `PTR` or `TXT` RRs.

Probably there is no need to take care about stale `TXT` RRs. But when I get timeout errors from the ServiceResolver (because short living `A`/`AAAA` RRs couldn't be refreshed), I use this as an opportunity to reconfirm long living `SRV`/`PTR` entries. My goal is to remove stale hosts from the ServiceBrowser result ideally **before** user tries to connect to them.

nkarstens:
> It doesn't cover subtypes as far as I can remember and that's where that discussion started I think. The question boils down to what avahi should reconfirm automatically. It it was really "low-level" it wouldn't touch PTR records and let clients deal with that but it covers that so it's kind of reasonable to assume that it can provide some knobs to purge more than just one RR.

Is reconfirming these ancillary records a requirement for accepting this pull request or can we view that as a future improvement?

evverx:
I wouldn't say it's a requirement but since it's a public API with all the backward compatibility guarantees it entails it would be great to come up with something that can cover as many use cases as possible (or can be extended without introducing `ReconfirmRecord2` that does the same thing a bit differently or something like that).

ceggers-arri:
> Is reconfirming these ancillary records a requirement for accepting this pull request or can we view that as a future improvement?

At least for my application it is ok as it is (fixing the problem with "periodic" ReconfirmRecord requests would be nice). In general I have no problem using the proposed "low level" API, but for me it took some time until I understood how to use ReconfirmRecord at all. I never had any use for a RecordBrowser before, so this was not obvious for me.

nkarstens:
> I wouldn't say it's a requirement but since it's a public API with all the backward compatibility guarantees it entails it would be great to come up with something that can cover as many use cases as possible (or can be extended without introducing `ReconfirmRecord2` that does the same thing a bit differently or something like that).

I think you could add that functionality in the future without changing the public API. You would just add logic similar to what this pull request is doing as far as locating other RRs to query. Any client that manually implemented reconfirming additional RRs in the interim would continue to function as before, there would just be multiple sources for reconfirming the additional RRs (the client and internal avahi).

nkarstens:
> > Is reconfirming these ancillary records a requirement for accepting this pull request or can we view that as a future improvement?
> 
> At least for my application it is ok as it is (fixing the problem with "periodic" ReconfirmRecord requests would be nice). In general I have no problem using the proposed "low level" API, but for me it took some time until I understood how to use ReconfirmRecord at all. I never had any use for a RecordBrowser before, so this was not obvious for me.

What would your proposal for a high-level API look like, and would it be able to account for RRs that are "shared" (not "unique" -- see RFC 6762 section 2)?

ceggers-arri:
> What would your proposal for a high-level API look like, and would it be able to account for RRs that are "shared" (not "unique" -- see RFC 6762 section 2)?

Probably I am not qualified answering this question.

At least for me, having an API for getting rid of all RRs related to a specific service would be nice (including any `PTR` records which comes from _service subtypes_). I also think that this is not possible with the existing API (I can only enumerate `PTR` records for known service subtypes, not for arbitrary ones).

e.g. `ReconfirmService(name, domain, flags)`.

Many Avahi APIs have additional `interface` and `protocol` parameters, maybe they should also be added for consistency.

Currently I have not the overview which RRs whether there are any shared RRs.

evverx:
> Any client that manually implemented reconfirming additional RRs in the interim would continue to function as before

Those clients can reconfirm PTR RRs themselves too so it shouldn't be necessary to special-case SRV RRs. If it wasn't there nobody would probably expect avahi to reconfirm anything else apart from exactly one RR. (I'll take a look at what the mDNS responder does. I didn't see any packets autoreconfirming PTRs along with SRVs but it could be I missed something).

It isn't obvious what `avahi_reconfirm_record` should return either. It should be something useful and it should be set in stone before it's released.

evverx:
Anyway stale RRs should be addressed one way or another and I think this PR is going in the right direction.

@pemensik I've just added it to the 0.9.1 milestone. if you have any concerns it would be great if you could comment here.

nkarstens:
> I will fix this and push a new patch set.

This issue with repeated requests to reconfirm the same RR resulting in that RR never being removed from the cache has been addressed. `avahi_cache_entry_reconfirm()` now calls `avahi_cache_entry_reconfirming()` to see if the RR is already being reconfirmed before restarting the timer.

evverx:
Thanks! I'll try to take a close look later (in 2024 probably). I'm not sure why it rejects `AVAHI_IF_UNSPEC`/`-1`:
```sh
$ busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server2 ResolveHostName "iisiu" -- -1 0 "host-3.local" 0 0               
iisisu 3 0 "host-3.local" 0 "192.168.56.3" 5
$ busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server ReconfirmRecord iisqquay -- -1 0 host-3.local 1 1 0 4 192 168 56 3
Call failed: Invalid interface index
```

(@nkarstens the CI isn't run automatically on PRs created by first-time contributors so to avoid waiting for me (or anyone else) to trigger the CI manually it should be possible to fix a typo from https://fossies.org/linux/test/avahi-master.tar.gz/codespell.html or something from https://scan.coverity.com/projects/avahi-daemon in a separate PR. I could merge that and all your force-pushes should start trigger the CI automatically)

evverx:
> One thing to keep in mind is that mDNS is meant for service discovery, not host discovery. You shouldn't need to work directly with host RRs like A or AAAA.

I've been thinking about this and I can't say I agree with that (It's true that unlike, say, LLMNR mDNS was designed with that use case in mind though). Other than that LLMNR and other things capable of resolving hostnames only are effectively being phased out so all that is slowly going to mDNS so use cases where it's desirable to get the latest A and AAAA RRs (and the corresponding PTR RRs) regardless of what avahi keeps in its cache are perfectly valid and it's possible that services aren't even involved there at all.

I think `ReconfirmRecord` itself should be sync by analogy with `ResolveHostName` and come with its async counterpart that can be used by the compat lib (or any other client where it should be async). That would make this API consistent with avahi. (anyway I can't say I spend a lot of time experimenting with this so I'm still not 100% what it should look like).

evverx:
Before I forget it seems in its current form "reconfirm" either does nothing if an RR with no matching rdata is in the cache or known-answer suppression kicks in for some reason with out-of-date rdata. The `"QM" question A 192.168.56.100` packets popped up when I was reconfirming `192.168.56.3` with `192.168.56.100` in the cache:
```sh
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.100 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.100 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
```

nkarstens:
> Before I forget it seems in its current form "reconfirm" either does nothing if an RR with no matching rdata is in the cache or known-answer suppression kicks in for some reason with out-of-date rdata. The `"QM" question A 192.168.56.100` packets popped up when I was reconfirming `192.168.56.3` with `192.168.56.100` in the cache:
> 
> ```shell
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.100 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.100 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
> MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
> MDNS 88 Standard query 0x0000 A host-3.local, "QM" question A 192.168.56.100
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 72 Standard query 0x0000 A host-3.local, "QM" question
> MDNS 183 Standard query response 0x0000 A, cache flush 192.168.56.3 NSEC, cache flush Z._qotd._tcp.local NSEC, cache flush host-3.local
> ```

The current behavior is inspired by mDNSResponder. What do you think the behavior should be? Returning more information would either cause the synchronous call to run longer, or we'd have to change the API to provide a callback. All things being equal I think keeping it synchronous is preferable.

I can't say I've seen the known-answer suppression issue before. Can you email me a .pcap file?

Neustradamus:
Dear Avahi team, @lathiat, @evverx, @pemensik,

We are now in 2025 (6 years, 1 month, 13 days already), any progress about @nkarstens PR?

PR #236, opened 2019-07-16
Added LLVM Fuzzer Support for Fuzzing Avahi when Receiving Packets via IPv6 and IPv4

hardikgoyal:
Wrote a fuzzer to send random packets to avahi to find if any combination can cause a crash. This is done using LLVM fuzzer support, where you can supply certain good inputs and it will then take them and modify it to see if it can find a crash. Should a crash be found, the fuzzer reports the crash along with the input that caused it. I can add documentations on how to build and run it, if it is necessary.


invidian:
@hardikgoyal can you add some explanation why you add it to PR description/commit message? Or maybe issue reference?

hardikgoyal:
Just added the description @invidian

evverx:
I'll go ahead and close this PR. The fuzzing infrastructure is in https://github.com/lathiat/avahi/tree/master/fuzz. It isn't integrated into the build system partly because the build system needs fixing first to get it to work with clang and ASan/UBSan/MSan. It's built and tested on OSS-Fuzz (and on CIFuzz on PRs) and can already be used to add more fuzz targets/improve the existing ones to bump the coverage. PRs are welcome. Thanks!

PR #249, opened 2019-09-17
pkgconfig: Match the first three parameters

neheb:
For consistency between projects. Might also fix several cross compilation
cases.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


neheb:
Hrm this is not right. libdir evaluates to /usr/lib , which is incorrect. libdir should be either lib or lib64 (or maybe lib32). Some extra changes are needed.

PR #255, opened 2019-12-03
Protect against NULL-pointer-deference with malformed txt records

mm120:
If the txt record is malformed, or is empty ("foo="), then we can hit
a NULL-pointer-de-reference.

Fixes: 3b067e06b274c60a9745bb45467e2eb0d9f3f0be

Signed-off-by: Mark Marshall <mark.marshall@omicronenergy.com>


lathiat:
Merged an alternate fix for this in https://github.com/lathiat/avahi/commit/4a6d6cc962d2c03cb5b40c02b92f04a8b1128009

PR #257, opened 2020-01-02
Avoid strlen(NULL) crash on some txt-records

andhe:
A record with only key= and empty value causes a crash, eg.
<txt-record>note=</txt-record>

http://bugs.debian.org/947891


andhe:
Of course I should have checked merge requests first.... This is the same problem as https://github.com/lathiat/avahi/pull/255 (with cosmetic differences).

Issue #267, opened 2020-02-25
fuzz test runtime error: left shift of 186 by 24 places cannot be represented in type 'int' in dns.c


traffic-millions:
the fuzz test err are as follows:
```
#3774   NEW    cov: 68 ft: 83 corp: 17/246b lim: 21 exec/s: 0 rss: 23Mb L: 17/21 MS: 1 ChangeBit-
dns.c:458:20: runtime error: left shift of 255 by 24 places cannot be represented in type 'int'
    #0 0x4b5b39 in avahi_dns_packet_consume_uint32 /src/avahi/avahi-core/dns.c:458:20
    #1 0x4b60e9 in avahi_dns_packet_consume_record /src/avahi/avahi-core/dns.c:634:9
    #2 0x4b3ea7 in LLVMFuzzerTestOneInput /src/avahi_packet_consume_record_fuzzer.cc:24:24
    #3 0x443081 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556:15
    #4 0x4427a5 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:470:3
    #5 0x444b47 in fuzzer::Fuzzer::MutateAndTestOne() /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:698:19
    #6 0x4458d5 in fuzzer::Fuzzer::Loop(std::Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:830:5
    #7 0x433a58 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:824:6
    #8 0x45dad2 in main /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
    #9 0x7f11c0bf382f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
[...]

PR #268, opened 2020-02-25
modify fuzz test err

traffic-millions:
#267  the solution;

I have test it's effective.

### DIFF
--- a/avahi-core/dns.c
+++ b/avahi-core/dns.c
@@ -455,7 +455,7 @@ int avahi_dns_packet_consume_uint32(AvahiDnsPacket *p, uint32_t *ret_v) {
         return -1;
 
     d = (uint8_t*) (AVAHI_DNS_PACKET_DATA(p) + p->rindex);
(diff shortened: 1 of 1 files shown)


traffic-millions:
@lathiat

lathiat:
Thanks for the fix!

Committed but I updated your commit message to be more clear. Copy included below for reference. In future also feel free to just file a PR (instead of Bug+PR) since in GitHub they are effectively the same.

avahi_dns_packet_consume_uint32: fix potential undefined behavior

avahi_dns_packet_consume_uint32 left shifts uint8_t values by 8, 16 and 24 bits to combine them into a 32-bit value. This produces an undefined behavior warning with gcc -fsanitize when fed input values of 128 or 255 however in testing no actual unexpected behavior occurs in practice and the 32-bit uint32_t is always correctly produced as the final value is immediately stored into a uint32_t and the compiler appears to handle this "correctly".

Cast the intermediate values to uint32_t to prevent this warning and ensure the intended result is explicit.

Closes: #267 
Closes: #268 
Reference: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19304

PR #270, opened 2020-03-01
Fix bytestring decoding for proper display

### LINKED ISSUE #158: avahi-discover: UnicodeDecodeError with fr_BE.UTF-8 locales
Hi,

With fr_BE.UTF-8 locales I get the following traceback when clicking on the devices in avahi-discover:

```Service data for service 'HP LaserJet P2055dn [447A47]' of type '_http._tcp' in domain 'local' on 3.0:
	Host NPI447A47.local (192.168.1.52), port 80, TXT data: []
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 604, in msg_reply_handler
    reply_handler(*message.get_args_list(**get_args_opts))
  File "/usr/bin/avahi-discover", line 113, in service_resolved
    self.update_label(interface, protocol, name, stype, domain, host, aprotocol, address, port, avahi.txt_array_to_string_array(txt))
  File "/usr/bin/avahi-discover", line 246, in update_label
    infos = infos % (stype, name, domain, self.siocgifname(interface), self.protoname(protocol), host, address, port, txts.strip())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 20: ordinal not in range(128)```
 
Switching to LC_ALL=C fixes the issue

Avahi: 0.7

merwok:
On a Debian system with `fr_CA.UTF-8` locale, I got UnicodeDecodeErrors when trying to select any item in the tree view shown by avahi-discover.

I found out that the translated strings use non-breaking spaces before punctuation, but these strings are bytestrings rather than text (unicode) strings, which causes line 247/250 to fail (bytes-text mixing in Python leading to implicit conversion to ASCII which fails to handle `\xC2\xA0`).

A proper fix would be to use `ugettext` (would also prepare to switch to Python 3, where gettext works in unicode), but it seems that my version of `gettext` module doesn’t have it, so this is a workaround instead.

fixes #158

### DIFF
--- a/avahi-python/avahi-discover/avahi-discover.py
+++ b/avahi-python/avahi-discover/avahi-discover.py
@@ -238,12 +238,15 @@ class Main_window:
                 txts+="<b>" + _("TXT") + " <i>%s</i></b> = %s\n" % (k,v)
         else:
             txts = "<b>" + _("TXT Data:") + "</b> <i>" + _("empty") + "</i>"
+        
+        txts = txts.decode("utf-8")
 
         infos = "<b>" + _("Service Type:") + "</b> %s\n"
         infos += "<b>" + _("Service Name:") + "</b> %s\n"
         infos += "<b>" + _("Domain Name:") + "</b> %s\n"
         infos += "<b>" + _("Interface:") + "</b> %s %s\n"
         infos += "<b>" + _("Address:") + "</b> %s/%s:%i\n%s"
+        infos = infos.decode("utf-8")
         infos = infos % (stype, name, domain, self.siocgifname(interface), self.protoname(protocol), host, address, port, txts.strip())
         self.info_label.set_markup(infos)
 


gatuno1:
Another workaround for this problem is calling avahi-discover changing LANGUAGE to english:
$ LANGUAGE="en_US.UTF-8" avahi-discover

merwok:
Yes, avoiding translations with strings containing non-ASCII chars is a workaround, but this PR makes the window work for all languages :slightly_smiling_face:

slokhorst:
This breaks avahi-discover on Python 3.. Please roll back

merwok:
I had the impression that avahi only supported Python 2!

slokhorst:
The [Avahi 0.7 release notes](https://github.com/lathiat/avahi/releases/tag/v0.7) say: "avahi-discover (python): Updated for Python3 & GTK3"

Arch Linux has Avahi packaged to use Python 3 since then (>2 years ago), and it has been working great.

merwok:
A quick patch is here: https://github.com/lathiat/avahi/pull/282 (forwarded from Debian)

The other way would be to use ugettext, but this conversation depends on whether the avahi project wants to support Python 3 only or also Python 2.

PR #271, opened 2020-03-08
Feature: Allow specifying interface visibility when using reflector

khwaaj:
Adds the configuration option **reflect-interface-visibility:*source-interface*=** with the ability to specify the visibility of an interface to a set of other interfaces. Each interface listed will be able to participate in reflection of services with the source interface. The default is to allow reflection with all interfaces.

If an empty value or `!` is given, no services from the interface will be reflected to any other interface (but it will still receive reflections from other interfaces, unless prohibited by their configuration). Conversely, giving a `*` value allows reflection to all interfaces (same as the default).

**Example of use:**

Say you have three VLAN interfaces configured: `eth0.iot`, `eth0.guest` and `eth0.main`. You want to make `iot` services visible to `guest` and `main`, `guest` services visible to `main`, and any `main` services private. Adding the following three lines to the reflector configuration would accomplish that:

```
reflect-interface-visibility:eth0.main=!
reflect-interface-visibility:eth0.guest=eth0.main
reflect-interface-visibility:eth0.iot=eth0.main,eth0.guest
```

This can also be used with the **reflect-filter=** configuration to further tune what services are globally reflected (a per-interface filter option would make it even more flexible though).

---

I'm currently using the feature as a part of my local VLAN setup, and it appears to work well from what I have seen so far. It is still new though and would likely benefit from a few more eyeballs on it. If you find this a useful feature I would wery much like for it, or some version of it, to become a part of upstream at some point.

### DIFF
--- a/avahi-core/core.h
+++ b/avahi-core/core.h
@@ -31,6 +31,7 @@ typedef struct AvahiServer AvahiServer;
 #include <avahi-common/watch.h>
 #include <avahi-common/timeval.h>
 #include <avahi-core/rr.h>
+#include <avahi-core/hashmap.h>
 
 AVAHI_C_DECL_BEGIN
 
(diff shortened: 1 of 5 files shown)


khwaaj:
Closing this for now, I have a better implementation in the works.

PR #281, opened 2020-04-24
Include build-db in tarball releases

smcv:
* build: Include build-db in tarball releases
    
    SCRIPTS are not included in tarballs by default, because they might
    be generated from a template. This one isn't, so use the dist_
    prefix to distribute it.
    
    Resolves: https://github.com/lathiat/avahi/issues/277

* build: Look for build-db in the srcdir
    
    Otherwise, srcdir != builddir builds (and hence distcheck) will fail.

---

Tested with:

```
make -j5 DISTCHECK_CONFIGURE_FLAGS="--disable-libevent --disable-qt5 --disable-mono --with-systemdsystemunitdir='\$\$dc_install_base/\${systemdsystemunitdir}'" distcheck
```

(I haven't tried distcheck with libevent, Mono or Qt since I don't currently have those installed.)

### DIFF
--- a/service-type-database/Makefile.am
+++ b/service-type-database/Makefile.am
@@ -28,7 +28,7 @@ dist_noinst_SCRIPTS=build-db
 pkglibdata_DATA+=service-types.db
 
 service-types.db: service-types
-	$(AM_V_GEN)$(PYTHON) build-db $< $@.coming && \
+	$(AM_V_GEN)$(PYTHON) $(srcdir)/build-db $< $@.coming && \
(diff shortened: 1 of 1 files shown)


evverx:
I ran into this bug when I tried to run `make distcheck` on the CI (where both `python` and `GDBM` are enabled among other things) and I can confirm that with this patch applied `make distcheck` passes with all optional features enabled.

@pemensik given that without this patch avahi tarballs are somewhat broken I think it would probably make sense to merge it before the next release.

PR #282, opened 2020-04-24
avahi-discover: Don't decode unicode strings, only bytestrings

smcv:
Unicode strings (unicode in Python 2, str or unicode in Python 3) don't
have a decode method; only bytestrings (str or bytes in Python 2,
bytes in Python 3) have that. Decode exactly the strings that need
decoding.

Resolves: https://github.com/lathiat/avahi/issues/275


smcv:
Confirmed to work with both:

```
PYTHONPATH=./avahi-python python2 ./avahi-python/avahi-discover/avahi-discover
PYTHONPATH=./avahi-python python3 ./avahi-python/avahi-discover/avahi-discover
```

knzivid:
Is this PR scheduled for the next release 0.9? @lathiat

pemensik (approved):
Looks good to me

Issue #78, opened 2016-09-15
avahi_service_resolver_new causes recursion until stack is full


rlbartle:
I have found that attempting to resolve a service's address when the service is provided by the local machine can be problematic.  For some reason most hostnames seem fine, but in this case it is 100% reproducible on this machine when the hostname is axmm-00003.  

Here is the final call that queues up this doomed case.  Adding the AVAHI_LOOKUP_NO_ADDRESS flag prevents the issue.

`avahi_service_resolver_new(client, interface, protocol, name, type, domain, AVAHI_PROTO_INET, AVAHI_LOOKUP_USE_MULTICAST, resolve_callback, userdata)`

[backtrace.txt.gz](https://github.com/lathiat/avahi/files/475550/backtrace.txt.gz)

lathiat:
That is uh, quite excellent.  I tried just setting that hostname which wasn't enough to reproduce it for some super strange reason (that would have been nice :-) 

But looking at the backtrace it looks like we're continuously looking up a cname.  Do you have something on the network publishing aliases, like avahi-aliases or some code otherwise like https://github.com/airtonix/avahi-aliases/blob/master/avahi_aliases/bin/avahi-alias ?

I can reproduce it if I use that script to publish a CNAME for ourselves, triggers the recursion.  So we should detect this recursive lookup and fail out.

rlbartle:
Yes, there is also code that adds an alias.

pemensik:
I think some kind to limited TTL should be used on CNAME follows. And lookup_handle_cname should create new lookup only if previous queries did not contain the current destination pointer. Not sure if we keep that history, I think we should. At least for cnames.

pemensik:
It seems to me we just need to check l->cname_lookups for equal key. If we find it, then this lookup contains loop and we have to abort the search, call *lookup_drop_cname* to remove this search. Not sure how do we make search failed. Another check might ensure n->refs <= 1. But *l->cname_lookups* seems better place to check.

evverx:
I revamped this issue in https://github.com/lathiat/avahi/issues/78. @pemensik I wonder if RH assigned a CVE?

pemensik:
I haven't found a vulnerability bug for avahi, which would sound similar. No CVE id for this issue it seems.

PR #286, opened 2020-05-14
function lookup_find() doesn't find the correct AvahiSRBLoockup instance

szenker:
... in case the protocol member equals AVAHI_PROTO_UNSPEC. I'm note sure about the side effects of my fix, but it somehow looks correct to me.

### DIFF
--- a/avahi-core/browse.c
+++ b/avahi-core/browse.c
@@ -169,7 +169,7 @@ static AvahiSRBLookup *lookup_find(
     for (l = b->lookups; l; l = l->lookups_next) {
 
         if ((l->interface == AVAHI_IF_UNSPEC || l->interface == interface) &&
-            (l->interface == AVAHI_PROTO_UNSPEC || l->protocol == protocol) &&
+            (l->protocol == AVAHI_PROTO_UNSPEC || l->protocol == protocol) &&
             l->flags == flags &&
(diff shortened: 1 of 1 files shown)


pemensik (approved):
Yes, it seems to me this fixes copy & paste error. But the value used is the same, so it does not affect the behaviour.

evverx:
This PR makes it a bit easier to trigger https://github.com/lathiat/avahi/issues/78 so I'd put it on hold until that segfault is fixed.

evverx:
Before I forget there is another place where `interface` should be replaced with `protocol`: https://github.com/lathiat/avahi/blob/287494b526dc84a90349cf44fe1686f664dc4cc5/avahi-core/browse.c#L441-L443

pemensik:
@evverx #78 is merged. Does it mean we can merge also this one?

evverx:
This PR should be merged along with https://github.com/avahi/avahi/pull/551 (or any other PR fixing the CNAME issue).

The bug this PR fixes prevents avahi from always crashing (it's a load bearing bug so to speak :-))

friendlyhugo:
I opened #729 which cherry-picks these commits and fixes another similar bug in `lookup_drop_cname`.

evverx:
Closing this one

PR #292, opened 2020-06-24
Load distribution name from /etc/os-release

hadess:
/etc/lsb-release really isn't used much since systemd started exporting
/etc/os-release. Reuse existing lsb-release parsing code to parse
os-release. This should allow avahi to call hosts without a hostname by
their distribution name, as requested in:
https://bugzilla.redhat.com/show_bug.cgi?id=1392924

### DIFF
--- a/avahi-core/domain-util.c
+++ b/avahi-core/domain-util.c
@@ -51,13 +51,16 @@ static void strip_bad_chars(char *s) {
 }
 
 #ifdef __linux__
-static int load_lsb_distrib_id(char *ret_s, size_t size) {
+static int load_id(char *ret_s, size_t size, const char *filename, const char *field) {
     FILE *f;
 
     assert(ret_s);
     assert(size > 0);
+    assert(filename);
+    assert(field);
+    assert(strchr(field, '='));
 
-    if (!(f = fopen("/etc/lsb-release", "r")))
+    if (!(f = fopen(filename, "r")))
         return -1;
 
     while (!feof(f)) {
@@ -66,10 +69,10 @@ static int load_lsb_distrib_id(char *ret_s, size_t size) {
         if (!fgets(ln, sizeof(ln), f))
             break;
 
-        if (strncmp(ln, "DISTRIB_ID=", 11))
+        if (strncmp(ln, field, strlen(field)))
             continue;
 
-        p = ln + 11;
+        p = ln + strlen(field);
         p += strspn(p, "\"");
         p[strcspn(p, "\"")] = 0;
 
@@ -81,6 +84,14 @@ static int load_lsb_distrib_id(char *ret_s, size_t size) {
 
     fclose(f);
     return -1;
+}
+
+static int load_lsb_distrib_id(char *ret_s, size_t size) {
+    return load_id(ret_s, size, "/etc/lsb-release", "DISTRIB_ID=");
+}
+
+static int load_os_id(char *ret_s, size_t size) {
+    return load_id(ret_s, size, "/etc/os-release", "ID=");
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


hadess:
(Note that I haven't tested this at all)

hadess:
Builds correctly after the commit above is added.

pemensik (approved):
Seems good to me

evverx:
I think the second commit can be removed. That issue should be fixed in https://github.com/lathiat/avahi/pull/281.

Looking at https://coveralls.io/builds/54825736/source?filename=avahi-core%2Fdomain-util.c it seems this code isn't covered by any unit tests and in terms of adding new tests eventually I think it would be better if the function could accept `FILE *` instead of `char *` to make it possible to create files with `fmemopen` or something like that, write stuff like `ID=` and pass it to `load_id` to make it sure it can handle it. Also according to https://www.freedesktop.org/software/systemd/man/os-release.html#ID= when `ID` isn't set `linux` should be used by default and `load_id` doesn't seem to follow that (I'm not sure whether it should though).

I also went to https://bugzilla.redhat.com/show_bug.cgi?id=1392924#c0 and found the following paragraph there
> though it might hurt privacy[1]
overriding it is still possible, and it helps with branding recognition.

and I'm not sure whether avahi should use those IDs given that they are more specific than what usually goes to `HINFO` (disabled by default with `publish-hinfo=no`). It's kind of weird I think.

pemensik:
I have merged PR #281, therefore the second commit is not necessary and would conflict. Please rebase on current master and use just first commit. I think it is ready to be merged.

avahi-core/domain-util.c
```diff
@@ -66,10 +69,10 @@ static int load_lsb_distrib_id(char *ret_s, size_t size) {
         if (!fgets(ln, sizeof(ln), f))
             break;
 
-        if (strncmp(ln, "DISTRIB_ID=", 11))
+        if (strncmp(ln, field, strlen(field)))
```
pemensik, line 72: Should be compute strlen(field) only once per function call, not twice per line? Not sure optimization can guess it won't change, but wouldn't hurt if we did that explicitly once just after file were opened. It is no longer literal, but it will never change.

avahi-core/domain-util.c
```diff
@@ -82,6 +85,14 @@ static int load_lsb_distrib_id(char *ret_s, size_t size) {
     fclose(f);
     return -1;
 }
+
+static int load_lsb_distrib_id(char *ret_s, size_t size) {
+    return load_id(ret_s, size, "/etc/lsb-release", "DISTRIB_ID=");
+}
+
+static int load_os_id(char *ret_s, size_t size) {
+    return load_id(ret_s, size, "/etc/os-release", "ID=");
```
pemensik, line 94: Oh, there is defined also ``DEFAULT_HOSTNAME=`` variable in man os-release, which Fedora distribution actually uses. It is now the same value as ID= and only optional. But I think we use that identification for hostname, so we should attempt to use that hostname if present. This is defined exactly for our use case. I think we should use DEFAULT_HOSTNAME from that file if present, ID as a fallback. Only ``ID=`` is present on Debian. But then the function should be called *load_os_hostname* or *load_os_default_hostname*.

pemensik (changes requested):

hadess:
I've rebased it, and it applies cleanly. I'll let somebody else takeover with the fixes. Would be great if you could use the co-developed-by, and such when you do: https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by

pemensik:
Did a quick fix in your branch, hope you do not mind.

pemensik:
Unfortunately the linux specific static functions are not exported in any public function before gethostname call. This makes the function not testable by unittest, unless the host would have empty hostname. Should be made testable in the future, but it seems okay for now.

PR #301, opened 2020-08-27
Replace gendered pronouns with gender neutral ones.

leeceeksdee:


lathiat:
This looks great, thank you!

Issue #305, opened 2020-09-17
Logic for point-to-point on tunnels incorrectly requires IFF_MULTICAST


pprindeville:
In the case of a point-to-point tunnel, there's only a single recipient at the other end, and therefore the distinction of unicast/multicast/broadcast is moot.

Requiring the `IFF_MULTICAST` flag is therefore meaningless in the present of `IFF_POINTOPOINT`.

evverx:
I think it can be closed.

PR #306, opened 2020-09-17
Logic for p2p on tunnels incorrectly requires IFF_MULTICAST (#305)

pprindeville:
Point-to-point interfaces don't have MULTICAST set on them, since anything other than a unicast is a meaningless concept.

I've back ported this fix to CentOS 8 Stream and avahi-0.7 and it works fine.

Addresses issue #305.

### DIFF
--- a/avahi-core/iface-linux.c
+++ b/avahi-core/iface-linux.c
@@ -105,8 +105,8 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
             (ifinfomsg->ifi_flags & IFF_UP) &&
             (!m->server->config.use_iff_running || (ifinfomsg->ifi_flags & IFF_RUNNING)) &&
             ((ifinfomsg->ifi_flags & IFF_LOOPBACK) ||
-             (ifinfomsg->ifi_flags & IFF_MULTICAST)) &&
-            (m->server->config.allow_point_to_point || !(ifinfomsg->ifi_flags & IFF_POINTOPOINT));
+             (ifinfomsg->ifi_flags & IFF_MULTICAST) ||
+             ((ifinfomsg->ifi_flags & IFF_POINTOPOINT) && m->server->config.allow_point_to_point));
 
         /* Handle interface attributes */
         l = NLMSG_PAYLOAD(n, sizeof(struct ifinfomsg));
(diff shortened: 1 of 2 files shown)


pprindeville:
Can I please get a review of this?

pprindeville:
Still hoping for a review...

pprindeville:
Bueller... Bueller...

skleeschulte:
I came across this PR while trying to fix a problem with avahi listiening on a wireguard interface. Someone please correct me if I'm wrong, but as far as I know: Wireguard, by default, uses interfaces flagged as POINTOPOINT, but _not_ MULTICAST. A wireguard interface can have multiple "peers". Wireguard does not support multicast traffic - it will forward multicast packets (only) to the first peer that has the multicast address configured in its AllowedIPs, or discard it. However, for the special case of an interface with only one peer and (a) multicast address(es) configured in this peer's AllowedIPs, it makes sense to add the MULTICAST flag to the wireguard interface.

In short: Wireguard interfaces by default are point-to-point interfaces which cannot handle multicast traffic. So it does make sense to check for the multicast flag on point-to-point interfaces.

pprindeville:
Still hoping for a merge...

pemensik:
Is there any bug this change is attempting to fix? Sure, p2p interface lacks usually multicast flag for a good reason. If you enable p2p interface explicitly in configuration, what is expected outcome? Is it able to use unicast with remote peer instead?

pprindeville:
> Is there any bug this change is attempting to fix? Sure, p2p interface lacks usually multicast flag for a good reason. If you enable p2p interface explicitly in configuration, what is expected outcome? Is it able to use unicast with remote peer instead?

I use it with IPsec tunnels that are marked multicast and it works as expected: I can peer with other Anahi's running on remote firewalls. Testing for multicast prevents having to configure a large number of unicast peers, especially if that list is highly dynamic.

pemensik (approved):
Yes, this thing uses multicast flag even on p2p devices and when they are present, it does not require explicit enabling of p2p. I think that makes sense. If that p2p is capable doing multicast, then why not using it. If not, then you can still enable it explicitly.

pemensik:
@pprindeville It would be nice if you could provide links to distributions already using this patch. It looks good to me, but I don't have any connections to actually test that change. I expect chance for regressions is low, but still possible. I haven't found this patch in Fedora rawhide, or [CentOS Stream](https://gitlab.com/redhat/centos-stream/rpms/avahi/-/tree/c9s).

You have said CentOS 8 has i merged. Were you talking about your own builds for CentOS? Just looking on patch names, it does not seem any of them solves this problem. Anyway, I am not against it, but I would like to know how many people might have tested this change.

pemensik:
So IPSec tunnels actually support multicast even on p2p interfaces?

pprindeville:
> It would be nice if you could provide links to distributions already using this patch.

@pemensik Did that above. OpenWRT has been shipping like this for a while, though I don’t know if the standard build uses Avahi or mDNS-Responder.

pprindeville:
As for the reference to CentOS above, I’m a packager but not for this. At a previous company we were prototyping radio gateways that used Strongswan and Avahi to create mesh networks and we used this patch for our internal builds.

PR #309, opened 2020-10-02
Don't conflict with self

vladmovchan:
If during registration of our own records there is an address change or address removed and re-added then `avahi-daemon` interprets its own probe as conflicting (due to presence of different or temporary non-existing address in the probe). And then it changes its name to `hostname-2` and retries.
Change of hostname without good reason could be unexpected and annoying.
To resolve this we ignore presence of non-matching records if they were sent by us.

This change resolves https://github.com/lathiat/avahi/issues/117 for me.

### DIFF
--- a/avahi-core/server.c
+++ b/avahi-core/server.c
@@ -193,7 +193,7 @@ static void withdraw_rrset(AvahiServer *s, AvahiKey *key) {
         withdraw_entry(s, e);
 }
 
-static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *i) {
+static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *i, int from_local_iface) {
     AvahiEntry *e, *n;
     int ours = 0, won = 0, lost = 0;
 
@@ -210,7 +210,7 @@ static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *
         if (e->dead)
             continue;
 
-        if ((cmp = avahi_record_lexicographical_compare(e->record, record)) == 0) {
+        if ((cmp = avahi_record_lexicographical_compare(e->record, record)) == 0 || from_local_iface) {
             ours = 1;
             break;
         } else {
@@ -639,7 +639,7 @@ static void handle_query_packet(AvahiServer *s, AvahiDnsPacket *p, AvahiInterfac
             if (!avahi_key_is_pattern(record->key)) {
                 if (!from_local_iface)
                     reflect_probe(s, i, record);
-                incoming_probe(s, record, i);
+                incoming_probe(s, record, i, from_local_iface);
             }
 
             avahi_record_unref(record);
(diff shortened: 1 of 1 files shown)


zonque:
I can confirm the patch in this PR fixes the problem for me on an embedded platform that operates avahi on a network bridge device. Thanks a lot for sharing!

vladmovchan:
Thank you for the feedback!

ffleandro:
What's the ETA on this?
I will try this patch to see if this fixes my observed issues.

ffleandro:
Can you provide instructions on how to test this locally in a debian installation (installed throug apt-get)?

vladmovchan:
@ffleandro, you could try to ask project author about the ETA in the original bug: https://github.com/lathiat/avahi/issues/117. Last reply from lathiat there is dated Dec 7, 2020.
And sorry I can't provide you with instructions. If you don't know how to apply patch and how to build avahi from sources I don't think you should even try. As you can put your system to a broken state from which you don't know how to rollback by yourself.

SoongNoonien:
Hi!
I'm running this patch for about two weeks now and never encountered #117 again. But I found a regression, MYHOSTNAME.local is now only reachable from the other machines in the network. Is this intended behavior?

vladmovchan:
Thank you for the feedback.
The regression which you described is not intended. I follow up if I see how to address it.

pemensik:
I think this is important fix. I have seen myself changed name of my laptop. I am pretty sure there were not another device with the same name, but my laptop were connected to two interfaces simultaneously.

pemensik (approved):
Code change seems reasonable. But I haven't tested myself whether it helps indeed.

guillerodriguez:
> Code change seems reasonable. But I haven't tested myself whether it helps indeed.

[This comment](https://github.com/lathiat/avahi/pull/309#issuecomment-1242675947) mentions a regression which is apparently acknowledged by the original author of the patch (next comment). Are you sure this change is safe?

pemensik:
No, I am not sure. I looks well in general. Thanks for pointing to a regression. Do I understand it correctly that this change should make not working this command: ``avahi-resolve -n $HOSTNAME.local`` ?

But yes, it changes the behaviour of local resolving of hostname and its reverse IPv4 address.

pemensik (changes requested):
After aplying this change and testing it shortly, it causes different behaviour.

- ``avahi-resolve -n $HOSTNAME.local -4`` timeouts, where it succeeded before.
- ``avahi-resolve -n $HOSTNAME.local -6`` still works.
- ``avahi-resolve -a $IPV4`` timeouts, but ``avahi-resolve -a $IPV6`` with addresses from correct interface

Not sure why only IPv4 is affected and IPv6 is not. But that needs to be solved before merging.

SoongNoonien:
Hi, I may have found another regression. Mounting Samba shares using the `.local` hostname is significantly slower than without the patch. But maybe this is actually a Samba issue.

evverx:
I think this PR can be closed.

I've also opened https://github.com/avahi/avahi/issues/611 with a regression affecting local things similar to the one observed here. There was an attempt to fix similar conflicts with the reflector on and that regression was introduced in the process.

Thank you everyone for testing the patch.

PR #312, opened 2020-11-17
fix QT3 build

rdieter:
recent commit d1e71b320d96d0f213ecb0885c8313039a09f693 adding QT5
support added a new conditional
but failed to actually set the define.  This commit adds that to
allow successful build when enabling QT3 support

### DIFF
--- a/avahi-qt/Makefile.am
+++ b/avahi-qt/Makefile.am
@@ -38,7 +38,7 @@ libavahi_qt3_la_SOURCES = \
 qt-watch.moc3: qt-watch.cpp
 	$(AM_V_GEN)$(MOC_QT3) $^ > $@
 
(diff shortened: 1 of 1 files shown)


rdieter:
See also downstream fedora bug,
https://bugzilla.redhat.com/show_bug.cgi?id=1897925

rdieter:
ping, pretty simple straight-forward fix here

rdieter:
ping, one more try before giving up

pemensik:
Hello @rdieter , is this change still relevant? Should it be merged in current form? I expect Qt3 is not used anymore in Fedora, is that correct?

rdieter:
Still relevant, I think it ought to be merged

pemensik (approved):

PR #318, opened 2021-01-10
Fix Readme's URL 

nozo-moto:
## What
remove ohlow's url and add openhub's url to README

## WHY
ohlow was deleted.
https://www.openhub.net/p/ohloh


Darrion35:
#[](url)
Duplicate of #

Darrion35:
ry opening github for an answear.

PR #320, opened 2021-01-24
close txt-record markups in examples in man 

bmillemathias:


EskoDijk (approved):
This fixes the XML syntax error that is present in the 'man avahi.service' page XML examples.

Issue #321, opened 2021-02-01
avahi-autoipd


cyrilbur-adder:
Hello,

Question regarding autoipd.

When dealing with event == EVENT_PACKET, we could be in STATE_SLEEPING. I can't help but wonder if a) we shouldn't deal with EVENT_PACKET if we're STATE_SLEEPING or, at least not transition to STATE_WAITING_PROBE if we're state sleeping.

I'm happy to prepare a patch depending on what seems more sane, I'll admit I likely haven't contemplated all the possible caveats, I've only looked at this code for a day or two. I think not entering event == EVENT_PACKET if we're sleeping makes most sense.

Kind regards,

Cyril

PR #322, opened 2021-02-03
autoipd: Don't forget SLEEPING state on conflicting packet

cyrilbur-adder:
If a event type EVENT_PACKET is received and it is determined to be a
conflicting packet the state machine will be kicked into
STATE_WAITING_PROBE.

This is a problem as the autoipd may have been sleeping
(STATE_SLEEPING). This means that despite the presence of a routable
address on the interface autoipd will call its action script.

It seems appropriate that autoipd should only process packets if it is
not sleeping.

### DIFF
--- a/avahi-autoipd/main.c
+++ b/avahi-autoipd/main.c
@@ -1206,7 +1206,7 @@ static int loop(int iface, uint32_t addr) {
                 retval_sent = 1;
             }
 
-        } else if (event == EVENT_PACKET) {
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR.

Looks like it's related to https://github.com/avahi/avahi/issues/321 so I linked this PR to that issue to prevent it from getting lost. If it's still relevant could you attach logs/pcaps there to make it easier to figure out what this PR should fix? In the meantime I'll go ahead and close this PR. It can always be reopened once it's confirmed that it works as expected.

PR #330, opened 2021-03-26
Avoid infinite-loop in avahi-daemon by handling HUP event in client_work

ret2libc:
If a client fills the input buffer, client_work() disables the
AVAHI_WATCH_IN event, thus preventing the function from executing the
`read` syscall the next times it is called. However, if the client then
terminates the connection, the socket file descriptor receives a HUP
event, which is not handled, thus the kernel keeps marking the HUP event
as occurring. While iterating over the file descriptors that triggered
an event, the client file descriptor will keep having the HUP event and
the client_work() function is always called with AVAHI_WATCH_HUP but
without nothing being done, thus entering an infinite loop.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=984938

### DIFF
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -424,6 +424,11 @@ static void client_work(AvahiWatch *watch, AVAHI_GCC_UNUSED int fd, AvahiWatchEv
         }
     }
 
+    if (events & AVAHI_WATCH_HUP) {
+        client_free(c);
+        return;
+    }
+
     c->server->poll_api->watch_update(
         watch,
         (c->outbuf_length > 0 ? AVAHI_WATCH_OUT : 0) |


ret2libc:
By the way, I've requested a CVE through Red Hat for this issue.

carnil:
CVE-2021-3468 has been assigned for this issue according to https://bugzilla.redhat.com/show_bug.cgi?id=1939614

utkarsh2102:
@lathiat, could you take a look at this whenever free, please? 😅

ret2libc:
Ping on this.

mbiebl:
@lathiat could you have a look at this PR?
This was raised downstream at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=984938 but before applying this PR into the Debian package, I'd appreciate if you could review the changes.

pemensik (approved):
It seems an important, yet simple fix.

carnil:
> @lathiat could you have a look at this PR? This was raised downstream at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=984938 but before applying this PR into the Debian package, I'd appreciate if you could review the changes.

@lathiat any chance you could review the pull request so we can have the fixes applied in time for the next Debian release 'bookworm'?

pemensik:
Merged, haven't found anything this might break. Has this change been tested on any distribution already, at least rolling one? It was not tested in Fedora.

mbiebl:
It wasn't tested in Debian either, as I wanted to have an ack from upstream first.

PR #366, opened 2022-01-03
Do not disable timeout cleanup on watch cleanup

kov:
This was causing timeouts to never be removed from the linked list that
tracks them, resulting in both memory and CPU usage to grow larger over
time.

### DIFF
--- a/avahi-common/simple-watch.c
+++ b/avahi-common/simple-watch.c
@@ -238,7 +238,7 @@ static void cleanup_watches(AvahiSimplePoll *s, int all) {
             destroy_watch(w);
     }
 
-    s->timeout_req_cleanup = 0;
+    s->watch_req_cleanup = 0;
 }
 
 static AvahiTimeout* timeout_new(const AvahiPoll *api, const struct timeval *tv, AvahiTimeoutCallback callback, void *userdata) {


lathiat:
Oh nice find thank you!

abrender:
Friendly ping @lathiat - is there any chance of cutting a new release which includes this (and other?) fixes? That way we can get the fixes upstreamed in Ubuntu, Fedora, etc.

Thanks!

pemensik:
I would like to have solved bugs marked as *important* before releasing final release. But it might have to prepare at least rc candidate tag, so we get more testing of current master with already included fixes. Alternative would be releasing just 0.8.1 release with current fixes and try to fix important issue for 0.9 release.

https://github.com/lathiat/avahi/issues?q=is%3Aissue+is%3Aopen+label%3Aimportant

But I have been a bit behind the schedule myself, with not much time to spend on preparing avahi changes.

Doomsdayrs:
> @pemensik - it makes total sense to wait for important bugs to get solved before releasing a new final release.
> 
> I think the alternative idea of release a minor release - 0.8.1 with the current fixes sounds like a good idea! That way everyone in the downstream distributions gets to benefit from all the squashed bugs so far.
> 
> Thanks for considering this!

@evverx It seems that this has been on the back burner to release for a long time, Which means this issue is plaguing Fedora installs for 7 years now.

evverx:
@Doomsdayrs it was discussed recently in https://github.com/avahi/avahi/issues/550 and currently it's being discussed in https://github.com/avahi/avahi/issues/503. Some distros roll forward along with avahi and some distros backported this patch as far as I know so it doesn't affect all of them at least.

BloodyIron:
How is it Avahi hasn't had a release in 4 years, let alone a fix for this bug. Like this isn't how modern software development works. I'm at a LAN party and pretty sure this bug is pegging one of my CPU cores. Just release v0.8.1 or similar already please.

evverx:
> Just release v0.8.1 or similar

https://github.com/avahi/avahi/releases/tag/v0.9-rc1 was already tagged and distros like Arch and Wolfi (https://github.com/wolfi-dev/os/pull/12735) already switched to it. Other distros backported the patches they are interested in. Generally the status is tracked in https://github.com/avahi/avahi/issues/503.

PR #369, opened 2022-01-30
avahi-discover: Escape strings substituted into Pango markup

smcv:
Otherwise, a TXT entry containing a URL with '&' will cause an error.

---

Branch also includes #282, otherwise it would have conflicts with that PR.

### DIFF
--- a/avahi-python/avahi-discover/avahi-discover.py
+++ b/avahi-python/avahi-discover/avahi-discover.py
@@ -238,15 +238,17 @@ class Main_window:
                 txts+="<b>" + _("TXT") + " <i>%s</i></b> = %s\n" % (k,v)
         else:
             txts = "<b>" + _("TXT Data:") + "</b> <i>" + _("empty") + "</i>"
-        
-        txts = txts.decode("utf-8")
+
+        if isinstance(txts, bytes):     # Python 2
+            txts = txts.decode("utf-8")
 
         infos = "<b>" + _("Service Type:") + "</b> %s\n"
         infos += "<b>" + _("Service Name:") + "</b> %s\n"
         infos += "<b>" + _("Domain Name:") + "</b> %s\n"
         infos += "<b>" + _("Interface:") + "</b> %s %s\n"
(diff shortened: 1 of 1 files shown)


pemensik (approved):
Seems good. Not able to test it.

evverx:
What's weird it that CodeQL fails to extract this particular script for some reason. I'll try to get it to work to make it possible to at least analyze it statically. It isn't much in terms of testing but CodeQL certainly should be able to catch unescaped strings injected into any kind of markup.

smcv:
> CodeQL certainly should be able to catch unescaped strings injected into any kind of markup

How would it know what is markup and what is a correctly- escaped string? Static analysis of a dynamic language isn't straightforward, especially when the types are on the other side of a FFI boundary (as they are here).

evverx:
> How would it know what is markup and what is a correctly- escaped string? 

It tends to complain about things like this a lot (where it should and where it shouldn't) so I was kind of assuming (or more precisely hoping) that it would automagically work in this particular case (given that at this point it's the only place where the Python code is somewhat "tested"). It didn't work out though.

> Static analysis of a dynamic language isn't straightforward especially when the types are on the other side of a FFI boundary (as they are here)

Agreed.

Just to clarify I'm trying to set up some sort of CI here (to mostly cover basic things like https://github.com/lathiat/avahi/pull/420, https://github.com/lathiat/avahi/pull/425  and https://github.com/lathiat/avahi/pull/422) and in the process I'm trying to figure out how it can be improved. In this particular case I noticed the file couldn't be extracted and now it should be fixed. It isn't possible to get far in terms of testing the Python code at this point though.

smcv:
Now that https://github.com/lathiat/avahi/pull/282 has been merged, I believe this will apply cleanly if rebased.

pemensik:
Ah, my fault. I just need python3-twisted installed. Yes, this fixes it for me.

pemensik:
Thanks! This python version is not packaged on Fedora at the moment. I hat to tune few things.

PR #372, opened 2022-02-28
skip error handling.

ihsinme:
using an assignment instead of a comparison means that these parts of the code cannot signal possible errors.

which results in no error handling.


ihsinme:
good afternoon.
any news on this PR?

pemensik (approved):
Yes, that seems correct fix

pemensik:
This fixes the same issue as #202 PR. Closing as duplicate.

PR #377, opened 2022-05-21
build: Use runstatedir instead home-brewn variable

jtojnar:
Switching to a standard variable will allow to simplify downstream packages. The variable was introduced in Autoconf 2.70: https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

Fixes: https://github.com/lathiat/avahi/issues/137

### DIFF
--- a/avahi-autoipd/Makefile.am
+++ b/avahi-autoipd/Makefile.am
@@ -25,7 +25,7 @@ AM_CFLAGS= \
 
 # This cool debug trap works on i386/gcc only
 AM_CFLAGS+='-DDEBUG_TRAP=__asm__("int $$3")' \
-	-DAVAHI_RUNTIME_DIR=\"$(avahi_runtime_dir)/\" \
+	-DAVAHI_RUNTIME_DIR=\"$(runstatedir)/\" \
 	-DAVAHI_IPCONF_SCRIPT=\"$(pkgsysconfdir)/avahi-autoipd.action\" \
 	-DAVAHI_IPDATA_DIR=\"$(localstatedir)/lib/avahi-autoipd\"
 
--- a/avahi-daemon/Makefile.am
+++ b/avahi-daemon/Makefile.am
@@ -29,7 +29,7 @@ introspectiondir=$(datadir)/dbus-1/interfaces
 dbussystemservicesdir=$(datadir)/dbus-1/system-services
 
 AM_CFLAGS+= \
-	-DAVAHI_DAEMON_RUNTIME_DIR=\"$(avahi_runtime_dir)/avahi-daemon/\" \
+	-DAVAHI_DAEMON_RUNTIME_DIR=\"$(runstatedir)/avahi-daemon/\" \
 	-DAVAHI_SOCKET=\"$(avahi_socket)\" \
 	-DAVAHI_SERVICE_DIR=\"$(servicedir)\" \
 	-DAVAHI_CONFIG_FILE=\"$(pkgsysconfdir)/avahi-daemon.conf\" \
@@ -80,7 +80,7 @@ dist_pkgdata_DATA = \
 
 %.socket: %.socket.in
 	$(AM_V_GEN)sed -e 's,@sbindir\@,$(sbindir),g' \
-		-e 's,@avahi_runtime_dir\@,$(avahi_runtime_dir),g' $< > $@
+		-e 's,@runstatedir\@,$(runstatedir),g' $< > $@
 
 if HAVE_SYSTEMD
 systemdsystemunit_DATA = \
@@ -168,4 +168,4 @@ xmllint:
 	done
 
 install-data-local:
-	test -z "$(avahi_runtime_dir)" || $(MKDIR_P) "$(DESTDIR)$(avahi_runtime_dir)"
+	test -z "$(runstatedir)" || $(MKDIR_P) "$(DESTDIR)$(runstatedir)"
--- a/avahi-daemon/avahi-daemon.socket.in
+++ b/avahi-daemon/avahi-daemon.socket.in
@@ -19,7 +19,7 @@
 Description=Avahi mDNS/DNS-SD Stack Activation Socket
 
 [Socket]
-ListenStream=@avahi_runtime_dir@/avahi-daemon/socket
+ListenStream=@runstatedir@/avahi-daemon/socket
 
 [Install]
 WantedBy=sockets.target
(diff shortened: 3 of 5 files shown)


evverx:
> The variable was introduced in Autoconf 2.70

I opened a PR recently where I also needed to bump `autoconf` to be able to bring the latest version of `AX_PTHREAD` to compile avahi with clang and -fsanitize=address out of the box. In the process I tried to figure out how safe it is to bump it and it seems that it should be relatively safe to bump autoconf to 2.69 but it turns out that 2.70 cuts off a lot of build environments.

All in all I think it boils down to how conservative the avahi project should be when build dependencies are bumped and I'm not sure there is any policy covering stuff like that.

evverx:
It seems this PR should address https://github.com/lathiat/avahi/issues/137 and https://github.com/lathiat/avahi/issues/150. In https://github.com/lathiat/avahi/issues/137#issuecomment-384910280 it was suggested that something like
```
+# Autoconf 2.70 will support this, and many distros patch this option in,
+# but Autoconf 2.70 hasn't actually been released yet.
+AS_IF([test -z "${runstatedir}"], [runstatedir='${localstatedir}/run'])
+AC_SUBST([runstatedir])
```
can be used to make `runstatedir` work with `autoconf < 2.70` and I think as long as `autoconf 2.69` is still relatively widely used that option would be better.

jtojnar:
2.70 is [over two years old](https://lwn.net/Articles/839395/) and [widely supported](https://repology.org/project/autoconf/versions) now.

And if someone did not update autoconf for years, why would they care about updating avahi?

evverx:
> And if someone did not update autoconf for years, why would they care about updating avahi?

That's a good question. For some reason some people try to build upsream projects using 2-year-old (or older) toolchains. For example relatively recently `systemd` decided to not bump `meson` (and bring some kludges along the way) to avoid some kind of friction it supposedly could have caused on Debian stable on something like that.

For the record personally I'm totally fine with bumping autoconf.

eli-schwartz:
Bumping autoconf is significantly lower cost than bumping Meson, since Meson requires the build toolchain update Meson, while autoconf is run on the developer machine while building release tarballs and the build toolchain does not have to have autoconf installed at all.

evverx:
@eli-schwartz I used it as an example of bringing upstream projects to older systems because I can't actually build systemd normally using the latest version of meson and it annoys me so much that I remember about it. I agree that it isn't comparable.

mbiebl:
Speaking as the Debian maintainer, I'd be totally fine, if the minimum autoconf version is bumped to 2.70, especially since it's only relevant for those generating the build files. So if you want to build on a system, that doesn't have 2.70, you can just use a pre-generated source tarball.

evverx:
>  if you want to build on a system, that doesn't have 2.70, you can just use a pre-generated source tarball

I'm not sure unsigned tarballs downloaded from the internet should be shipped anywhere but it's not that anything else is signed here either :-)

Anyway I don't think avahi is actively maintained any more so I think this repository should just be made a public archive to make it clear.

evverx:
> I don't think avahi is actively maintained any more so I think this repository should just be made a public archive to make it clear

I moved that discussion where it belongs: https://github.com/lathiat/avahi/issues/388#issuecomment-1565467687

evverx (approved):
LGTM. Thanks!

I ran it on top of https://github.com/evverx/avahi/pull/5 just in case and the CI is happy.

@pemensik I'm not sure if you reviewed it or not so I'll leave it to you to merge it.

mbiebl:
I looked at the PR, LGTM as well (also with my Debian maintainer hat on).
Please consider merging for the upcoming 0.9 release.

evverx:
While this PR still looks good to me I think it should be mentioned somewhere that if that variable isn't set explicitly it can end up being `/var/run` and that's kind of backward incompatible with 60bd97ceeb1f7924df2b347b5ac18fa5949898cc. That commit wasn't backward-compatible either though :-)

evverx:
I'm not sure. I think it depends on whether that default should be kept. Based on all the bug reports I've seen here it seems avahi is used on all sorts of IOT stuff with no systemd much more often and they seem to have been happy with `/var/run` to judge from https://github.com/lathiat/avahi/issues/150. That being said they probably changed their build scripts and if it's changed again it's going to break something again. Maybe it would make sense to set it to `/run` if the variable isn't passed explicitly.

mbiebl:
Personally, I probably wouldn't override the default and simply document it in the release notes.

evverx:
Hm, looking at https://bugzilla.redhat.com/show_bug.cgi?id=1937406 it seems `runstatedir` isn't as widely supported by packaging stuff as I thought. I guess that's why this PR stalled.

Either way I'll relabel it because it isn't just "nice to have". It fixes the regression reported quite a few times here in various places.

evverx:
Looks like `nss-mdns` switched to `runstatedir` back in 2021: https://github.com/lathiat/nss-mdns/commit/df4364ec5b12082b66b7bed9311adbec88b0fd51 and there hasn't been any pushback as far as I can see so it's convincing enough for me. Either way this PR has been open long enough so I'll let it sit for a week, rebase it and merge it if there aren't any objections.

jtojnar:
Rebased but it looks like some CI checks do not support Autoconf 2.70 :crying_cat_face: 

We should probably update it but alternatively, we can use the suggestion from above https://github.com/lathiat/avahi/pull/377#issuecomment-1348105385 to make it work without 2.70: https://github.com/lathiat/avahi/compare/master...jtojnar:avahi:runstatedir-wo-270

evverx:
The OSS-Fuzz base-builder image is based on Ubuntu Focal and it comes with autoconf-2.69 apparently. Given that they switched from Xenial when it reached "End of Standard Support" I'm pretty sure Focal will be kept there until 2025. It should be possible to get it around in the OSS-Fuzz build-script (https://github.com/lathiat/avahi/blob/master/fuzz/oss-fuzz.sh) by either installing the latest autoconf or replacing `runstatedir` with `sed` (because that variable doesn't affect the fuzz targets) but I guess Simon's suggestion would be the cleanest way to address it.

evverx:
@jtojnar thanks!

evverx:
@jtojnar if you by any chance is involved in packaging avahi on NixOS it would be great if you could take a look at https://github.com/lathiat/avahi/issues/516. Thanks again!

jtojnar:
Thanks for letting me know, I will check it out.

PR #381, opened 2022-06-05
Fix encoding of avahi-common/domain.h to be UTF-8

mbiebl:
spotted by lintian:
```
W: libavahi-common-dev: national-encoding usr/include/avahi-common/domain.h
N:
N:   A file is not valid UTF-8.
N:
N:   Debian has used UTF-8 for many years. Support for national encodings is being phased out. This file probably appears to users
N:   in mangled characters (also called mojibake).
N:
N:   Packaging control files must be encoded in valid UTF-8.
N:
N:   Please convert the file to UTF-8 using iconv or a similar tool.
```

### DIFF
--- a/avahi-common/domain.h
+++ b/avahi-common/domain.h
@@ -33,7 +33,7 @@ AVAHI_C_DECL_BEGIN
  * is calculated like this: RFC1034 mandates maximum length of FQDNs
  * is 255. The maximum label length is 63. To minimize the number of
  * (non-escaped) dots, we comprise our maximum-length domain name of
- * four labels แ 63 characters plus three inner dots. Escaping the
+ * four labels รก 63 characters plus three inner dots. Escaping the
  * four labels quadruples their length at maximum. An escaped domain
  * name has the therefore the maximum length of 63*4*4+3=1011. A
  * trailing NUL and perhaps two unnecessary dots leading and trailing


mbiebl:
The alternative would be to avoid the usage of `á`

avahi-common/domain.h
```diff
@@ -33,7 +33,7 @@ AVAHI_C_DECL_BEGIN
  * is calculated like this: RFC1034 mandates maximum length of FQDNs
  * is 255. The maximum label length is 63. To minimize the number of
  * (non-escaped) dots, we comprise our maximum-length domain name of
- * four labels � 63 characters plus three inner dots. Escaping the
+ * four labels á 63 characters plus three inner dots. Escaping the
```
EskoDijk, line 36:
```suggestion
 * four labels of 63 characters plus three inner dots. Escaping the
```

EskoDijk (changes requested):
Maybe just use 'of' ? The proposed character has the wrong accent, it should be à. Wikipedia said the French à usage evolved into English in the @ character but that's best not used in code. So for me approved with the edit I propose to 'of'.

avahi-common/domain.h
mbiebl: You are right. The semantically correct symbol is à. It's used in German as a preposition when talking about quantities, describing a property of a set of elements.
I'd say the meaning is something like this:
`...four labels, each of them consisting of 63 characters, plus three inner dots.`

PR #382, opened 2022-06-05
Create luciano-codeql.yml

lucianoCircelli:
22_06_05 I try to link my router that have an IP6 corelate with my number phone from my provider (RIR)

### DIFF
--- a/.github/workflows/luciano-codeql.yml
+++ b/.github/workflows/luciano-codeql.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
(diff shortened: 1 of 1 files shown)


pemensik:
What should this do?

evverx:
> What should this do?

From https://github.com/github/codeql-action#codeql-action
> This action runs GitHub's industry-leading semantic code analysis engine, CodeQL, against a repository's source code to find security vulnerabilities. It then automatically uploads the results to GitHub so they can be displayed in the repository's security tab. CodeQL runs an extensible set of [queries](https://github.com/github/codeql), which have been developed by the community and the [GitHub Security Lab](https://securitylab.github.com/) to find common vulnerabilities in your code.

I'm not sure it works in its current form though. Looks like this PR was opened semi-automatically through the security tab or something like that.

evverx:
I got it to work in https://github.com/evverx/avahi/pull/1. I'm not sure whether GHActions are turned on here at all but I'll try to send a PR soon. Apart from the static analysis It should help to make sure that avahi is more or less buildable on x86_64. It's not much but it's certainly better than nothing.

evverx:
FWIW I think this PR can be closed because it should be superseded by https://github.com/lathiat/avahi/pull/413 where CodeQL works.

pemensik:
This is duplicating what I have merged already. Merged in #413 instead.

Issue #383, opened 2022-06-23
Interface removal while group announcements probing leaves group hanging 


awy:
When `remove_announce()` is called, perhaps because an interface is no longer relevant, and the announcer was probing, then the associated group's `n_pending` count is not decremented and so remaining announcers in the group will never move to ` AVAHI_ANNOUNCING`.

evverx:
As far as I understand it was presumably fixed in https://github.com/lathiat/avahi/pull/39. I think this issue can be closed.

PR #384, opened 2022-06-23
Fix remove of probing announcer

awy:
If an announcer is in state AVAHI_PROBING when it is removed (because
the interface is no longer relevant, for example) then it must decrement
the n_probing count in the associated group (if any).

Otherwise, any remaining announcers in the group will only advance
to state AVAHI_WAITING and never be triggered to advance to state
AVAHI_ANNOUNCING.

Resolves: #383
Signed-off-by: Alan Young <consult.awy@gmail.com>

### DIFF
--- a/avahi-core/announce.c
+++ b/avahi-core/announce.c
@@ -38,6 +38,12 @@ static void remove_announcer(AvahiServer *s, AvahiAnnouncer *a) {
     assert(s);
     assert(a);
 
+    if (a->state == AVAHI_PROBING && a->entry->group) {
+        assert(a->entry->group->n_probing);
+        a->entry->group->n_probing--;
+        avahi_s_entry_group_check_probed(a->entry->group, 1);
+    }
+
     if (a->time_event)
         avahi_time_event_free(a->time_event);
 


pemensik:
Would you know steps for reproducing described issue? How can I check this fixed described bug. Do you know how to trigger it, @awy?

awy:
It is actually really difficult to reproduce because it is so timing related. This might work:
1. Have two active network interfaces: say eth0 and wlan0.
2. Add a new service so that the probing announcer is triggered. Just adding a service file is probably sufficient although my use-case always used the API (Bonjour compat).
3. Within the probing period bring down one of the interfaces.
4. Observe that probing does not complete and announcing does not happen.

The timing interval between 2 and 3 is critical. Too short and probing will not have started before the loss of the interface is handled. Too long and probing will have already completed when the loss of the interface is handled.

Batchyx:
This is almost the same fix as #39.

pemensik:
I agree. I have merged #39, which seems to be addressing the same problem. Please report if that does not work.

PR #387, opened 2022-07-20
Georgian translation

NorwayFun:
Initial Georgian translation


pemensik:
Thank you for the contribution. Do not know Georgian, so merging as it is.

PR #390, opened 2022-07-29
Fix escape representation of non-printable characters

drepper:
The avahi_escape_label function generates a string which uses an escape
sequence to represent non-printable characters.  The representation used
does not conform to shell/C/… rules, though.  It uses a zero-padded
decimal representation.  A space = U+0020 character is therefore printed
as \032.  In shell/C/… strings this represents U+001a.  At least for me
this is highly confusing and I questioned my settings.

Therefore I suggest to use the traditional octal representation.

### DIFF
--- a/avahi-common/domain.c
+++ b/avahi-common/domain.c
@@ -159,9 +159,9 @@ char *avahi_escape_label(const char* src, size_t src_length, char **ret_name, si
                 return NULL;
 
             *((*ret_name) ++) = '\\';
-            *((*ret_name) ++) = '0' + (char)  ((uint8_t) *src / 100);
-            *((*ret_name) ++) = '0' + (char) (((uint8_t) *src / 10) % 10);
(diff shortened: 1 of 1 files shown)


pemensik:
I think octal representation is quite impractical. I would suggest hex escape sequence if possible. I think that is muc more convenient for unprintable characters.

evverx:
I can't say I like this format either but it seems it follows https://datatracker.ietf.org/doc/html/rfc4343#section-2-2. Other than that those strings are most likely consumed by various scripts running something like `avahi-browse -arpt` and some of them probably somehow rely on those escape sequences so it seems switching to anything else can break them.

evverx:
I think it can be closed. avahi itself expects labels to be encoded like that in `avahi_unescape_label` and it normalizes domain names by unescaping and then escaping them so this change breaks that too.

PR #401, opened 2022-10-18
Hide desktop icons by default

gallodustin:
By default, the following files will appear in an applications menu for any common desktop environment, for example gnome. I suggest it would be a better default to hide these icons, so that the average user doesn't see an icon they don't understand or need to use.

- avahi-discover.desktop
- bvnc.desktop
- bssh.desktop

### DIFF
--- a/avahi-python/avahi-discover/avahi-discover.desktop.in.in
+++ b/avahi-python/avahi-discover/avahi-discover.desktop.in.in
@@ -8,3 +8,4 @@ Type=Application
 Icon=network-wired
 Categories=GNOME;System;
 StartupNotify=false
+NoDisplay=true
--- a/avahi-ui/bssh.desktop.in.in
+++ b/avahi-ui/bssh.desktop.in.in
@@ -8,3 +8,4 @@ Type=Application
 Icon=network-wired
 Categories=GNOME;Network;
 StartupNotify=false
+NoDisplay=true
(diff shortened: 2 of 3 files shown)


pemensik:
Why would you want those applications hidden? I think it is possible to uninstall the ui package if not needed. But if you have them, then they should be visible somewhere in launcher. I admit they do not have great user experience, but hiding does not solve that.

gallodustin:
My thought is that the vast majority users who have downloaded these UI applications (e.g. as part of the gnome dependencies) do not need to use them and do not intend to do anything custom. Most users want to use the DNS service as something in the background they don't think about and would consider these icons as extra clutter.

If the issue is with gnome needlessly installing a UI package that is not usually needed, that is a fair point and a different issue.

pemensik:
In Fedora we have all those UI programs in separate subpackage. That should be not dragged by any UI programs as a hard dependency, which could not be installed. If it is, then I think the distribution specific bugs should be filled to make those GUI utilities optional and uninstallable. But I don't think this should change here in upstream, which is eventual source for all distributions. Including those like Fedora, where this is (I think) solved appropriately by packaging.

What is a distribution you are using btw?

gallodustin:
I believe you are right - the root cause is that I included optional dependencies in Arch's avahi package. User issue.

https://archlinux.org/packages/extra/x86_64/avahi/

PR #403, opened 2022-10-27
fix memory leak in avahi-daemon

WeiLiCalix:
The packet is malloced in wide-area.c:336

`    /* We keep the packet around in case we need to repeat our query */
    l->packet = avahi_dns_packet_new(0);
`
But not freed in function 
`void avahi_wide_area_lookup_free(AvahiWideAreaLookup *l) {
`


pemensik:
Could you please ``git rebase -i``, then make fix typo commits just fixups merged to the first commit. Then please ``git push -f``. I think single commit is okay.

pemensik:
Hmm, ``avahi_multicast_lookup_free()`` seems very similar, but it does not have such packet free also. That seems the l->packet should be freed on the other place.

pemensik (changes requested):
Would you be able to use ``git rebase -i`` and merge those typo fixes into just single commit? I think one commit should be enough for this change. It would make history unnecessary scattered.

guillerodriguez:
> Would you be able to use `git rebase -i` and merge those typo fixes into just single commit? I think one commit should be enough for this change. It would make history unnecessary scattered. Use then `git push --force` to update the source branch.

If the commits are ok and you just want them squashed, can’t you just do that while merging? See: https://docs.github.com/es/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits

(Not sure if that option is enabled in this repo)

WeiLiCalix:
@pemensik 
Sorry for late. 
I have created a new PR with a single commit in https://github.com/lathiat/avahi/pull/411

EskoDijk:
I've added my comments to #411 just now.

PR #404, opened 2022-10-28
Fix document for avahi-daemon.conf and avahi-publish

Dieken:
The patch is obvious.

### DIFF
--- a/man/avahi-daemon.conf.5.xml.in
+++ b/man/avahi-daemon.conf.5.xml.in
@@ -204,7 +204,7 @@
       are resolved via unicast DNS. If you want to maintain multiple
       different multicast DNS domains even with this option enabled
       we encourage you to use subdomains of .local, such as
-      "kitchen.local". This option defaults to "yes".</p>
+      "kitchen.local". This option defaults to "no".</p>
     </option>
 
   </section>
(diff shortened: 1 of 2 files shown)


pemensik (approved):
Indeed. ``c->enable_wide_area = 0;`` is in ``avahi_server_config_init()``. I have found that enabled by default only in avahi-test and querier-test

EskoDijk (approved):
Also checked this, looks correct!

Dieken:
Just signed the comments and rebased,  please help get this merged, thanks!

Dieken:
OpenWRT ships that file with that line commented out.

I think when we say default config, we mean what the config is if not explicitly configure at all.

EskoDijk:
Thanks for the additional context @evverx . Since the man page for the config file needs to tell what the default setting values are (in case the config field isn't present in the config file), the proposed edit is still correct. Seems we all agree on that! Whether the default is good or not, is another question.

Dieken:
Could we get this little patch in? This is really obvious minor issue, I don’t mind this patch is rejected, just hope the issue is fixed soon, it’s not worth a month to discuss.

evverx:
@Dieken I don't think this patch has been rejected. Even though personally I think the documentation is somewhat misleading I was outvoted anyway. The PR hasn't been merged yet because the repository is being transferred at the moment (https://lists.freedesktop.org/archives/avahi/2022-December/002587.html) and there are various issues that need addressing first (plus the new year is coming along with vacations, holidays and so on). I think it should be more active next year.

PR #408, opened 2022-11-17
Accept dots in service name

pemensik:
turris.cz device with OS < 6 advertises itself with unusual name, which breaks avahi-discover-standalone. It has two labels service name followed by type, then domain. Handle such unusual cases and start type parsing on the first label leading with '_'. Should work for all valid services. But may result in different breakage if name without _ were advertised.

avahi-browse shows such device as:
``enp0 IPv4 0000000D3000FFF0  foris-controller._mqtt._tcp.local``

This change prevents assertion fail on avahi-discover-standalone and displays the record correctly in it.

### DIFF
--- a/avahi-common/domain.c
+++ b/avahi-common/domain.c
@@ -477,22 +477,6 @@ int avahi_service_name_join(char *p, size_t size, const char *name, const char *
     return AVAHI_OK;
 }
 
-#ifndef HAVE_STRLCPY
-
-static size_t strlcpy(char *dest, const char *src, size_t n) {
-    assert(dest);
-    assert(src);
-
-    if (n > 0) {
-        strncpy(dest, src, n-1);
-        dest[n-1] = 0;
-    }
-
-    return strlen(src);
-}
-
-#endif
-
 int avahi_service_name_split(const char *p, char *name, size_t name_size, char *type, size_t type_size, char *domain, size_t domain_size) {
     enum {
         NAME,
@@ -500,6 +484,7 @@ int avahi_service_name_split(const char *p, char *name, size_t name_size, char *
         DOMAIN
     } state;
     int type_empty = 1, domain_empty = 1;
+    int name_labels = 0;
 
     assert(p);
     assert(type);
@@ -524,9 +509,18 @@ int avahi_service_name_split(const char *p, char *name, size_t name_size, char *
 
         switch (state) {
             case NAME:
-                strlcpy(name, buf, name_size);
+		if (name_labels == 0 || buf[0] != '_') {
+		    /* allow multiple label service name */
+		    if (name_labels > 0)
+			strncat(name, ".", name_size-1);
+                    strncat(name, buf, name_size-1);
+		    name[name_size-1] = 0;
+		    name_labels++;
+                    break;
+		}
+
                 state = TYPE;
-                break;
+                /* fall through */
 
             case TYPE:
 
--- a/avahi-discover-standalone/main.c
+++ b/avahi-discover-standalone/main.c
@@ -322,6 +322,7 @@ int main(int argc, char *argv[]) {
     GtkTreeViewColumn *c;
     gint error;
     AvahiGLibPoll *poll_api;
+    GError *gerror = NULL;
 
     gtk_init(&argc, &argv);
 
@@ -331,7 +332,12 @@ int main(int argc, char *argv[]) {
 
     ui = gtk_builder_new();
     gtk_builder_set_translation_domain(ui, "avahi");
-    gtk_builder_add_from_file(ui, AVAHI_INTERFACES_DIR"avahi-discover.ui", NULL);
+    if (!gtk_builder_add_from_file(ui, AVAHI_INTERFACES_DIR"avahi-discover.ui", &gerror)) {
+	fprintf(stderr, "Failed to load UI interface: %s\n", gerror->message);
+	g_error_free(gerror);
+	g_object_unref(G_OBJECT(ui));
+	return 1;
+    }
     main_window = GTK_WIDGET(gtk_builder_get_object(ui, "main_window"));
     g_signal_connect(main_window, "delete-event", (GCallback) main_window_on_delete_event, NULL);
 


evverx:
@pemensik even though `avahi` is fuzzed on OSS-Fuzz this particular parser isn't covered at all. Given that based on my experiments avahi parsers like that crash a lot I think it would make sense to cover them first before changing anything. I'll send an OSS-Fuzz PR shortly that should make it easier to bring the fuzzing infrastructure to the repository to test PRs when they are opened.

evverx:
-> https://github.com/google/oss-fuzz/pull/9107

evverx:
FWIW with https://github.com/lathiat/avahi/pull/420 merged the testsuite should be able to cover this part of avahi automatically and `domain-test.c` could even be extended when bugs are fixed to make sure `avahi_service_name_split` works as expected.

pemensik:
Hmm, for some reason, at least nss-mdns does not allow names with more than two labels. Not sure if MDNS or DNS-SD rfc even allows more then one label before service type identifier. Perhaps it should be modified just to not crash, but print just some warning.

evverx:
> It has two labels service name followed by type, then domain.

Dots in service instance names should be escaped
>    If client software takes the Instance, Service, and Domain
   portions of a Service Instance Name and internally concatenates them
   together into a single string, then because the Instance portion is
   allowed to contain any characters, including dots, appropriate
   precautions MUST be taken to ensure that DNS label boundaries are
   properly preserved.  Client software can do this in a variety of
   ways, such as character escaping. ... The backslash-escaping allows literal dots in the name
   (escaped) to be distinguished from label-separator dots (not
   escaped)

and it appears those devices don't follow that. I don't think avahi should try to get it around and I don't think it should crash either.

https://github.com/lathiat/avahi/issues/212 is somewhat similar in the sense that there avahi receives PTR records where service names don't start with `_` and trips on that.

it appears skipping PTR records like that with warnings of some kind is the way to go. @EskoDijk I wonder what you think? I can't seem to find any standards saying what exactly should happen when PTR records with unescaped dots like `a.b._qotd._tcp.local` or `b._qotd._tcp.local` during service discovery (https://github.com/lathiat/avahi/issues/212  and https://github.com/lathiat/avahi/issues/251) are received.

EskoDijk:
@evverx PTR records can have a wide range of characters including dots as detailed in Section 4.1.1 of RFC 6763. This is stored as a single DNS label. See also Section 16 of RFC 6762.

At the same time RFC 6763 Section 4.3 recommends a producer of a service name to use backslash escaped characters (for example for '.' and '\') .  But this is not a MUST, just good practice to avoid the issue that the mDNS library of the producing device sees the '.' as a label separator.  One reason for this is that it is also recommended that the human user can modify or select the Service Instance Name.  And this human could type a dot '.' of course. Escaping all these characters is then a safeguard against causing problems in the mDNS library below the UI level.  When the name is being printed on screen by another mDNS client and shown to a user, then escape characters would of course not be printed on screen so all is good.

So an mDNS client/browser/discoverer must be able to handle the full character set. How it would display the discovered names is another question. It could just show the '.'. It could also apply the DNS master file notation of Section 5.1 of RFC 1035. Then it becomes '\.' and a space becomes '\032'.  This notation is also used in RFC 6763 for examples. That might be ok for a technical/debugging type of UI. For an end-user UI, the expectation is that a dot would show as '.', an escaped dot '\.' would show also as dot '.', and a space would just show up as space instead of '\032'. 

So again the RFCs do clearly permit devices to not follow the escaping recommendation! The "... MUST be taken to ensure that DNS label boundaries .... " part is only a requirement for a producer of mDNS names to treat a configured or human-entered Service Instance Name with dots in it correctly as a single label. It's not a requirement to apply escape characters.
Also the RFCs require an mDNS device (consumer/discoverer) to handle the full character set.

evverx:
@EskoDijk got it. Thanks!

> So again the RFCs do clearly permit devices to not follow the escaping recommendation!

I don't think they can expect `avahi` (or any other consumers/discoverers) to split instance service names correctly then. In this particular case it seems to be possible to guess that the dots aren't escaped and look for the first label starting with `_`. On the other hand if `avahi` received, say, `A.B.C.D_qotd._udp._ipp._tcp.local` that heuristic would fail. It could of course switch to another heuristic and extract `_ipp._tcp` but it's fragile as well and breaks when subtypes without underscores are involved: https://github.com/lathiat/avahi/issues/396.

> Also the RFCs require an mDNS device (consumer/discoverer) to handle the full character set.

Agreed. Though my interpretations is that mDNS devices should handle the full character set in the sense that they shouldn't crash on receiving malformed/invalid service names. They are free to reject them with clear log messages though.

pemensik:
I think avahi has configured list of domains it should browse, with only ``local.`` being a default. For avahi-browse it would be anything you pass to -d command line parameter. So I think it should be safe to assume the right most part is a one of domains from a known set. Therefore it might work more reliably if it removed domain from right side and compared it with a domain from known list. I do not think such list of handled domains is propagated now to used function, but it should not be hard to do that. then we could assume 1-2 labels with leading _ would be a type before that domain. And anything before that would be a name of the service. It could have unlimited count of labels.

I am not sure how much time to invest into fixing this case. TurrisOS 6.x has it already fixed. I does not crash *avahi-daemon* or *avahi-browse*, but only *avahi-discover-standalone* GUI tool.

EskoDijk:
@evverx Maybe I don't fully understand the problem of the dots here. That could be caused by a lack of pre-knowledge of the design of Avahi. The idea is that an mDNS implementation would receive a binary DNS-formatted message, parses the individual labels, and because the string "A.B.C.D_qotd._udp" appears in a single DNS label of 17 characters it would automatically know that this is a Service Instance Name without further parsing or heuristics.

So why would we need here heuristics to guess what the Service Instance Name is? Avahi ought to know it already just because it's a single DNS label.

> They are free to reject them with clear log messages though.
Malformed names may be rejected, sure. One example is a name with control characters in it as defined by RFC 6763.
But a name with dots is clearly defined as valid so MUST NOT be rejected by a compliant implementation. It would be really bad if it would do so.

pemensik:
@EskoDijk At least in my case what I were trying to fix, it is not about escaping dot in service name. Instead the service name is 2 labels in dns wire format, before two more labels with service type follows. I think what we should handle somehow is case when name ``A.B.C.D_qotd._udp.local`` appears as a record with 7 separate binary safe labels. It must not crash the application, but may hide invalid names with some warning. I am not sure they are invalid however.

Presentation of those labels are a bit different thing. I think we never need to specify service by its name. I am not sure if we have to escape ``A.B.C.D`` into ``A\.B\.C\.D`` if that were just single label. If that is always just presented to the user, I think it does not make a difference.

evverx:
> So why would we need here heuristics to guess what the Service Instance Name is?

@EskoDijk apologies. I should have been more specific. The problem is that `avahi_service_name_split` can't split service names like `A.B._qotd._tcp.local`. At this point it assumes that the first label where the instance name should be is properly escaped so if it receives something like `A\.B._qotd._tcp.local` or `A\046B._qotd._tcp.local` it works but when it receives `A.B._qotd._tcp.local` it thinks that the instance name is `A`, the domain is `B._qotd._tcp.local` and the type is empty. The heuristic is needed to guess where the service name is when the instance name isn't properly escaped (or isn't escaped at all).

> Therefore it might work more reliably if it removed domain from right side and compared it with a domain from known list.

@pemensik I think it's an interesting idea. I'm not sure how it mixes with escaped labels though.

evverx:
> I am not sure how much time to invest into fixing this case. 

@pemensik `avahi_service_name_split` is also involved in https://github.com/lathiat/avahi/issues/212 and this PR can also fix that issue so I think of it as a way to fix `avahi_service_name_split` once and for all. Locally I fixed that issue by rejecting PTR records like that with clear log messages but if it's somehow possible to make it work it would probably be even better.

pemensik:
Okay, finally found it. It is specified in [RFC 6763, section 4.1.1](https://www.rfc-editor.org/rfc/rfc6763.html#section-4.1.1), where it says only one instance is used as an identifier. That means on Turris 5.x it violated that RFC, because those name contained two labels before the service type.

As an invalid name I think we do not have to display it, but we have to ensure it does not crash the whole program. We should be able to log it as warning or lower severity. Ideally once per cache insertion, not once per received response.

Section [4.1.2](https://www.rfc-editor.org/rfc/rfc6763.html#section-4.1.2) also states service type is always two labels. Not a should or may is used. So we cannot even make service type longer. Both parts have to start with ``_``.

That means this my proposal is not necessary. We just have to be able to skip invalid services, if they appear on the network.

Just for a reference, this was reported to [Turris #845](https://gitlab.nic.cz/turris/os/packages/-/issues/845)

EskoDijk:
@pemensik Indeed that's the reference saying that it is one label only:

>    This <Instance> portion of the Service Instance Name is stored
>    directly in the DNS as a single DNS label of canonical precomposed
>    UTF-8

> Section 4.1.2 also states service type is always two labels. Not a 
> should or may is used. So we cannot even make service type longer. Both parts have to start with _.

The section also points to Section 7 for more details, where 7.1 explains that subtype can be used. For example:  `MyPrinterName._printer._sub._http._tcp.local` which refers to a web page of a printer. So then we get more DNS labels for the Service name. So it seems the parsing for the Service part takes either 2 or 4 DNS labels. The final label is always _udp or _tcp if I'm correct. "_udp" was defined as a name for "everything else not being TCP" so I believe the intention was to stick only to these two labels forever. That provides a way to detect at least where the domain name part begins.

EskoDijk:
@evverx Ok thanks for explaining the issue. So it's more caused by Avahi's internal representation of a PTR record in an AvahiRecord, it seems. It doesn't keep the DNS label information which would make the parsing part much easier. For example you could then take first DNS label as the Service Instance Name.

Maybe an additional extra `size_t` field could be considered to mark then length of the first DNS label string in the `record->data.ptr.name` field. That would avoid messing around with heuristics which maybe don't catch all cases.

pemensik:
Hmm, interesting. [Section 7.1](https://www.rfc-editor.org/rfc/rfc6763.html#section-7.1) allows also subtypes, which *do not* have to start with ``_``. So service type can be either 2 labels or 4. If ``<subtype>._sub.<type>._tcp`` is used, then subtype is allowed to omit leading ``_``. Current code is not able to handle those subtypes correctly, it does not handle subtypes at all.

evverx:
>  it does not handle subtypes at all

It kind of does in a separate function where labels without leading underscores are rejected outright https://github.com/lathiat/avahi/blob/335c7bb8c626468925f5b13a86805e1a63f967f6/avahi-common/domain.c#L319-L320

pemensik:
Yes, exactly. That is why I think it would be more reliable to remove domain part from the right, because we should have a set of supported domains already. Anyway, the second label of type is rather strict, so we can find just for ``_tcp`` or ``_udp`` from the right side and then test labels left from it.

evverx:
I think it would make sense to take a look at what for example bonjour does. From https://github.com/apple-oss-distributions/mDNSResponder/blob/806254210edec4cab01f794f6fd28658aa6ba59d/mDNSShared/dns_sd.h#L804
```
 * The exceptions, that do not use escaping, are the routines where the full
 * DNS name of a resource is broken, for convenience, into servicename/regtype/domain.
 * In these routines, the "servicename" is NOT escaped. It does not need to be, since
 * it is, by definition, just a single literal string. Any characters in that string
 * represent exactly what they are. The "regtype" portion is, technically speaking,
 * escaped, but since legal regtypes are only allowed to contain US ASCII letters,
 * digits, and hyphens, there is nothing to escape, so the issue is moot.
 * The "domain" portion is also escaped, though most domains in use on the public
 * Internet today, like regtypes, don't contain any characters that need to be escaped.
 * As DNS-SD becomes more popular, rich-text domains for service discovery will
 * become common, so software should be written to cope with domains with escaping.
```
...
```
 * For most software, these issues are transparent. When browsing, the discovered
 * servicenames should simply be displayed as-is. When resolving, the discovered
 * servicename/regtype/domain are simply passed unchanged to DNSServiceResolve().
 * When a DNSServiceResolve() succeeds, the returned fullname is already in
 * the correct format to pass to standard system DNS APIs such as res_query().
 * For converting from servicename/regtype/domain to a single properly-escaped
 * full DNS name, the helper function DNSServiceConstructFullName() is provided.
```

I also found the place where it concatenates strings into full names: https://github.com/apple-oss-distributions/mDNSResponder/blob/bad6277713e5132b78f4c47328aa2895e08f3b12/mDNSShared/dnssd_clientlib.c#L99 but it's somewhat tangential here.

evverx:
I took a quick look at bonjour and it appears it goes left to right through unescaped labels without cutting off domains and rejects all those labels if the second label doesn't start with `_` (or if the first label doesn't start with `_` if it's a PTR record for `_services._dns-sd._udp.domain`). I'm not sure when exactly it splits them though. If it happens after it gets `SRV` and `TXT` records it would mean that it doesn't reject names like `A.B._qotd._tcp.local` as long as it can actually get their `SRV` and `TXT` records. It seems it most likely rejects the devices from https://github.com/lathiat/avahi/issues/251 because there PTR records for `_services._dns-sd._udp.local` don't start with `_`. I haven't actually run anything though so I might be wrong.

evverx:
I think it can be closed. I don't think the parser should be relaxed. If anything it should be stricter to avoid all those crashes where clients expect valid service names and get all sorts of gibberish instead.

PR #409, opened 2022-11-18
Make logging in avahi not limited to fixed length

### LINKED ISSUE #402: Log messages cut off after 256 characters
https://github.com/lathiat/avahi/blob/master/avahi-core/log.c#L35 cuts off log messages after 256 characters which renders especially debug logging less useful.

pemensik:
When logging to stderr, there is no reason to have any maximal length. For a function logging first try formatting the string into 2k buffer. If even that is not enough, allocate temporary buffer in heap and use it for formatting.

Fixes #402

### DIFF
--- a/avahi-core/log.c
+++ b/avahi-core/log.c
@@ -23,6 +23,7 @@
 
 #include <stdio.h>
 #include <stdarg.h>
+#include <stdlib.h>
 
 #include "log.h"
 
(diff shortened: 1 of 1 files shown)


evverx:
The right fix here would probably be to keep this particular function as is to prevent old clients some of which can expect 256 characters from breaking/buffer overflows and introduce another function that could handle more than 256 characters and switch avahi-daemon to it. I think this PR can be closed.

PR #411, opened 2022-11-28
Fix memory leak

WeiLiCalix:


pemensik (approved):

pemensik:
a new PR were not necessary, it would work fine if you just pushed changes again to your repository and github would do the rest. But okay.

EskoDijk (changes requested):
The proposed code was not intended by the creators, I think.  The reason is the following:
* instead of cleaning up and freeing immediately, two flags are set: l->dead and l->engine->cleanup_dead .
* these flags make sure that later on lookup_destroy() is called.
* in function lookup_destroy() in the same file, there is already the proper code for deallocating the packet.
* in that same code also the other members of the AvahiWideAreaLookup struct are properly deallocated, in particular key and cname_key.
* also the entire structure itself is freed there in the last line: avahi_free(l)
* the lookup_destroy() code is called periodically on a timer
* and also this code is called indirectly from browse.c in lookup_wide_area_callback() when the wide-area lookup is done. It calls lookup_unref(l) which eventually calls the cleanup function lookup_destroy()

So to me the current code looks okay  and there's an intentional design to not clean up the packet immediately but let the lookup_destroy() code do that, later. 

Similar design pattern is used for multicast lookup.

So I can't assume that there actually is a memory leak yet. In what situation could the memory leak occur i.e. when are the cleanup routines as I outlined above not called? We need to be first sure about this before making such a change.

pemensik:
Yes, you are right. Function ``sender_timeout_callback`` relies on l->packet being present. It seems unref should mark the request just dead, but leave it for later destruction. Perhaps allow retries on timeouts.

Thank you for spotting it, @EskoDijk. @WeiLiCalix would you know, how to trigger the issue this change is attempting to fix? Can it be spotted on browsing *dns-sd.org* domain?

WeiLiCalix:
To trigger the memory leak, you can do the command, which called the dns-sd query every 10 seconds. 

watch -n 10 avahi-browse-domains --domain=dns-sd.org -arpt

Then check the memory footprint.
cat /proc/`pidof avahi-daemon`/status |grep -i Vm

guillerodriguez:
@pemensik May I suggest that you approach existing PRs with some caution? I would say that bugs must be demonstrable and reproducible  before a proposed “fix” is considered, and the fix itself should be well understood before changes are approved.

WeiLiCalix:
1, In function avahi_wide_area_lookup_free, the memory in `l->packet ` was not freed.
image

2, In function lookup_unref, then input argment l->wide_area is set to null.
image

3, when the callstack is `lookup_unref->avahi_wide_area_lookup_free,` and then return, it has no chance to free the packet memory.

guillerodriguez:
The packet is freed in `lookup_destroy` which is called periodically during normal operation (`cleanup_time_event_callback` -> `avahi_cleanup_dead_entries` -> `avahi_browser_cleanup` ->  `avahi_wide_area_cleanup` -> `lookup_destroy`). It is also called on shutdown (`avahi_server_free ` -> `avahi_wide_area_engine_free` -> `lookup_destroy`).

In `lookup_unref`, `avahi_wide_area_lookup_free` is only called if the refcount falls down to zero. Have you verified that this can actually happen while there are dead packets that have not yet been deallocated?

WeiLiCalix:
Yes, when do below commands on console
`watch -n 10 avahi-browse-domains --domain=dns-sd.org -arpt`
about 700 MB memory is used by avahi-daemon.

WeiLiCalix:
test steps:
1, added debug log in the source code, see line 128, 133, 135
image

2,  did the command on console:
image

3, checked the log, we can see the line 128, 133, 135. so the 
image

4, about 5 minutes later, checked the avahi-daemon memory usage, 4 times. The VmSize, VmRss and etc all are increasing.
image

guillerodriguez:
Even if there is a leak, I don't think the proposed fix is correct; please check the comment from @EskoDijk [above](https://github.com/lathiat/avahi/pull/411#pullrequestreview-1196495834): the code sets flags l->dead and l->engine->cleanup_dead which are meant to be checked from `lookup_destroy()`, so the intention was obviously not to free the packet right away.

pemensik:
Oh, please use 
```
\```c
int C(); // Remove \backslash
\```
```
blocks instead of images. Use links to the code repo to point to correct functions. Those records should be freed by [avahi_browser_cleanup](https://github.com/lathiat/avahi/blob/master/avahi-core/browse.c#L604), called conditionally from [avahi_cleanup_dead_entries](https://github.com/lathiat/avahi/blob/master/avahi-core/entry.c#L142).

``l->wide_area`` should not be used for freeing packets, but list maintained on engine should. Because there are some conditions in the way, I suspect cleanup might not be scheduled correctly. Have you tried adding logs to avahi_wide_area_cleanup or lookup_destroy, whether they are always called? It might make sense to log number of ``e->lookups`` entries if we suspect they are left allocated.

WeiLiCalix:
In function lookup_unref(AvahiSRBLookup *l):
if l->wide_area is not null, then it will be freed by calling avahi_wide_area_lookup_free(l->wide_area) and the l->wide_area is set to null.
Where and when the l->wide_area->packet should be freed? Nowhere records the pointer anymore AFTER l->wide_area setting to null.

pemensik:
I have played a bit with *gdb*, but at least with my Fedora package there is always matching call to avahi_wide_area_lookup_new, followed by avahi_wide_area_lookup_free, then after few seconds also to lookup_destroy in wide-area.c. If there are any leaks, then they should not happen in this part.

I don't think the pointer is just lost. It would be visible when running avahi-daemon under valgrind, but I have not seen report about leaked memory. It might be bloating its memory usage, but it seems not at cost of unfreed memory.

The packet is still freed from [cleanup](https://github.com/lathiat/avahi/blob/master/avahi-core/wide-area.c#L320), few seconds after avahi_wide_area_lookup_free is called on it.

WeiLiCalix:
@pemensik 
Do you execute the command ` avahi-browse-domains --domain=dns-sd.org -arpt` to make a DNS-SD wide area query while you do the gdb debug?

pemensik:
@WeiLiCalix Sure! I am running that  command to trigger new query, which then gets queried. Not too often, because I am checking  the output just by looking at it. Tried even running it with watch, but at least package ``avahi-0.8-18.fc37.x86_64`` does not seem to grow visibly. How fast is it growing to you? I can see some increases, but seems quite minimal. It is probably just normal daemon running.

WeiLiCalix:
@pemensik 
I use the source code from avahi-0.8.tar.gz and my machine is arm32 platform, OpenWrt 19.07

The query command avahi-browse-domains --domain=dns-sd.org -arpt will get about 60 dns records (60 lines of output).
The memory seems increase quickly.
[router] / # cat /proc/`pidof avahi-daemon`/status | grep -i vm
VmPeak:   542448 kB
VmSize:   542448 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:     86296 kB
VmRSS:     86296 kB
VmData:   540760 kB
VmStk:       132 kB
VmExe:        96 kB
VmLib:      1420 kB
VmPTE:       538 kB
VmSwap:        0 kB

pemensik:
Your VmData is ridiculously high. How long has been that process running? Can you please try with source from the master branch? There are some fixes merged after 0.8 release, but not yet contained in a new release. Fedora backports some of them. But haven't seen any memory leak fix among them.

My stats of avahi:

```
Every 2,0s: grep ^Vm /proc/631170/status                                           pemensik-pad: Tue Nov 29 18:52:01 2022

VmPeak:    44420 kB
VmSize:    43684 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:     14812 kB
VmRSS:     14696 kB
VmData:    37140 kB
VmStk:       132 kB
VmExe:        80 kB
VmLib:      4216 kB
VmPTE:       128 kB
VmSwap:        0 kB
```

How many entries would report ``avahi-browse -ac``? I can try also testing that on raspberry pi, but doubt the architecture is responsible. I have no device running OpenWrt, so I cannot test the package itself.

EskoDijk:
In my case, on Ubuntu 22.04 running the latest master branch code, VmSize grows until it reaches the below sizes and then it stays like that:

```
VmPeak:	   32004 kB
VmSize:	   32004 kB
VmLck:	       0 kB
VmPin:	       0 kB
VmHWM:	    8324 kB
VmRSS:	    8324 kB
VmData:	   22872 kB
VmStk:	     132 kB
VmExe:	     104 kB
VmLib:	    6040 kB
VmPTE:	     100 kB
VmSwap:	       0 kB
```

evverx:
FWIW with this patch applied avahi crashes with
```
==98650==
==98650== Process terminating with default action of signal 6 (SIGABRT): dumping core
==98650==    at 0x49CCC0C: __pthread_kill_implementation (in /usr/lib64/libc.so.6)
==98650==    by 0x497C985: raise (in /usr/lib64/libc.so.6)
==98650==    by 0x49667F3: abort (in /usr/lib64/libc.so.6)
==98650==    by 0x496671A: __assert_fail_base.cold (in /usr/lib64/libc.so.6)
==98650==    by 0x4975535: __assert_fail (in /usr/lib64/libc.so.6)
==98650==    by 0x488F12E: avahi_dns_packet_free (dns.c:126)
==98650==    by 0x48944DA: lookup_destroy (wide-area.c:288)
==98650==    by 0x4894D94: avahi_wide_area_engine_free (wide-area.c:632)
==98650==    by 0x487EBE1: avahi_server_free (server.c:1545)
==98650==    by 0x407ABB: run_server (main.c:1301)
==98650==    by 0x407ABB: main (main.c:1686)
==98650==
```

Also I have been running running avahi under Valgrind and ASan/UBSan for a couple of days (for some unrelated testing purposes) and neither of them has been able to catch this particular memory leak

WeiLiCalix:
@pemensik 
`avahi-browse -ac` returns with no output in my device.
I check the difference between current master and release 0.8, it doesn't differ greatly, only some null pointer checking.

I added some log in function avahi_dns_packet_new as below:
```c
AvahiDnsPacket* avahi_dns_packet_new(unsigned mtu) {
    AvahiDnsPacket *p;
    size_t max_size;

    if (mtu <= 0)
        max_size = AVAHI_DNS_PACKET_SIZE_MAX;
    else if (mtu >= AVAHI_DNS_PACKET_EXTRA_SIZE)
        max_size = mtu - AVAHI_DNS_PACKET_EXTRA_SIZE;
    else
        max_size = 0;

    if (max_size < AVAHI_DNS_PACKET_HEADER_SIZE)
        max_size = AVAHI_DNS_PACKET_HEADER_SIZE;

    if (!(p = avahi_malloc(sizeof(AvahiDnsPacket) + max_size)))
        return p;

    avahi_log_error("avahi_dns_packet_new, malloc size=%u+%u, p=%p", sizeof(AvahiDnsPacket), max_size, p );
```

Notice, I logged: 
```c
avahi_log_error("avahi_dns_packet_new, malloc size=%u+%u, p=%p", sizeof(AvahiDnsPacket), max_size, p );
```

Very interesting, the memory block with big size: 24+65813 Bytes, whose pointer address continuously increases.
Line   222: Tue Nov 29 19:39:43 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+65813, p=0x1a10d0
...
Line 35379: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+65813, p=0xf460450
Line 35472: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+65813, p=0xf4708c0

But the small memory block, such as the 24+132 Bytes block, the pointer of them can be seen the same. That means the small block memory is freed and then malloced again next time.
Line 35470: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+132, p=**0xf0a9590**
Line 35474: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+132, p=**0xf0a9590**
Line 35476: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+132, p=**0xf0a9590**
Line 35478: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+132, p=**0xf0a9590**
Line 35480: Tue Nov 29 19:49:19 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+132, p=**0xf0a9590**
See the same phenomenon on other small size:  24+128 Bytes
	Line 35130: Tue Nov 29 19:49:13 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+128, p=0xd0c67e0
	Line 35136: Tue Nov 29 19:49:13 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+128, p=0xd0c67e0
	Line 35140: Tue Nov 29 19:49:13 2022 daemon.err avahi-daemon[2024]: avahi_dns_packet_new, malloc size=24+128, p=0xd0c67e0

So I am sure in my device, the big size 24+65813 packet is not freed.

evverx:
@WeiLiCalix could you install `valgrind`, replace
```
ExecStart=/usr/sbin/avahi-daemon -s
```
with
```
/usr/bin/valgrind --leak-check=full /usr/sbin/avahi-daemon -s
```
in the `avahi.service`, run `systemctl daemon-reload`, `systemctl restart avahi-daemon`, trigger that memory leak somehow, stop `avahi` with `systemctl stop avahi-daemon` and copy-paste the output from the journal. It should look like
```
==65371== HEAP SUMMARY:
==65371==     in use at exit: 7,323 bytes in 54 blocks
==65371==   total heap usage: 52,067 allocs, 52,013 frees, 31,316,421 bytes allocated
==65371==
...
==65371== LEAK SUMMARY:
==65371==    definitely lost: 336 bytes in 4 blocks
==65371==    indirectly lost: 456 bytes in 11 blocks
==65371==      possibly lost: 0 bytes in 0 blocks
==65371==    still reachable: 6,531 bytes in 39 blocks
==65371==         suppressed: 0 bytes in 0 blocks
==65371== Reachable blocks (those to which a pointer was found) are not shown.
==65371== To see them, rerun with: --leak-check=full --show-leak-kinds=all
```

WeiLiCalix:
@evverx 

It seems there is not leak in the valgrind report
```c
==8889==
==8889== HEAP SUMMARY:
==8889==     in use at exit: 6,851 bytes in 75 blocks
==8889==   total heap usage: 204,760 allocs, 204,685 frees, 96,946,951 bytes allocated
==8889==
==8889== 36 bytes in 3 blocks are definitely lost in loss record 1 of 4
==8889==    at 0x4891BC8: malloc (in /usr/lib/valgrind/vgpreload_memcheck-arm-linux.so)
==8889==
==8889== LEAK SUMMARY:
==8889==    definitely lost: 36 bytes in 3 blocks
==8889==    indirectly lost: 0 bytes in 0 blocks
==8889==      possibly lost: 0 bytes in 0 blocks
==8889==    still reachable: 6,815 bytes in 72 blocks
==8889==         suppressed: 0 bytes in 0 blocks
==8889== Reachable blocks (those to which a pointer was found) are not shown.
==8889== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==8889==
==8889== Use --track-origins=yes to see where uninitialised values come from
==8889== For lists of detected and suppressed errors, rerun with: -s
==8889== ERROR SUMMARY: 133120 errors from 6 contexts (suppressed: 0 from 0)

```
But the VmRSS and VmData shows it used much memory. When I applied the fix, the memory was not leaking obviously and the avahi-daemon is not crash.

evverx:
> It seems there is not leak in the valgrind report

I think it means that the memory was freed at some point. If it hadn't been `valgring` would have complained that it was definitely lost.

> I applied the fix, the memory was not leaking obviously and the avahi-daemon is not crash

I'm not sure why it didn't crash because with this patch applied `avahi` tries to free the same memory twice (and unlike that memory leak I can reproduce the newly introduced crash reliably on 64 and 32 bit VMs.

> But the VmRSS and VmData shows it used much memory

I wonder if the kernel is heavily patched? Could it be that it fails to account memory properly?

Another option would be that avahi holds some memory much longer than it should and frees it when it's stopped but I haven't been able to reproduce it either.

evverx:
Just out of curiosity since it seems to be a router I wonder what kind of `libc` is used there? Does it come with some custom memory allocator that is supposed to be optimized for devices like that?

evverx:
I managed to trigger an OOM so it appears avahi accumulates a lot of memory somewhere. It's kind of bad but I think it's better than the crash
```
Registering new address record for ::1 on lo.*.
Registering new address record for 127.0.0.1 on lo.IPv4.
Server startup complete. Host name is a6976252a23a.local. Local service cookie is 2614732498.
Out of memory, aborting ...
==8393==
==8393== Process terminating with default action of signal 6 (SIGABRT): dumping core
==8393==    at 0x49D31D7: __pthread_kill_implementation (pthread_kill.c:44)
==8393==    by 0x49820D0: raise (raise.c:26)
==8393==    by 0x496B269: abort (abort.c:79)
==8393==    by 0x485A614: oom.cold (malloc.c:58)
==8393==    by 0x485A7C4: xmalloc (malloc.c:69)
==8393==    by 0x485A7C4: avahi_malloc (malloc.c:107)
==8393==    by 0x48856A2: avahi_dns_packet_new (dns.c:53)
==8393==    by 0x488B3F8: avahi_wide_area_lookup_new (wide-area.c:237)
==8393==    by 0x4880A48: lookup_start (browse.c:327)
==8393==    by 0x4880A48: lookup_go (browse.c:392)
==8393==    by 0x4880A48: lookup_go (browse.c:374)
==8393==    by 0x4881242: defer_callback (browse.c:474)
==8393==    by 0x486E92E: expiration_event (timeeventq.c:94)
==8393==    by 0x485DF31: timeout_callback (simple-watch.c:447)
==8393==    by 0x485EDEE: avahi_simple_poll_dispatch (simple-watch.c:570)
==8393==
==8393== HEAP SUMMARY:
==8393==     in use at exit: 3,112,218,331 bytes in 1,122,513 blocks
==8393==   total heap usage: 8,052,918 allocs, 6,930,405 frees, 3,258,749,389 bytes allocated
==8393==
...
Thread 1: status = VgTs_Empty (lwpid 8393)
client stack range: [0xFE94E000 0xFE950FFF] client SP: 0xFE9508BC
valgrind stack range: [0x82956000 0x82A55FFF] top usage: 17284 of 1048576
==8393==
==8393==     Valgrind's memory management: out of memory:
==8393==        newSuperblock's request for 4493312 bytes failed.
==8393==        4,190,281,728 bytes have already been mmap-ed ANONYMOUS.
==8393==     Valgrind cannot continue.  Sorry.
```

WeiLiCalix:
> It is musl libc

pemensik:
I think valgrind's [massif](https://valgrind.org/docs/manual/ms-manual.html) tool might help with finding what is wrong. I tried it running as ``sudo valgrind --tool=massif /home/pemensik/upstream/avahi/avahi-daemon/.libs/avahi-daemon --debug --no-chroot``. It was the first time I used massif tool, but it reported *wide-area.c:237* to be responsible for *~70%* of allocated memory.

And indeed, with mtu == 0 it allocates maximal TCP packet size, which is kept for the lifetime of dns query. But majority of queries are quite small and this big packet is wasting of memory. If there are more dns queries, then indeed it would have high memory demands. I am quite sure that should be lower. Especially when the query is only retransmitted as it is. I think it is never modified since creation. So the size of memory used should be adjusted just to minimum for given packet.

WeiLiCalix:
@pemensik Yes, it is. The wide-area request more than 64KB memory. 
In most cases, it only uses the first 4kb page, so the VmSize inflates much quickly than VmRss. This is what I have seen in my router.

BTW, my memory allocator is the musl libc's malloc.

WeiLiCalix:
> The packet is freed in `lookup_destroy` which is called periodically during normal operation (`cleanup_time_event_callback` -> `avahi_cleanup_dead_entries` -> `avahi_browser_cleanup` -> `avahi_wide_area_cleanup` -> `lookup_destroy`). It is also called on shutdown (`avahi_server_free ` -> `avahi_wide_area_engine_free` -> `lookup_destroy`).
> 
> In `lookup_unref`, `avahi_wide_area_lookup_free` is only called if the refcount falls down to zero. Have you verified that this can actually happen while there are dead packets that have not yet been deallocated?

Thanks for your reply. 

I agree that the packet can be freed in `cleanup_time_event_callback` if it is periodic triggered or in `avahi_server_free` when avahi-daemon quits.

The problem in my router is that I can't see the periodic calling of `avahi_cleanup_dead_entries` (I added log in it). That is why the growing memory footprint was found in my router.  

Theoretically, the memory is not lost because it is still referenced by the pointer, but it is not freed during app lifetime. 

I will check the reason why my router fails to call the `avahi_cleanup_dead_entries ` periodically.

evverx:
> I will check the reason why my router fails to call the avahi_cleanup_dead_entries  periodically.

@WeiLiCalix I think they probably get disabled somewhere along the way by analogy with https://github.com/lathiat/avahi/pull/366 and I suspect the right fix here should look like that PR. It would be great if you could track it down and hopefully the CI should be ready soon to test this PR because I don't think it should be necessary to checkout PRs locally and run `avahi` manually to catch attempts to free memory twice.

evverx:
> BTW, my memory allocator is the musl libc's malloc.

@WeiLiCalix got it. Thanks! At the time I hadn't reproduced the issue so I was kind of guessing what else could have triggered it. Now that I can trigger it reliably it should be safe to say that the allocator has nothing to do with it and it's an avahi issue.

WeiLiCalix:
@evverx
 
Something interesting happened recently. When my router's interfaces changes, as you my know, the avahi-daemon will receive the netlink event and in the following processing, I see the call stack:
(`cleanup_time_event_callback` -> `avahi_cleanup_dead_entries` -> `avahi_browser_cleanup` -> `avahi_wide_area_cleanup` -> `lookup_destroy`)
Certainly avahi-daemon crashes due to double free in my previous fix.

The problem in my router is that there is not a `cleanup_time_event_callback` be called periodically, except when my router's interfaces change. 

I noticed that the `avahi_cleanup_dead_entries` can be called in `mcast_socket_event` or `legacy_unicast_socket_event`. If the test environment has some mDNS devices, the memory allocated by "wide area look up engine" will be freed in the 2 functions. But My test environment didn't have a mDNS device, I think that is why I had seen the growing memory footprint when sending lots of widre area queries.


I have a fix in the wide_area.c yesterday, it tests good.  All the codes are based on 0.8 release.
```
static void socket_event(AVAHI_GCC_UNUSED AvahiWatch *w, int fd, AVAHI_GCC_UNUSED AvahiWatchEvent events, void *userdata) {
    AvahiWideAreaLookupEngine *e = userdata;
    AvahiDnsPacket *p = NULL;

    if (fd == e->fd_ipv4)
        p = avahi_recv_dns_packet_ipv4(e->fd_ipv4, NULL, NULL, NULL, NULL, NULL);
    else {
        assert(fd == e->fd_ipv6);
        p = avahi_recv_dns_packet_ipv6(e->fd_ipv6, NULL, NULL, NULL, NULL, NULL);
    }

    if (p) {
        handle_packet(e, p);
        avahi_dns_packet_free(p);

+        /*
+        after processing a DNS response packet, we can do some clean-up work.
+       the same logic is found in mcast_socket_event and legacy_unicast_socket_event
+      */
+        avahi_cleanup_dead_entries(e->server);
    }
}
```

I will try the #366 to see if it can fix my issue.
thanks!

WeiLiCalix:
@evverx 
I just test the fix of #366 In my test environment, the memory still growing quickly.

avahi-daemon memory footprint:

at startup:
```Thu Dec 22 00:45:53 CST 2022
VmPeak:     1964 kB
VmSize:     1964 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:       908 kB
VmRSS:       908 kB
VmData:      156 kB
VmStk:       132 kB
VmExe:        96 kB
VmLib:      1536 kB
VmPTE:        12 kB
VmSwap:        0 kB
Thu Dec 22 00:45:53 CST 2022
Every 60s: ./test.sh           
```

after 200 tests:
```
Thu Dec 22 00:59:54 CST 2022
VmPeak:   493472 kB
VmSize:   493472 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:     69808 kB
VmRSS:     69808 kB
VmData:   491664 kB
VmStk:       132 kB
VmExe:        96 kB
VmLib:      1536 kB
VmPTE:       492 kB
VmSwap:        0 kB
Thu Dec 22 00:59:54 CST 2022

```

test script:
```
for i in `seq 100000`
do
avahi-browse-domains --domain=dns-sd.org -arpt
echo ""
echo ""
echo "========  $i test #############  "
echo ""
echo ""
done
```

WeiLiCalix:
@pemensik 
Is it proper to add the `avahi_cleanup_dead_entries()` in function `socket_event` in my previous comment [https://github.com/lathiat/avahi/pull/411#issuecomment-1362472060](url) 

The same logic exists in the `mcast_socket_event` and `legacy_unicast_socket_event`, which I think are similar functions as `socket_event`.

guillerodriguez:
@WeiLiCalix :

> The problem in my router is that there is not a cleanup_time_event_callback be called periodically, except when my router's interfaces change.

If that is the problem, wouldn't it be better to investigate why that is happening, instead of patching the code to work around the issue ?

WeiLiCalix:
@guillerodriguez 
Yes, I agree if the cleanup_time_event_callback should be called periodically. 
But I still not understand, why the `avahi_cleanup_dead_entries` is called in `mcast_socket_event` but not called in `socket_event`, In my option, they have similar processing logic, the former is the event callback to process multicast DNS query packet and the latter is to process the wide area DNS query response.

pemensik:
``avahi_cleanup_dead_entries`` is called from 3 places:

- timer cleanup_time_event_callback initiated from ``schedule_cleanup()``, called in ``avahi_s_entry_group_free()`` and ``avahi_s_entry_group_reset()``.
- [mcast_socket_event()](https://github.com/lathiat/avahi/blob/d1e71b320d96d0f213ecb0885c8313039a09f693/avahi-core/server.c#L1067)
- [legacy_unicast_socket_event()](https://github.com/lathiat/avahi/blob/d1e71b320d96d0f213ecb0885c8313039a09f693/avahi-core/server.c#L1103), both in *avahi-core/server.c*
- I admit [socket_event()](https://github.com/lathiat/avahi/blob/335c7bb8c626468925f5b13a86805e1a63f967f6/avahi-core/wide-area.c#L553) seems very similar to them. I see no direct connection to group cleaning, which is the only other caller of dead entries cleanup.

pemensik:
Those cleanups were added in commit f7e3b4fab6ba1bc98c6dbe859ccdef4fde6c5db2, quite long in history. It is possible noone has noticed it should have been used also in *wide-area.c*. It should be safe also from wide-area.c, because it cleans only records marked dead. It should be unnecessary in the worst case.

pemensik:
I have created change in PR #431 to log pending lookups in reaction to SIGUSR1 signal. But even when I set ``allowed_interfaces=`` in avahi-daemon.conf, those lookups are always freed after some time. @WeiLiCalix are you able to reproduce that issue with PR #431 included? But even after disabling all interfaces, which I think effectively turns off MDNS, those dead lookups are removed in my case. Would you be able to prepare steps, when those lookups are not removed after few seconds?

evverx:
@pemensik it should be possible to reproduce it by disabling multicast with something like
```
ip link set <iface1> multicast off
ip link set <iface2> multicast off
...
```
turning `avahi` into a resolver by setting `disable-publishing=yes` and restarting it. Another option would be to run `avahi` in a podman container. It should create a single interface without multicast that still can be used to resolve wide-area services through the resolver on the host.

WeiLiCalix:
I agree the cleanup work done by `avahi_cleanup_dead_entries`, as @pemensik mentioned above in 3 places:

#1,
timer `cleanup_time_event_callback` initiated from `schedule_cleanup()`, called in `avahi_s_entry_group_free()` and `avahi_s_entry_group_reset()`.
But in my test environment, I haven't seen the `avahi_s_entry_group_free()` and `avahi_s_entry_group_reset()` was called during its startup to stable, **which means the cleanup work is not a routine work in avahi-daemon.**
Only under some circumstances the timer callback function can be registerred and the cleanup work can be done finally.

#2, in mMDS processing

#3, in legacy_unicast processing

As @evverx mentioned above, which is just how I used the avahi-daemon, "only used it to do the wide area DNS-SD query", so my case is that I haven't have the conditions below:
#1 a routine "cleanup work" (except that I manually restart my router interfaces to trigger the netlink event and finally to call the `avahi_s_entry_group_free()` and `avahi_s_entry_group_reset()`, or in other cases)
#2 multicast processing
#3 legacy_unicast processing.

there are lots of dead entries seen in my private debug log yesterday.

I thought the cleanup work would be a routine work in the main process, but it wasn't.


Here is the config.
```
[server]
#host-name=foo
#domain-name=local
#browse-domains=0pointer.de, zeroconf.org
use-ipv4=yes
use-ipv6=yes
#allow-interfaces=eth0
#deny-interfaces=eth1
#check-response-ttl=no
#use-iff-running=no
#enable-dbus=yes
#disallow-other-stacks=no
#allow-point-to-point=no
#cache-entries-max=4096
#clients-max=4096
#objects-per-client-max=1024
#entries-per-entry-group-max=32
ratelimit-interval-usec=1000000
ratelimit-burst=1000

[wide-area]
enable-wide-area=yes

[publish]
#disable-publishing=no
#disable-user-service-publishing=no
#add-service-cookie=no
#publish-addresses=yes
publish-hinfo=no
publish-workstation=no
#publish-domain=yes
#publish-dns-servers=192.168.50.1, 192.168.50.2
#publish-resolv-conf-dns-servers=yes
#publish-aaaa-on-ipv4=yes
#publish-a-on-ipv6=no

[reflector]
#enable-reflector=no
#reflect-ipv=no
#reflect-filters=_airplay._tcp.local,_raop._tcp.local

[rlimits]
#rlimit-as=
#rlimit-core=0
#rlimit-data=8388608
#rlimit-fsize=0
#rlimit-nofile=768
#rlimit-stack=8388608
#rlimit-nproc=3
[router] / #

```

evverx:
@WeiLiCalix I think this issue wasn't noticed because usually avahi is run with interfaces where multicast is enabled and mdns packets lead to cleaning up wide-area lookups sooner or later. Other than that I assume various commands calling the EntryGroupNew/Free/Reset dbus methods are used and that leads to cleaning up those lookups as well. When only `avahi-browse -ar -d ...` is run there doesn't seem to be anything that can actually schedule `schedule_cleanup`. I'm inclined to say that `avahi_cleanup_dead_entries` should probably be called by `socket_event`.


Just to clarify regarding turning off `mdns` `(enable|disable)_interface` doesn't turn off `mdns`. `avahi` keeps receiving packets from them but it just rejects them. When it's run with `--debug` it prints something like
```
Received packet from invalid interface.
```
and `avahi_cleanup_dead_entries` is still called by `mcast_socket_event`.

WeiLiCalix:
evverx, you summarized what I mean! 

In most of scenario, there are always some devices (with mDNS feature) in the same network as avahi-daemon, so the `avahi_cleanup_dead_entries` will always be called when avahi-daemon processes a mDNS packet. mDNS's processing unintentionally freed the wide area DNS query.

I am glad you agree that avahi_cleanup_dead_entries should be called in socket_event. 

Because 
1, there is not an unconditional and periodical cleanup routine in application scope.
2, the counterpart functions such as `mcast_socket_event` and `legacy_unicast_socket_event` already explicitly called the avahi_cleanup_dead_entries.


And I have tested the fix and the result is good. The memory footprint kept the same after sending 3000 `avahi-browse-domains --domain=dns-sd.org -arpt` commands.

WeiLiCalix:
@pemensik 
I have merged the #411 and do the command 
```
kill -SIGUSR1 $(pidof avahi-daemon)
```
after sending about 200 ` avahi-browse-domains --domain=dns-sd.org -arpt` commands.

The cache about 60 items and the dead lookups are about 9000 items. 

Here is some of the log:
```
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: Got SIGUSR1, dumping record data.
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; ZONE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: 1.0.0.127.in-addr.arpa        IN      PTR router.local ; ttl=120 ; iface=1 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      A 127.0.0.1 ; ttl=120 ; iface=1 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: 1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa      IN      PTR router.local ; ttl=120 ; iface=1 proto=-1
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      AAAA ::1 ; ttl=120 ; iface=1 proto=-1
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: 217.91.245.10.in-addr.arpa    IN      PTR router.local ; ttl=120 ; iface=8 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      A 10.245.91.217 ; ttl=120 ; iface=8 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: c.b.6.d.0.4.e.f.f.f.f.8.6.7.2.d.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa      IN      PTR router.local ; ttl=120 ; iface=8 proto=-1
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      AAAA fe80::d276:8fff:fe40:d6bc ; ttl=120 ; iface=8 proto=-1
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: 1.1.168.192.in-addr.arpa      IN      PTR router.local ; ttl=120 ; iface=20 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      A 192.168.1.1 ; ttl=120 ; iface=20 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: 1.2.254.169.in-addr.arpa      IN      PTR router.local ; ttl=120 ; iface=20 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: router.local  IN      A 169.254.2.1 ; ttl=120 ; iface=20 proto=0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;; dead entries: 0
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; INTERFACE br-lan.IPv4 ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; CACHE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; INTERFACE eth5.IPv6 ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; CACHE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; INTERFACE eth5.IPv4 ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; CACHE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; INTERFACE lo.IPv6 ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; CACHE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; INTERFACE lo.IPv4 ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;;; CACHE DUMP FOLLOWS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;; WIDE AREA CACHE ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.google.com        IN      A 142.251.46.196 ; ttl=300
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: a.akamaiedge.net      IN      SOA \# 49 03 6E 30 61 C0 14 0A 68 6F 73 74 6D 61 73 74 65 72 06 61 6B ; ttl=834
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.ebay.com  IN      CNAME slot9428.ebay.com.edgekey.net ; ttl=251
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: multicastdns.org      IN      A 96.76.212.165 ; ttl=2802
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ftp.microsoft.akadns.net      IN      A 134.170.188.232 ; ttl=208
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: multicastdns.org      IN      SOA \# 45 03 6E 73 31 04 62 6F 6C 6F 03 6E 65 74 00 08 63 68 65 73 68 ; ttl=224
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: zeroconf.org  IN      SOA \# 45 03 6E 73 31 04 62 6F 6C 6F 03 6E 65 74 00 08 63 68 65 73 68 ; ttl=223
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: dns-sd.org    IN      SOA \# 45 03 6E 73 31 04 62 6F 6C 6F 03 6E 65 74 00 08 63 68 65 73 68 ; ttl=224
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.cnn.com   IN      CNAME cnn-tls.map.fastly.net ; ttl=256
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: slashdot.org  IN      A 104.18.28.86 ; ttl=238
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: slashdot.org  IN      A 104.18.29.86 ; ttl=238
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: zeroconf.org  IN      A 96.76.212.165 ; ttl=17046
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: dns-sd.org    IN      SOA \# 45 03 6E 73 31 04 62 6F 6C 6F 03 6E 65 74 00 08 63 68 65 73 68 ; ttl=300
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: wired.com     IN      SOA \# 66 06 6E 73 2D 32 37 36 09 61 77 73 64 6E 73 2D 33 34 C0 12 11 ; ttl=823
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.dns-sd.org        IN      A 96.76.212.165 ; ttl=16953
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: dns-sd.org    IN      A 96.76.212.165 ; ttl=16953
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: vanya.dns-sd.org      IN      A 10.0.1.203 ; ttl=17000
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: rose.dns-sd.org       IN      A 64.142.82.154 ; ttl=17368
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: cube.dns-sd.org       IN      A 1.2.3.4 ; ttl=17000
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: apple.com     IN      SOA \# 57 11 75 73 6D 73 63 32 2D 65 78 74 78 66 72 2D 30 30 31 03 64 ; ttl=900
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www-amazon-com.customer.fastly.net    IN      AAAA 2606:2cc0::374 ; ttl=1167
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www-amazon-com.customer.fastly.net    IN      AAAA 2606:2cc0:1::374 ; ttl=1167
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www-amazon-com.customer.fastly.net    IN      AAAA 2606:2cc0:2::374 ; ttl=1167
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.apple.com IN      CNAME www.apple.com.edgekey.net ; ttl=1080
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ftp.microsoft.com     IN      CNAME ftp.microsoft.akadns.net ; ttl=3419
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: tools.ietf.org        IN      A 50.223.129.194 ; ttl=264
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: wired.com     IN      A 151.101.2.194 ; ttl=2297
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: wired.com     IN      A 151.101.66.194 ; ttl=2297
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: wired.com     IN      A 151.101.194.194 ; ttl=2297
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: wired.com     IN      A 151.101.130.194 ; ttl=2297
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: news.bbc.co.uk        IN      CNAME news.bbc.co.uk.pri.bbc.co.uk ; ttl=17157
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.amazon.com        IN      CNAME tp.47cf2c8c9-frontier.amazon.com ; ttl=1722
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.apple.com.edgekey.net.globalredir.akadns.net      IN      CNAME e6858.e19.s.tl88.net ; ttl=1311
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: www.apple.com.edgekey.net     IN      CNAME www.apple.com.edgekey.net.globalredir.akadns.net ; ttl=6280
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: slot9428.ebay.com.edgekey.net IN      CNAME e9428.a.akamaiedge.net ; ttl=10166
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: tools.ietf.org        IN      AAAA 2001:559:c4c7::100 ; ttl=371
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: ;; WIDE AREA LOOKUPS ;;;
Fri Dec 23 02:03:58 2022 daemon.info avahi-daemon[20773]: rose.dns-sd.org       IN      AAAA ; n_send=1 (dead)
... 
remaining 8000+ lines ...
...


```

WeiLiCalix:
> The proposed code as in this PR (immediately freeing the packet) was not intended by the creators, I think. The reason is the following:
> 
> * instead of cleaning up and freeing immediately, two flags are set: l->dead and l->engine->cleanup_dead .
> * these flags make sure that later on lookup_destroy() is called.
> * in function lookup_destroy() in the same file, there is already the proper code for deallocating the packet.
> * in that same code also the other members of the AvahiWideAreaLookup struct are properly deallocated, in particular key and cname_key.
> * also the entire structure itself is freed there in the last line: avahi_free(l)
> * the lookup_destroy() code is called periodically on a timer
> * and also this code is called indirectly from browse.c in lookup_wide_area_callback() when the wide-area lookup is done. It calls lookup_unref(l) which eventually calls the cleanup function lookup_destroy()
> 
> So to me the current code looks okay and there's an intentional design to not clean up the packet immediately but let the lookup_destroy() code do that, later.
> 
> Similar design pattern is used for multicast lookup.
> 
> So I can't assume that there actually is a memory leak yet. In what situation could the memory leak occur i.e. when are the cleanup routines as I outlined above not called? We need to be first sure about this before making such a change.

@EskoDijk  
Yes, you are right. Some functions name ened by string "free" just do the dereference work, not really free the memory.
The memory should be freed in the destorying functions.

You mentioned that
> Similar design pattern is used for multicast lookup
Yes, it is. The `mcast_socket_event` explicitly call the `avahi_cleanup_dead_entries`, similarly, the `socket_event` should explicitly call the `avahi_cleanup_dead_entries` too. 
What is your opinion?

WeiLiCalix:
@pemensik 
Sorry, this issue is closed automatically by my wrongly force-push. I will create a new PR if all of you guys agree that `avahi_cleanup_dead_entries` should call `socket_event ` to clean up dead items.
thanks.

pemensik:
No, @WeiLiCalix please just push into a separate branch and modify this PR. We would lose previous discussion if we open always new PR. Do you have trouble using ``git mergetool`` after doing ``git fetch origin && git rebase origin/master`` to solve conflicts?

WeiLiCalix:
@pemensik  I failed to commit the code to this PR, so I have to commit it in the PR #434.

PR #413, opened 2022-12-02
CI: build avahi with gcc and clang, run the unit tests under ASan/UBsan and analyze the codebase using CodeQL

evverx:


evverx:
> Does Look like a great start.

Looking at https://github.com/lathiat/avahi/blob/master/.travis.yml (where avahi was just built with gcc and clang on Travis CI without running any unit tests) I think this PR already covers more than that with the static analysis and ASan/UBsan.

 > But I would like the actual memory leak fix last and merged in separate PR, where we could test the actual workflow does work.

It was already tested and it works because that memory leak was found by this action in the first place in https://github.com/evverx/avahi/actions/runs/3603292757/jobs/6071310293  and I needed to fix it to make the CI green. I can move it to a separate PR though.

pemensik:
Yes, it seems we are missing just enabled actions. Created [PR in my fork](https://github.com/pemensik/avahi/pull/1), and it shows actions right away. Even when it is yet to be added.

Please Trent @lathiat, would you be able to enable on [Actions settings](https://github.com/lathiat/avahi/settings/actions) page? Can you enable at least github vendored actions? I think also requiring approval for the first time contributors would be a good choice.

pemensik (approved):
Definitely improvement over current situation, where nothing is done on PR.

PR #416, opened 2022-12-02
Draft: Ci stuff

pemensik:
Just to check if it would be enabled.


pemensik:
oh, it is approved already, when it shares the commit with your PR.

PR #418, opened 2022-12-05
CI: turn on CIFuzz

evverx:
It should help to somewhat excercise the code paths consuming DNS packets under ASan, UBSan and MSan when PRs are opened: https://google.github.io/oss-fuzz/getting-started/continuous-integration/. Other than that it should also help to make sure that new PRs don't break OSS-Fuzz builds.

FWIW I'm planning to move the fuzzing infrastructure to the avahi repository to make it easier to extend the coverage but until then it still should be possible to run the fuzz targets on PRs.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,57 @@
+---
+# See: https://google.github.io/oss-fuzz/getting-started/continuous-integration/
+
+name: CIFuzz
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  Fuzzing:
+    permissions:
+      actions: read
+      contents: read
+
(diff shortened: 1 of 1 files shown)


evverx:
The CI failed because https://github.com/lathiat/avahi/pull/415 hasn't been merged yet. CIFuzz seems to be working fine.

pemensik (approved):
I am not sure if fuzzing build should be required for each PR. The change looks good, but I think we have not enough tests covering basic features.

evverx:
> I think we have not enough tests covering basic features

Agreed. I'm just trying to get as much coverage as I can with what I have. I think apart from fuzzing this PR is useful because it builds some part of avahi on i386 and until Packit is integrated (https://github.com/lathiat/avahi/issues/412#issuecomment-1338248629) CIFuzz is going to be the only place where 32-bit builds are somewhat compile tested.

PR #419, opened 2022-12-05
avahi-core: rearrange deallocations in avahi_time_event_queue_new a bit

evverx:
Reported by Coverity Scan.

I was testing a script sending avahi to Coverity Scan and it was the first warning it threw at me. I think combined with https://github.com/lathiat/avahi/pull/202 it should make Coverity a bit more happy.


pemensik (approved):
Definitely fixes use-after-free problem.

PR #420, opened 2022-12-05
tests: run more unit tests automatically

evverx:
To get `make check` to run tests automatically they should be added to the TESTS variable: https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html

Unlike functional tests like, say, avahi-test or client-test that should be run in suitable environments those tests are actually unit tests and can safely be run everywhere.

It should make the coverage go from 3% (https://coveralls.io/builds/54821508) to almost 10% (https://coveralls.io/builds/54825736). Admiteddly it isn't much but at least it should cover PRs/patches touching avahi-common and since the testing infrastructure is somewhat ready it should be possible to start adding/extending basic tests or use it to catch obvious regressions automatically.

The PR was tested on aarch64, i386, ppc64le, s390x and x86_64.

### DIFF
--- a/avahi-common/Makefile.am
+++ b/avahi-common/Makefile.am
@@ -48,6 +48,15 @@ noinst_PROGRAMS = \
 	watch-test \
 	watch-test-thread \
 	utf8-test
+
+TESTS = \
+	strlst-test \
+	domain-test \
+	alternative-test \
+	timeval-test \
+	watch-test \
+	watch-test-thread \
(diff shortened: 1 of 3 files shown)


pemensik (approved):
Great! I have not noticed we have already some unit tests, which are not yet added to the default target. I have usually checked avahi-test, but yes, we definitely want every working tests enabled by default.

PR #423, opened 2022-12-06
CI: run the unit tests under ASan/UBSan coming with clang

evverx:
~~I think avahi fails to build with clang and ASan/UBsan at the moment so it should be a draft until it works.~~

avahi still fails to build with clang and ASan/UBsan but I got it around with a kludge touching the CI only for now.


evverx:
It was moved to https://github.com/lathiat/avahi/pull/425. Clsoing.

PR #424, opened 2022-12-08
core: avoid populating NULL records in avahi_server_add_dns_server_address

evverx:
Reported by Coverity Scan.


evverx:
Looks like there are more issues with that part of the code. `AVAHI_PROTO_VALID` actually allows passing `AVAHI_PROTO_UNSPEC` apart from `AVAHI_PROTO_INET` and `AVAHI_PROTO_INET6` and in that case `r` is unitialized and depending on what it holds `avahi_server_add_dns_server_address` either returns or tries to pass uninitialized stuff around. I'd need to think a bit more how to fix it.

PR #425, opened 2022-12-08
CI: turn on as many fatal warnings as possible

evverx:
The idea is to turn on -Werror eventually to let compilers do their thing and catch various mistakes automatically. At this point it isn't possible to fully turn it on though so a bunch of warnings are just ignored. They should be fixed but until then the CI should at least be able to prevent new slip-ups from making it into the repository.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -4,6 +4,23 @@ set -eux
 set -o pipefail
 
 export ASAN_UBSAN=${ASAN_UBSAN:-false}
+export CFLAGS=${CFLAGS:-}
+export WERROR=${WERROR:-false}
+
+autogen_args=" --enable-compat-howl --enable-compat-libdns_sd --enable-tests"
+
+# CFLAGS can be used to turn off warnings globally but it doesn't make sense to
+# ignore, say, format-truncation everywhere when only one component triggers it.
+# This function patches Makefiles to disable warnings where it's actually necessary
+# to let the CI keep catching them in other places. It's a kludge and it should
+# be removed once all warnings avahi triggers are fixed.
+ignore_warnings() {
+    local _component="$1"
+    shift
+    for warning in "$@"; do
+        sed -i 's/^\(AM_CFLAGS=\)\(.*\)/\1-Wno-error='"$warning"' \2/' "$_component/Makefile.am"
+    done
+}
 
 case "$1" in
     install-build-deps)
@@ -12,15 +29,63 @@ case "$1" in
         apt-get build-dep -y avahi
         apt-get install -y libevent-dev qtbase5-dev
         apt-get install -y gcc clang
+        apt-get install -y autoconf-archive
         ;;
     build)
         if [[ "$ASAN_UBSAN" == true ]]; then
-            export CFLAGS="-fsanitize=address,undefined -g"
+            CFLAGS+=" -fsanitize=address,undefined -g"
             export ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
             export UBSAN_OPTIONS=print_stacktrace=1:print_summary=1:halt_on_error=1
         fi
 
-        ./bootstrap.sh --enable-tests --prefix=/usr
+        if [[ "$WERROR" == true ]]; then
+            # acx_pthread isn't compatible with CFLAGS=-Werror at all so it's just replaced
+            # with ax_pthread from autoconf-archive.
+            rm -rf common/acx_pthread.m4
+
+            CFLAGS+=" -Werror"
+
+            # The following warnings are ignored because it's hard to fix them all in one fell swoop.
+            # Ideally they should be fixed one by one and removed from here to let compilers catch them
+            # automatically on PRs.
+            if [[ "$CC" == gcc ]]; then
+                ignore_warnings avahi-autoipd unused-result
+                ignore_warnings avahi-common unused-result
+                ignore_warnings avahi-compat-howl cast-function-type missing-prototypes pedantic pointer-sign
+                ignore_warnings avahi-compat-howl/samples discarded-qualifiers pointer-sign shadow unused-parameter unused-variable
+                ignore_warnings avahi-compat-libdns_sd pedantic
+                ignore_warnings avahi-core format-truncation
+                ignore_warnings avahi-daemon stringop-truncation unused-result
+                ignore_warnings avahi-dnsconfd unused-result
+                ignore_warnings avahi-glib deprecated-declarations
+                ignore_warnings avahi-ui deprecated-declarations implicit-fallthrough maybe-uninitialized
+                ignore_warnings avahi-utils unused-result
+
+                # avahi-gobject triggers a lot of warnings like
+                # "Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE""
+                # and judging by https://gitlab.gnome.org/GNOME/glib/-/issues/2247
+                # they can't be disabled when code is built with gcc with -Werror.
+                autogen_args+=" --disable-gobject"
+            elif [[ "$CC" == clang ]]; then
+                CFLAGS+=" -Wno-error=strict-prototypes"
+
+                ignore_warnings avahi-autoipd cast-align
+                ignore_warnings avahi-client varargs
+                ignore_warnings avahi-compat-howl cast-align missing-prototypes
+                ignore_warnings avahi-compat-howl/samples cast-align incompatible-pointer-types-discards-qualifiers pointer-sign shadow unused-parameter unused-variable
+                ignore_warnings avahi-core cast-align logical-not-parentheses varargs
+                ignore_warnings avahi-daemon cast-align
+                ignore_warnings avahi-glib deprecated-declarations enum-conversion
+                ignore_warnings avahi-gobject deprecated enum-conversion
+                ignore_warnings avahi-ui deprecated-declarations
+            fi
+        fi
+
+        if ! ./autogen.sh $autogen_args --prefix=/usr; then
+            cat config.log
+            exit 1
+        fi
+
         make -j"$(nproc)" V=1
         make check VERBOSE=1
         ;;
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -4,6 +4,7 @@ set -eux
 set -o pipefail
 
 export ASAN_UBSAN=${ASAN_UBSAN:-false}
+export BUILD_ONLY=${BUILD_ONLY:-false}
 export CFLAGS=${CFLAGS:-}
 export WERROR=${WERROR:-false}
 
@@ -98,6 +99,11 @@ case "$1" in
         fi
 
         make -j"$(nproc)" V=1
+
+        if [[ "$BUILD_ONLY" == true ]]; then
+            exit 0
+        fi
+
         make check VERBOSE=1
         ;;
     *)
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -32,6 +32,11 @@ case "$1" in
         apt-get install -y autoconf-archive
         ;;
     build)
+        # acx_pthread isn't compatible with CFLAGS=-Werror at all. It isn't compatible with
+        # clang and -fsanitize=address,undefined either so it's just replaced here with ax_pthread
+        # from autoconf-archive.
+        rm -rf common/acx_pthread.m4
+
         if [[ "$ASAN_UBSAN" == true ]]; then
             CFLAGS+=" -fsanitize=address,undefined -g"
             export ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
@@ -39,10 +44,6 @@ case "$1" in
         fi
 
         if [[ "$WERROR" == true ]]; then
-            # acx_pthread isn't compatible with CFLAGS=-Werror at all so it's just replaced
-            # with ax_pthread from autoconf-archive.
-            rm -rf common/acx_pthread.m4
-
             CFLAGS+=" -Werror"
 
             # The following warnings are ignored because it's hard to fix them all in one fell swoop.
@@ -80,6 +81,16 @@ case "$1" in
                 ignore_warnings avahi-ui deprecated-declarations
             fi
         fi
+
+        # Some parts of avahi like avahi-qt and c-plus-plus-test are built with CXX so
+        # it should match CC. To avoid weird side-effects CXXFLAGS should fully match
+        # CFLAGS as well so CXXFLAGS is fully replaced with CFLAGS here.
+        if [[ "$CC" == gcc ]]; then
+            export CXX=g++
+        elif [[ "$CC" == clang ]]; then
+            export CXX=clang++
+        fi
+        export CXXFLAGS="$CFLAGS"
 
         if ! ./autogen.sh $autogen_args --prefix=/usr; then
             cat config.log
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,8 +23,10 @@ jobs:
       matrix:
         env:
           - { CC: "gcc" }
+          - { CC: "gcc", WERROR: "true" }
           - { CC: "gcc", ASAN_UBSAN: "true" }
           - { CC: "clang" }
+          - { CC: "clang", WERROR: "true" }
     env: ${{ matrix.env }}
     steps:
       - name: Repository checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,7 @@ jobs:
           - { CC: "clang", WERROR: "true", CFLAGS: "-O1" }
           - { CC: "clang", WERROR: "true", CFLAGS: "-O2" }
           - { CC: "clang", WERROR: "true", CFLAGS: "-O3" }
+          - { CC: "clang", ASAN_UBSAN: "true" }
     env: ${{ matrix.env }}
     steps:
       - name: Repository checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,10 +23,16 @@ jobs:
       matrix:
         env:
           - { CC: "gcc" }
-          - { CC: "gcc", WERROR: "true" }
+          - { CC: "gcc", WERROR: "true", CFLAGS: "-O0" }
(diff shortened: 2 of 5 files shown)


.github/workflows/build.sh
```diff
@@ -20,7 +24,37 @@ case "$1" in
             export UBSAN_OPTIONS=print_stacktrace=1:print_summary=1:halt_on_error=1
         fi
 
-        ./bootstrap.sh --enable-tests --prefix=/usr
+        if [[ "$WERROR" == true ]]; then
+            # acx_pthread isn't compatible with CFLAGS=-Werror at all so it's just replaced
+            # with ax_pthread from autoconf-archive.
+            rm -rf common/acx_pthread.m4
+
+            export CFLAGS="-Werror"
```
pemensik, line 32: I would use CFLAGS+= here too and below all changes just export CFLAGS. It would make possible configuring CFLAGS from yaml file.

.github/workflows/build.sh
```diff
@@ -20,7 +24,37 @@ case "$1" in
...
+            if [[ "$CC" == gcc ]]; then
+                CFLAGS+=" -Wno-error=format-truncation -Wno-error=deprecated-declarations -Wno-error=implicit-fallthrough"
+            elif [[ "$CC" == clang ]]; then
+                CFLAGS+=" -Wno-error=strict-prototypes -Wno-error=invalid-source-encoding -Wno-error=varargs -Wno-error=cast-align"
+                CFLAGS+=" -Wno-error=logical-not-parentheses -Wno-error=deprecated-declarations"
+            fi
+
+            # avahi-gobject triggers a lot of various warnings so it's just easier to turn
+            # it off entirely instead of disabling all the warnings with -Wno-error.
+            additional_autogen_args+=" --disable-gobject"
```
pemensik, line 46: Would be nice to fill at least issue that gobject produces a lot of warnings and including issue reference in comment.

.github/workflows/build.sh
```diff
@@ -20,7 +24,37 @@ case "$1" in
...
+
+            # For some reason "howl" and "libdns_sd" completely ignore -Wno-error= passed via
+            # CFLAGS so they are ignored until they are fixed somehow.
+            additional_autogen_args+=" --disable-compat-howl --disable-compat-libdns_sd"
```
pemensik, line 50: Interesting. That should be fixed, as Fedora requires CFLAGS passing from rpm to builds. Haven't noticed this is not passed everywhere. May I ask separate issue for it too?

.github/workflows/build.sh
```diff
@@ -4,6 +4,9 @@ set -eux
 set -o pipefail
 
 export ASAN_UBSAN=${ASAN_UBSAN:-false}
+export WERROR=${WERROR:-false}
+
+additional_autogen_args="--enable-compat-howl --enable-compat-libdns_sd"
```
pemensik, line 9: I would leave those expandable by environment if possible. We have something here and something below at autogen.sh. Would it make more sense to have configure options only on single place?

pemensik (changes requested):

.github/workflows/build.sh
evverx: I think ideally all those flags should probably be passed via environment variables. The problem is that at this point due to various build system bugs it's kind of hard to run the script directly. For example, I can test `avahi` locally with `clang` and `-Werror` by running just
```sh
CC=clang WERROR=true ./.github/workflows/build.sh build
```
If I dropped the WERROR variable and relied on `CFLAGS` and so on I would have to copy-paste CFLAGS from the yml file and the command line would look like
```
CC=clang \
CFLAGS="-Werror -Wno-error=strict-prototypes -Wno-error=invalid-source-encoding -Wno-error=varargs -Wno-error=cast-align -Wno-error=logical-not-parentheses -Wno-error=deprecated-declarations" \
ADDITIONAL_AUTOGEN_ARGS="--disable-gobject --disable-compat-howl --disable-compat-libdns_sd" \
.github/workflows/build.sh build
```
The script would still have to look for `-Werror` in `CFLAGS` to be able to remove `common/acx_pthread.m4`.

In other words variables like `WERROR` and `ASAN_UBSAN` are used to get around a bunch of issues and once those issues are fixed I think it would probably make sense to replace them with just `CFLAGS=-Werror` and `CFLAGS=-fsanitize=address,undefined`.

.github/workflows/build.sh
evverx: I think I got it to work a few hours ago. Let me double-check.

.github/workflows/build.sh
evverx: I'll replace `CFLAGS=` with `CFLAGS+=` but I would keep passing `WERROR` and `UBSAN_UBSAN` instead of `CFLAGS`: https://github.com/lathiat/avahi/pull/425#discussion_r1044925323 should be applicable here too.

evverx: I replaced `CFLAGS=` with `CFLAGS+=` and started using it to pass various optimization levels to catch even more warnings.

.github/workflows/build.sh
evverx: I brought those compat libraries back. I double-checked and it seems CFLAGS aren't ignored there. I'm not sure why they failed to build though. My guess would be that I missed a couple of flags to really silence all the warnings there.

.github/workflows/build.sh
evverx: I started building `avahi-gobject` with `clang` and adjusted the comment accordingly so it's ignored with `gcc` only because all those "deprecated" warnings can't be disabled for some reason with gcc: https://gitlab.gnome.org/GNOME/glib/-/issues/2247. I'm not sure if it makes sense to open an issue covering `avahi-gobject` specifically because all those warnings are visible when `avahi` is built and the script kind of already documents that `avahi` fails to compile with `-Werror` in general and goes even further by listing specific warnings every component triggers with both gcc and clang. For the record `gobject` fails to compile with `gcc` with `-Werror`:
<details>
  <summary>error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE"</summary>

```
ga-client.c: In function 'ga_client_init':
ga-client.c:67:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
   67 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function 'ga_client_set_property':
ga-client.c:81:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
   81 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(client);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function 'ga_client_get_property':
ga-client.c:98:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
   98 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(client);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function 'ga_client_class_init':
ga-client.c:116:5: error: 'g_type_class_add_private' is deprecated [-Werror=deprecated-declarations]
  116 |     g_type_class_add_private(ga_client_class, sizeof (GaClientPrivate));
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/gobject/gobject.h:24,
                 from /usr/include/glib-2.0/gobject/gbinding.h:29,
                 from /usr/include/glib-2.0/glib-object.h:22,
                 from ga-client.h:23,
                 from ga-client.c:27:
/usr/include/glib-2.0/gobject/gtype.h:1346:10: note: declared here
 1346 | void     g_type_class_add_private       (gpointer                    g_class,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function 'ga_client_dispose':
ga-client.c:157:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  157 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function '_avahi_client_cb':
ga-entry-group.c: In function 'ga_entry_group_init':
ga-entry-group.c:82:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
   82 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(obj);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c:217:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  217 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_get_property':
ga-entry-group.c:99:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
   99 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c:226:17: error: implicit conversion from 'AvahiClientState' to 'GaClientState' [-Werror=enum-conversion]
  226 |     priv->state = state;
      |                 ^
ga-entry-group.c: In function 'ga_entry_group_class_init':
ga-entry-group.c:115:5: error: 'g_type_class_add_private' is deprecated [-Werror=deprecated-declarations]
  115 |     g_type_class_add_private(ga_entry_group_class,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/gobject/gobject.h:24,
                 from /usr/include/glib-2.0/gobject/gbinding.h:29,
                 from /usr/include/glib-2.0/glib-object.h:22,
                 from ga-error.h:23,
                 from ga-entry-group.c:30:
/usr/include/glib-2.0/gobject/gtype.h:1346:10: note: declared here
 1346 | void     g_type_class_add_private       (gpointer                    g_class,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_dispose':
ga-entry-group.c:143:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  143 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c: In function 'ga_client_start_in_context':
ga-client.c:236:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  236 |     GaClientPrivate *priv = GA_CLIENT_GET_PRIVATE(client);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_finalize':
ga-entry-group.c:166:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  166 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c:248:36: error: implicit conversion from 'GaClientFlags' to 'AvahiClientFlags' [-Werror=enum-conversion]
  248 |                                priv->flags,
      |                                ~~~~^~~~~~~
ga-entry-group.c: In function '_avahi_entry_group_cb':
ga-entry-group.c:215:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  215 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(self);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c:222:17: error: implicit conversion from 'AvahiEntryGroupState' to 'GaEntryGroupState' [-Werror=enum-conversion]
  222 |     priv->state = state;
      |                 ^
ga-client.c: In function 'ga_client_get_property':
ga-client.c:105:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
  105 |             g_value_set_enum(value, priv->flags);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c:106:9: note: here
  106 |         default:
      |         ^~~~~~~
ga-entry-group.c: In function 'ga_entry_group_add_service_full_strlist':
ga-entry-group.c:378:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  378 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_add_record_full':
ga-entry-group.c:487:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  487 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_service_thaw':
ga-entry-group.c:517:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  517 |             (GA_ENTRY_GROUP_GET_PRIVATE(priv->group)->group,
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_attach':
ga-entry-group.c:575:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  575 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_commit':
ga-entry-group.c:599:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  599 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-entry-group.c: In function 'ga_entry_group_reset':
ga-entry-group.c:614:13: error: Deprecated pre-processor symbol: replace with "G_ADD_PRIVATE" [-Werror]
  614 |     GaEntryGroupPrivate *priv = GA_ENTRY_GROUP_GET_PRIVATE(group);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[3]: *** [Makefile:725: libavahi_gobject_la-ga-client.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
cc1: all warnings being treated as errors
make[3]: *** [Makefile:732: libavahi_gobject_la-ga-entry-group.lo] Error 1
make[3]: Leaving directory '/home/vagrant/avahi/avahi-gobject'
make[2]: *** [Makefile:608: all] Error 2
make[2]: Leaving directory '/home/vagrant/avahi/avahi-gobject'
make[1]: *** [Makefile:831: all-recursive] Error 1
make[1]: Leaving directory '/home/vagrant/avahi'
make: *** [Makefile:741: all] Error 2
```

.github/workflows/build.sh
evverx: Trying to bring back the compat libraries I noticed that they trigger a lot of warnings and decided that it doesn't make sense to silence them everywhere so instead of using `CFLAGS` the script patches Makefiles on a case-by-case basis. I started using CFLAGS where I can though and at this point it's helpful when it comes to passing various optimization levels.

evverx: I've just pushed two more commits where I fixed the "invalid-source-encoding" warnings and brought clang/ASan/UBsan.

evverx: I've just push one more commit where I dropped the CodeQL autobuild script to fix the extraction of the Python code and the C# code.

pemensik:
I think I would like something bind9 project has. They enable various default profiles when ``./configure --enable-developer`` is used. It switches warnings to errors and such settings. I think it would be nice to have simple way to enable such mode and not having all those settings in special build script. I think make target would be better, with just some step before.

I think also developers might want to do some checks and we should simplify that as much as possible.

evverx:
> I think it would be nice to have simple way to enable such mode and not having all those settings in special build script

I agree with that and I think that building `avahi` with `-Werror` should be as easy as just passing `CFLAGS=-Werror` but for that to work all the warnings should be fixed first. The build system itself needs fixing as well because right now `avahi` plainly fails to compile with `-fsanitize=address,undefined` and `-Werror` due to that out-of-date thread macro

> I think I would like something bind9 project has. They enable various default profiles when ./configure --enable-developer is used. It switches warnings to errors and such settings

I think `bind9` has been maintained much more actively than `avahi` so it doesn't need kludges making it buildable with the basic compiler flags. When `avahi` reaches that level it should be possible to add something like `--enable-errors` to just put `-Werror` in `CFLAGS` or something like that. Until then 12 Makefiles have to be patched to silence all the known warnings and at least prevent new warnings from making it into the repository. The idea behind this PR is not to make the build system perfect (it isn't possible at this point and it can take months to actually fix all the known issues) but to turn on some basic checks to catch new issues and also let the CI verify patches fixing the known warnings by dropping the corresponding kludges one-by-one (like I already did with `invalid-source-encoding`).

evverx:
Since avahi is no longer actively maintained I'm going to put this on hold.

PR #429, opened 2022-12-14
Increase max hostname specified in simple-protocol

pemensik:
MDNS protocol specifies maximal hostname length used is 255 characters plus terminating zero. One label has maximal 63 characters, but full hostname can be longer. Ensure we have space for any hostname allowed by the protocol.

### DIFF
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -36,6 +36,7 @@
 #include <systemd/sd-daemon.h>
 #endif
 
+#include <avahi-common/domain.h>
 #include <avahi-common/llist.h>
 #include <avahi-common/malloc.h>
 #include <avahi-common/error.h>
@@ -267,7 +268,7 @@ static void dns_server_browser_callback(
 }
 
 static void handle_line(Client *c, const char *s) {
-    char cmd[64], arg[64];
+    char cmd[64], arg[AVAHI_DOMAIN_NAME_MAX];
     int n_args;
 
     assert(c);
@@ -276,7 +277,7 @@ static void handle_line(Client *c, const char *s) {
     if (c->state != CLIENT_IDLE)
         return;
 
-    if ((n_args = sscanf(s, "%63s %63s", cmd, arg)) < 1 ) {
+    if ((n_args = sscanf(s, "%63s %1013s", cmd, arg)) < 1 ) {
         client_output_printf(c, "%+i Failed to parse command, try \"HELP\".\n", AVAHI_ERR_INVALID_OPERATION);
         c->state = CLIENT_DEAD;
         return;


evverx:
Looking at 4499f8fb35eddd and d1dd07145bb3db5 I think `AVAHI_DOMAIN_NAME_MAX` should be used here instead. But I wouldn't bump it just yet.

evverx:
I'm not sure where 256 came from but it isn't enough for escaped labels. There is another variation of this: https://github.com/lathiat/avahi/issues/495.

Those hostnames ends up in `avahi_s_host_name_resolver_prepare` and it accepts valid fqdns (according to `avahi_is_valid_fqdn`) and they are generally larger than 256 bytes. Other than that it should probably be consistent with the `ResolveHostName` D-Bus method where hostnames can be up to `AVAHI_DOMAIN_NAME_MAX`.

pemensik:
I think 256 is length in wire format. Looking at computation at AVAHI_DOMAIN_NAME_MAX, it seems to me completely wrong. It may work for .local multicast domain, but certainly would not work for wide area DNS-based records like ``service with spaces and maybe (various) interpuction._https._tcp.somelongdomainname.evenlongersubdomain.example.org``. But I think it is extremely wrong to not have separate types with appropriate maximal lengthts for wire format and presentation escaped text. It is confusing and obviously not only for me.

pemensik:
But the question is in which format are name supplied from nss-mdns and how it threats special characters like dots or other interpuction. I would say it should be name in current locale, but I doubt that were ever tested. Especially in cases stub sender and avahi run under different locale, maybe even different encoding.

evverx:
> But I think it is extremely wrong to not have separate types with appropriate maximal lengthts for wire format and presentation escaped text

Agreed. It is what it is unfortunately. The problem is that all the labels are normalized once they are received so internal buffers should be able to fit escaped labels.

> the question is in which format are name supplied from nss-mdns and how it threats special characters like dots or other interpuction

Regardless of what nss-mdns supplies this API in its current form can accept all sorts of labels as long as avahi can normalize them. `[` and `\091` are the same there. It kind of makes sense given that `avahi-browse` shows escaped labels and it's probably not uncommon to pass those strings back to avahi via D-Bus or the simple protocol.

avahi-daemon/simple-protocol.c
```diff
@@ -264,7 +264,7 @@ static void dns_server_browser_callback(
 }
 
 static void handle_line(Client *c, const char *s) {
-    char cmd[64], arg[64];
+    char cmd[64], arg[256];
```
evverx, line 267: I'm not sure what to do here. I still think it should be `AVAHI_DOMAIN_NAME_MAX` to make it consistent with the other API accepting domain names that aren't in the wire format. I guess it should be possible to bump it to 256 to address the most popular issue where long hostnames + '.local' can't fit but I'm not sure why the wire format limit should be used here.

Could you rebase it to trigger the CI? The simple protocol is tested by the smoke test with radamsa generating all sorts of commands with all sorts of arguments.

avahi-daemon/simple-protocol.c
```diff
@@ -276,7 +277,7 @@ static void handle_line(Client *c, const char *s) {
     if (c->state != CLIENT_IDLE)
         return;
 
-    if ((n_args = sscanf(s, "%63s %63s", cmd, arg)) < 1 ) {
+    if ((n_args = sscanf(s, "%63s %1013s", cmd, arg)) < 1 ) {
```
pemensik, line 280: I don't like not using constant here, but not sure how to get length here. Without using completely different calls.

evverx: I don't like it either but given that it's unlikely to ever change I'd just use
```c
char cmd[64], arg[1014];
```
instead to keep it simple and somewhat consistent with the hard-coded `cmd[64]`. Could you also add something like
```diff
diff --git a/.github/workflows/smoke-tests.sh b/.github/workflows/smoke-tests.sh
index b550a04..dcfc499 100755
--- a/.github/workflows/smoke-tests.sh
+++ b/.github/workflows/smoke-tests.sh
@@ -58,6 +58,7 @@ avahi-resolve -v -a "$ipv6addr"
 cmds=(
     "HELP"
     "RESOLVE-HOSTNAME $h"
+    "RESOLVE-HOSTNAME $(perl -le '(print q/A/ x 1014)')"
     "RESOLVE-HOSTNAME-IPV6 $h"
     "RESOLVE-HOSTNAME-IPV4 $h"
     "RESOLVE-ADDRESS $ipv4addr"
```
to the smoke test? Even though `radamsa` generates all sorts of weird things I think it would be better to test this hard-coded assumption explicitly too. Once the CI with that test passes feel free to merge the PR.

evverx: Let's keep AVAHI_DOMAIN_NAME_MAX here. Looks like this pattern is used in other places too. Ideally it should be replaced but it can be done later.

I'll add the test in a separate PR.

PR #431, opened 2022-12-22
Include wide-area lookups in cache dump

pemensik:
Include counts of dead entries in mdns resolution. Include lookups active and dead for wide-area. Attempt to debug easier PR#411 issue.


pemensik:
This change should make easier to debug issue in PR #411. It logs also number of dead entries and wide-area lookups in progress or dead.

pemensik:
This dump is created by ``sudo kill -SIGUSR1 $(pidof avahi-daemon)`` command.

evverx:
I wonder if it should be closed? That issue was resolved so it's no longer necessary to dump that data.

I guess in general it would be nice to make avahi more introspectable but that interface should probably be separate from the cache.

evverx:
Let's close this.

Issue #414, opened 2022-12-02
DNS wide-area requests allocate unnecessary large packets


pemensik:
During reviews on issue #403  and #411, I have found function ``avahi_dns_packet_new(0)`` allocates quite large packet buffer. The packet size of AVAHI_DNS_PACKET_SIZE_MAX is allocated, which is small bit bigger than AVAHI_DNS_RDATA_MAX (0xFFFF). That is completely unnecessary, because the allocated packet is kept for the same wide area lookup request.

- Minimal change would be to reallocate the packet size after the query is created. It would be safe to reallocate the packet just to hold ``p->size``, and make ``p->size == p->max_size``. Current code keeps allocated unnecessary large buffers until whole wide area browse request is destroyed.

- A better change would be using just ``avahi_dns_packet_extend()`` to increase whole packet size if the current buffer were not big enough. I think that would require using ``p->data`` for packet data, which is still handled by ``AVAHI_DNS_PACKET_DATA(p)`` macro. Because the basic pointer to p should not change, but calling *avahi_realloc()* on p->data should be safe enough. It should start with size suitable for quering common short names. Just start with small packet data and be able to increase the packet size if added sections/names are too long to fit.

pemensik:
The main problem is it might make sense to have a bit bigger packet than neccessary when building the packet and more names added to packet are pending. But as soon as the packet is final and will no longer be modified, which I think is true after leaving ``avahi_wide_area_lookup_new()`` function, there is no need to have max packet size bigger than actual used size for the packet. I think adding ``avahi_dns_packet_finalize()`` function to shrink to the minimal required size would be useful.

PR #434, opened 2023-01-10
fix memory leak in wide area lookup

WeiLiCalix:


WeiLiCalix:
commit for https://github.com/lathiat/avahi/pull/411

pemensik:
This is just similar to ``mcast_socket_event`` from avahi-core/server.c. After discussion in PR #411, I think this should be ok. Not 100% sure. Related also to issue #414, but that is not yet fixed. Should help at least.

PR #456, opened 2023-04-27
Prevent too long service to trigger assertion failure

### LINKED ISSUE #455: Reachable assertion in avahi_dns_packet_append_record
It can be triggered by unprivileged local users (unless `disable-user-service-publishing` is set to `yes` explicitly):
```sh
avahi-publish -s T _qotd._tcp 22 $(perl -le 'print "A " x 100000')
```
```sh
dns.c:806: avahi_dns_packet_append_record: Assertion `size <= AVAHI_DNS_RDATA_MAX' failed.
```
```sh
#0  0x00007f848f08ec0c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f848f03e986 in raise () from /lib64/libc.so.6
#2  0x00007f848f0287f4 in abort () from /lib64/libc.so.6
#3  0x00007f848f02871b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007f848f037536 in __assert_fail () from /lib64/libc.so.6
#5  0x00007f848fdd6524 in avahi_dns_packet_append_record (p=0x7f848e44b800, r=0x604000007ad0, cache_flush=0, max_ttl=0) at dns.c:806
#6  0x00007f848fda0055 in elapse_callback (e=0x60600002f6c0, data=0x60600002f660) at probe-sched.c:265
#7  0x00007f848fd3d5d7 in expiration_event (timeout=0x606000000aa0, userdata=0x6030000008e0) at timeeventq.c:94
#8  0x00007f849080a115 in timeout_callback (t=0x606000000aa0) at simple-watch.c:447
#9  0x00007f849080b6d2 in avahi_simple_poll_dispatch (s=0x60e000000040) at simple-watch.c:570
#10 0x00007f849080c03f in avahi_simple_poll_iterate (s=0x60e000000040, timeout=-1) at simple-watch.c:605
#11 0x0000000000413103 in run_server (c=0x51d4e0 <config>) at main.c:1268
#12 0x0000000000415069 in main (argc=5, argv=0x7fffeaf698a8) at main.c:1686
```

pemensik:
Handle failure if requested name is too long.

Fixes #455

### DIFF
--- a/avahi-core/dns.c
+++ b/avahi-core/dns.c
@@ -782,7 +782,7 @@ static int append_rdata(AvahiDnsPacket *p, AvahiRecord *r) {
 
 
 uint8_t* avahi_dns_packet_append_record(AvahiDnsPacket *p, AvahiRecord *r, int cache_flush, unsigned max_ttl) {
-    uint8_t *t, *l, *start;
+    uint8_t *t, *l, *start, *end;
     size_t size;
 
     assert(p);
@@ -802,8 +802,13 @@ uint8_t* avahi_dns_packet_append_record(AvahiDnsPacket *p, AvahiRecord *r, int c
     if (append_rdata(p, r) < 0)
         goto fail;
 
-    size = avahi_dns_packet_extend(p, 0) - start;
-    assert(size <= AVAHI_DNS_RDATA_MAX);
+    end = avahi_dns_packet_extend(p, 0);
+    if (!end)
+	goto fail;
+
+    size = end - start;
+    if (size > AVAHI_DNS_RDATA_MAX)
+        goto fail;
 
 /*     avahi_log_debug("appended %u", size); */
 


avahi-core/dns.c
```diff
@@ -802,8 +802,13 @@ uint8_t* avahi_dns_packet_append_record(AvahiDnsPacket *p, AvahiRecord *r, int c
...
         goto fail;
 
-    size = avahi_dns_packet_extend(p, 0) - start;
-    assert(size <= AVAHI_DNS_RDATA_MAX);
+    end = avahi_dns_packet_extend(p, 0);
+    if (!end)
+	goto fail;
+
+    size = end - start;
+    if (size > AVAHI_DNS_RDATA_MAX)
```
evverx, line 810: I think I'd expect `append_rdata` to return an error and that's probably what that assertion expected too. I'll take try to take a closer look. Either way to judge from https://coveralls.io/builds/62671171/source?filename=avahi-core%2Fdns.c#L784 this function isn't covered by any tests so maybe it would be great to add a test case.

evverx:
Before I forget I applied this patch and even though the crash is gone I don't think it works as expected in the sense that instead of failing avahi publishes the service without any TXT resource records and produces a lot of "Record too large, cannot send" messages along the way and every time they are requested.

evverx:
Turns out `systemd-resolved` isn't much better. It accepts services like that via D-Bus and then fails to send anything. Those services can be queried on machines where they are published but all the other hosts are in a limbo because resolved fails to reply due to large packages. Corner cases are hard.

Either way I think ideally services like that should be rejected as early as possible to avoid half-published services in the first place.

pemensik:
Yes, these my changes are just very basic. There should be methods computing size of response before the allocation of response body, but I have found none of that. I would like to advance unicast dns support to nicer interface and behaviour, but I guess that should be done after release of bugfix 0.9 release. Yes, we should have input checking, which would refuse any invalid names close to the input, not deeply inside processing. It should respond at least with FORMERR, REFUSED or SERVFAIL if the request were wrong.

As it is important part for unicast support, which we have only supported in RHEL, I agree this would need a lot improved testing for few possible responses and common failure handling. But I think no small refactoring would be required.

For multicast responses we know MTU of given interface. I think it should make test response during service registration. If the test response packet fails, then refuse even registering that service and return error. I am not yet sure how difficult that would be. Realizing there is something wrong with the service only when trying to answer is way too late, if the cause is something we could know in advance.

evverx:
Agreed. The right fix would be much more involved.

Let's put this on hold then. The crash can be mitigated by setting `disable-user-service-publishing` to `yes`. Other than that it's probably better to crash than to allow stuffing huge blobs into avahi until it can't cope.

evverx:
The discussion moved to https://github.com/lathiat/avahi/pull/488. I think this PR can be closed.

Issue #296, opened 2020-07-25
avahi-daemon/sd-daemon.c: outdated (non-working) systemd check 


ovitters:
avahi-daemon/sd-daemon.c seems to be copied ages ago from systemd. It checks if systemd is running by looking at the following paths:

- /sys/fs/cgroup
- /sys/fs/cgroup/systemd

See https://github.com/lathiat/avahi/blob/master/avahi-daemon/sd-daemon.c. The (buggy) code reads:

```
int sd_booted(void) {
#if defined(DISABLE_SYSTEMD) || !defined(__linux__)
        return 0;
#else

        struct stat a, b;

        /* We simply test whether the systemd cgroup hierarchy is
         * mounted */

        if (lstat("/sys/fs/cgroup", &a) < 0)
                return 0;

        if (lstat("/sys/fs/cgroup/systemd", &b) < 0)
                return 0;

        return a.st_dev != b.st_dev;
#endif
}
```

However, the second path does not exist on any recent systemd release. This due to the usage of cgroups2 without any legacy backup code. The sd_booted from (recent?) systemd versions does a different check, see https://www.freedesktop.org/software/systemd/man/sd_booted.html:
> Internally, this function checks whether the directory /run/systemd/system/ exists. A simple check like this can also be implemented trivially in shell or any other language.

It seems currently the sd_booted function can be found at:
https://github.com/systemd/systemd/blob/master/src/libsystemd/sd-daemon/sd-daemon.c#L622

Quoting that code:
```
_public_ int sd_booted(void) {
        /* We test whether the runtime unit file directory has been
         * created. This takes place in mount-setup.c, so is
         * guaranteed to happen very early during boot. */

        if (laccess("/run/systemd/system/", F_OK) >= 0)
                return true;

        if (errno == ENOENT)
                return false;

        return -errno;

````

Could you please adjust the sd_booted code according to the systemd code? It might be good to link against libsystemd. Suggest to perhaps also check if any other copied systemd function is outdated.

evverx:
I'll revert that PR shortly.

Issue #287, opened 2020-05-16
Don't use embedded code copy of sd-daemon.[ch]


mbiebl:
In the early days of libsystemd/systemd, embedding sd-daemon.[ch] was an approach recommended by upstream. This is no longer the case. Since libsystemd has matured, nowadays it's better to link against libsystemd instead.

pemensik:
Just hit it during SPDX license evaluation. Yes, these should instead use libsystemd. Not bundling the code itself.
./avahi-daemon/sd-daemon.c      Expat
./avahi-daemon/sd-daemon.h      Expat

evverx:
I'll revert that PR shortly.

PR #459, opened 2023-05-26
Stop using embedded copy of sd-daemon.{c,h} and use libsystemd instead

mbiebl:
Fixes: #287
Fixes: #296


mbiebl:
See also
https://salsa.debian.org/utopia-team/avahi/-/merge_requests/11
https://bugzilla.redhat.com/1907727

mbiebl:
The default for `enable-systemd` is yes. So this needs libsystemd-dev as build dependency now for the CI.

andhe:
Possibly not a *good* idea, but an idea: there would be a lot less ifdef HAVE_LIBSYSTEMD if sd_notifyf was just defined as an empty macro when we do not HAVE_LIBSYSTEMD.

evverx:
> So this needs libsystemd-dev as build dependency now for the CI

It can be added to https://github.com/lathiat/avahi/blob/6395c45bcd1210cbbb23fc92294fd61e968daaf2/.github/workflows/build.sh#L13

evverx:
The fuzz targets should be fixed on OSS-Fuzz though. Ideally they should be kept upstream and https://github.com/google/oss-fuzz/pull/9107 was supposed to help with that but I decided not to proceed with that because avahi isn't actively maintained any more. (That PR can be freely used of course should anyone decide to resurrect avahi).

evverx:
I've just opened https://github.com/lathiat/avahi/pull/460. If it's merged it should be possible to disable libsystemd by editing the "fuzz/oss-fuzz.sh" script here. Without that it should be turned off on OSS-Fuzz by opening a PR and signing their CLA.

The other CI workflows use https://github.com/lathiat/avahi/blob/6395c45bcd1210cbbb23fc92294fd61e968daaf2/.github/workflows/build.sh#L13

lathiat:
We'll merge #460 shortly, which will bring the OSS Fuzz scripts in repo. Then let's update this PR to include the changes to add libsystemd-dev to:
- avahi/.github/workflows/build.sh
- fuzz/oss-fuzz.sh

Then happy to merge this.

evverx:
Once https://github.com/google/oss-fuzz/pull/10412 is merged CIFuzz should start using the "fuzz/oss-fuzz.sh" script. libsystemd can be turned off by adding one more disable option to https://github.com/lathiat/avahi/blob/ff2caff0004f5d2219fb7dd98f86957d54f6eb1c/fuzz/oss-fuzz.sh#L24-L28

configure.ac
```diff
@@ -823,6 +823,25 @@ if test "x$HAVE_LIBDAEMON" = "xyes" ; then
 fi
 AM_CONDITIONAL(HAVE_LIBDAEMON, test "x$HAVE_LIBDAEMON" = "xyes")
 
+#
+# libsystemd
+#
+AC_ARG_ENABLE(libsystemd,
+        AS_HELP_STRING([--disable-libsystemd],[Disable use of libsystemd]),
+        [case "${enableval}" in
```
pemensik, line 831: I would prefer to default to auto value, which should use PKG_CHECK_MODULES to detect readiness of libsystemd development files. If detected, enable it. If not and ``${enablevar}`` were yes, emit fatal error.

configure.ac
```diff
@@ -823,6 +823,25 @@ if test "x$HAVE_LIBDAEMON" = "xyes" ; then
...
+                yes) HAVE_LIBSYSTEMD=yes ;;
+                no)  HAVE_LIBSYSTEMD=no ;;
+                *) AC_MSG_ERROR(bad value ${enableval} for --enable-libsystemd) ;;
+        esac],
+        [HAVE_LIBSYSTEMD=yes])
+
+if test "x$HAVE_LIBSYSTEMD" = "xyes" ; then
+    PKG_CHECK_MODULES(LIBSYSTEMD, [ libsystemd ])
```
pemensik, line 839: Do we want to enforce libsystemd inclusion unless developer explicitly opted it out? I think optional is fine.

pemensik (changes requested):
Optional dependency would be better I think

configure.ac
mbiebl: I followed the existing style of configure.ac, none of the (optional) library dependencies use `auto`, so it would be inconsistent if libsystemd was the only one doing that.

configure.ac
mbiebl: yeah, I think opt-out is fine in this case.

mbiebl:
@evverx are we good to go with merging this patch?

evverx:
I think the CI scripts need tweaking. The following patch should fix the OSS-Fuzz/CI build failues:
```diff
diff --git a/.github/workflows/build.sh b/.github/workflows/build.sh
index 2fe2c71..c2330f4 100755
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -12,7 +12,7 @@ case "$1" in
         sed -i -e '/^#\s*deb-src.*\smain\s\+restricted/s/^#//' /etc/apt/sources.list
         apt-get update -y
         apt-get build-dep -y avahi
-        apt-get install -y libevent-dev qtbase5-dev
+        apt-get install -y libevent-dev qtbase5-dev libsystemd-dev
         apt-get install -y gcc clang lcov
         ;;
     build)
diff --git a/fuzz/oss-fuzz.sh b/fuzz/oss-fuzz.sh
index cdb21c2..4812785 100755
--- a/fuzz/oss-fuzz.sh
+++ b/fuzz/oss-fuzz.sh
@@ -25,7 +25,7 @@ sed -i 's/check_inconsistencies=yes/check_inconsistencies=no/' common/acx_pthrea
     --disable-stack-protector --disable-qt3 --disable-qt4 --disable-qt5 --disable-gtk \
     --disable-gtk3 --disable-dbus --disable-gdbm --disable-libdaemon --disable-python \
     --disable-manpages --disable-mono --disable-monodoc --disable-glib --disable-gobject \
-    --disable-libevent
+    --disable-libevent --disable-libsystemd
 
 make -j"$(nproc)" V=1
 
```

pemensik:
I have made libsystemd-dev usage just optional, which should fix the CI by default. Added devel files installation anyway, we want it tested too. Not sure why should fuzzing skip it. If it can pass, I would keep fuzzing close way to what production builds would be.

pemensik (approved):

pemensik:
@mbiebl Would you be okay with my proposed changes?

evverx:
> Not sure why should fuzzing skip it. If it can pass, I would keep fuzzing close way to what production builds would be.

The OSS-Fuzz runtime environment is different from the environment where fuzz targets are built so to bring runtime dependencies they should either be linked statically or copied to `$OUT` (and `rpath` should be tweaked along the way): https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies. To avoid MSan false positives all the dependencies should be rebuilt with MSan too.

pemensik:
This change should be mentioned in release notes.

evverx:
> If it does not have libsystemd-dev installed during fuzzers build, it should not make it dependent. I expect unless we provide instructions to add more dependencies, it would solve it without additional changes.

It can be included in the base-builder image used by OSS-Fuzz under the hood so I wouldn't rely on it being always absent. I think it would be better to disable libsystemd-dev explicitly to avoid any weird side effects the avahi build script can't directly control. Other than that I don't think it makes sense to fuzz that part of avahi so it shouldn't be built there in the first place probably.

evverx:
I've been thinking about this and while I agree that it's better than including the files into avahi directly I'm a bit concerned that it's yet another thing that should be handled downstream and given that some avahi packages like https://packages.gentoo.org/packages/net-dns/avahi don't seem to be actively maintained anymore I wonder if at this point it would be better to update the files to fix the bugs and switch to libsystemd after the next release or something like that.

evverx:
Either way I think the previous version where avahi failed to build loudly was better because with that it was hard to miss that change but let's wait for @mbiebl here.

pemensik:
> Either way I think the previous version where avahi failed to build loudly was better because with that it was hard to miss that change but let's wait for @mbiebl here.

Do you know a package which uses systemd unit ``Type=notify`` for avahi-daemon by default? I think this would be often just optional feature. I would prefer to have much less optional components in the future. I expect it would work just fine for most people even without libsystemd-dev present. That is why I am okay with it being optional. I admit I have not tested a package for it. I expect Type=dbus do not require notify socket for operations.

evverx:
> Do you know a package which uses systemd unit Type=notify for avahi-daemon by default?

This feature has always been included in avahi by default so someone probably relies on that.

> That is why I am okay with it being optional

I'm not saying that it shouldn't be optional. It's just that I'm not a fan of autodetection and I think that it would be better if it had to be enabled/disabled explicitly. Then again if everyone else thinks it's OK to skip it silently if it isn't installed it's fine by me too.

mbiebl:
Personally I would prefer if the systemd switch would be consistent with the rest of the configure flags.
That said, if @pemensik feels strongly about it, then by all means, use autodetection for this case.
The important issue to me is to get rid of the embedded, outdated copy

mbiebl:
@pemensik would you mind dropping "Make libsystemd usage just optional" from this PR, merging this PR and submitting  "Make libsystemd usage just optional" as a separate, own PR?

evverx:
I'll go ahead and  squash the two CI commits and drop the controversial commit (it can be discussed in a separate PR). Also I agree with @andhe that the ifdiffery can be reduced but let's make it work first and then polish it if necessary.

evverx:
Looks like it doesn't work. systemd doesn't seem to receive any notification messages from avahi and `systemctl status avahi-daemon` doesn't show the "Status" field. I guess it should be reverted. The next attempt should come with a test calling something like `systemctl show --property StatusText avahi-daemon` to make sure it isn't empty.

evverx:
The Pr was reverted in https://github.com/lathiat/avahi/pull/512. I also added a regression test there to make the next attempt easier to review.

mbiebl:
```
michael@pluto:~$ ldd /usr/sbin/avahi-daemon 
	linux-vdso.so.1 (0x00007fff8bd96000)
	libavahi-common.so.3 => /lib/x86_64-linux-gnu/libavahi-common.so.3 (0x00007fd562dbf000)
	libavahi-core.so.7 => /lib/x86_64-linux-gnu/libavahi-core.so.7 (0x00007fd562d86000)
	libdaemon.so.0 => /lib/x86_64-linux-gnu/libdaemon.so.0 (0x00007fd562d7d000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007fd562d52000)
	libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2 (0x00007fd562d46000)
	libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3 (0x00007fd562ced000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd562b0b000)
	libsystemd.so.0 => /lib/x86_64-linux-gnu/libsystemd.so.0 (0x00007fd562a2a000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd562e13000)
	libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20 (0x00007fd5628e3000)
	liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1 (0x00007fd5628bd000)
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007fd56288b000)
	libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007fd5627ca000)
	libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0 (0x00007fd5627a2000)
michael@pluto:~$ systemctl show --property StatusText avahi-daemon --value
avahi-daemon 0.8 starting up.
```

@evverx what exactly are you missing? Did you maybe build avahi without libsystemd-dev installed?

evverx:
It was built with `libsystemd-dev` installed on the CI and the test failed there: https://github.com/evverx/avahi/actions/runs/6737894461/job/18316313459

mbiebl:
Can't reproduce here, unfortunately.

mbiebl:
Anyway, will investigate.

mbiebl:
See #513

PR #460, opened 2023-05-28
ci: import the OSS-Fuzz build script and fuzz targets

evverx:
to make it easier to maintain them.

It's prompted by https://github.com/lathiat/avahi/pull/459 where the build script has to be edited to disable libsystemd.

This PR is based on https://github.com/google/oss-fuzz/pull/9107 and
includes all those changes as well:

* the fuzz targets are rewritten in C;

* the build script is changed to use the headers and static libraries
  from the the build directory

The license was discussed in https://github.com/google/oss-fuzz/pull/10412#discussion_r1208682267.

~~It's a draft for now. I'll point OSS-Fuzz to this PR to make sure it works.~~ To judge from https://github.com/google/oss-fuzz/pull/10412 it works.

### DIFF
--- a/fuzz/fuzz-consume-key.c
+++ b/fuzz/fuzz-consume-key.c
@@ -1,0 +1,44 @@
+/***
+  This file is part of avahi.
+
+  avahi is free software; you can redistribute it and/or modify it
+  under the terms of the GNU Lesser General Public License as
+  published by the Free Software Foundation; either version 2.1 of the
+  License, or (at your option) any later version.
+
+  avahi is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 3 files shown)


lathiat (changes requested):
I'd just like to update the commit message title to be more clear, instead of
"Keep the fuzzing stuff upstream"

Perhaps something like
"Import Google OSS Fuzz support code"

Happy to approve then.

evverx:
@lathiat Thanks! I updated the commit message.

To get it to work https://github.com/google/oss-fuzz/pull/10412 should be merged too but this PR should land first. Once it's merged I'll convert that draft PR on OSS-Fuzz into a normal PR.

PR #461, opened 2023-05-28
Assorted CI tweaks

evverx:
* ci: turn on coverage reports

* ci: build compat-{howl,libdns_sd} as well
    Prompted by https://github.com/lathiat/avahi/pull/465

* make: run the howl and libdns_sd tests automatically
    Prompted by https://github.com/lathiat/avahi/pull/465
    The coverage went from [9.49%](https://coveralls.io/builds/60369370) to [12.944%](https://coveralls.io/builds/60943442) (but long paths aren't tested)

* ci: no longer rely on autobuild
    to get CodeQL to analyze the Python code and the compat libraries
    
* ci: add shellcheck
    Prompted by https://github.com/lathiat/avahi/pull/204

* ci: add dependabot.yml
    https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 2
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -25,7 +25,7 @@ case "$1" in
             export CFLAGS="--coverage"
         fi
 
-        ./bootstrap.sh --enable-tests --prefix=/usr
+        ./bootstrap.sh --enable-compat-howl --enable-compat-libdns_sd --enable-tests --prefix=/usr
         make -j"$(nproc)" V=1
         make check VERBOSE=1
 
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -4,6 +4,7 @@ set -eux
 set -o pipefail
 
 export ASAN_UBSAN=${ASAN_UBSAN:-false}
+export BUILD_ONLY=${BUILD_ONLY:-false}
 export COVERAGE=${COVERAGE:-false}
 
 case "$1" in
@@ -27,6 +28,11 @@ case "$1" in
 
         ./bootstrap.sh --enable-compat-howl --enable-compat-libdns_sd --enable-tests --prefix=/usr
         make -j"$(nproc)" V=1
+
+        if [[ "$BUILD_ONLY" == true ]]; then
+            exit 0
+        fi
+
         make check VERBOSE=1
 
         if [[ "$COVERAGE" == true ]]; then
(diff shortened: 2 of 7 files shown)


Ahmedtaher202 (approved):

pemensik:
Great change, I love the automated checks pipeline. Lets see how much they are stable.

evverx:
@pemensik turns out `address-test` fails on Packit with
```sh
FAIL: address-test
==================

*** WARNING *** The program 'address-test' uses the HOWL compatibility layer of Avahi.
*** WARNING *** Please fix your application to use the native API of Avahi!
*** WARNING *** For more information see <http://0pointer.de/blog/projects/avahi-compat.html>
address-test: address-test.c:39: main: Assertion `r == SW_OKAY' failed.
FAIL address-test (exit status: 134)
```
https://download.copr.fedorainfracloud.org/results/packit/evverx-avahi-3/fedora-37-x86_64/06405616-avahi/builder-live.log.gz

My guess would be that `gethostbyname` doesn't work in that sandbox. I guess the test can be excluded. Another option would be to replace that assertion with something like `if (...) return 77` to skip the test in overzealous environments but I'm not sure if it's worth it.

@jamacku I turned on the differential-shellcheck here and it produced a gazillion alerts :-) I wonder if there's a better way to handle `*.in` files other than skipping them: https://github.com/systemd/systemd/commit/97eb82682126e7f3ee956a025078ea2b801955cb. It would be great if it was possible to analyze the bash scripts after avahi is built somehow.

jamacku:
> @jamacku I turned on the differential-shellcheck here and it produced a gazillion alerts :-) I wonder if there's a better way to handle `*.in` files other than skipping them: [systemd/systemd@97eb826](https://github.com/systemd/systemd/commit/97eb82682126e7f3ee956a025078ea2b801955cb). It would be great if it was possible to analyze the bash scripts after avahi is built somehow.

That should be possible in the new version which should be released in a few weeks. I have added new input [`scan-directory`](https://github.com/redhat-plumbers-in-action/differential-shellcheck#scan-directory). After that, it should be possible able to call Differential ShellCheck after the build step with `scan-directory: build`. I haven't tested this scenario, but since the `build` directory isn't present in the repo, you can't upload the SARIF file to GitHub. But you should be able to see the defects in the console output of the check run.

I plan to propose this to the systemd once the new version of Differential ShellCheck is released. I can make the exact change here.

evverx:
Good to know. Thanks!

> the build directory isn't present in the repo, you can't upload the SARIF file to GitHub. But you should be able to see the defects in the console output of the check run.

I think it's the best compromise here.

The problem with avahi is that the build script generates a subset of bash scripts depending on where avahi builds so shellcheck should keep scanning the *.in files too to cover all the bash scripts. If the `*.in` files were excluded the false positives would be gone along with the true positives. I guess I'd need to figure out how to build all the bash scripts at once and then point shellcheck to them with the scan-directory feature.

jamacku:
> The problem with avahi is that the build script generates a subset of bash scripts depending on where avahi builds so shellcheck should keep scanning the *.in files too to cover all the bash scripts. If the `*.in` files were excluded the false positives would be gone along with the true positives. I guess I'd need to figure out how to build all the bash scripts at once and then point shellcheck to them with the scan-directory feature.

I'm not sure if I understand the problem. confused: We can discuss it later in PR.

Issue #111, opened 2017-03-31
avahi-autoipd.action.linux should flush previously set IP address


chombourger:
Noticed that my wlan0 interface ended-up with 2 wlan0:avahi labels, each having a different IPv4LL address. The first IP address got assigned on boot (by udhcpc?), the Bonjour Conformance Tests generate a bunch of conflicts. We eventually get an address and do our BIND. Unfortunately the avahi-autoipd.action.linux is just adding yet another ip address. 

Will submit a PR to clear any IP addresses previously set on the xxx:avahi interface/label

evverx:
Looks like this issue can be closed.

(I came across this issue mostly because that `flush` command came up in https://github.com/lathiat/avahi/pull/462)

chombourger:
indeed. for completeness, here is the PR that got merged a while ago: #112

PR #462, opened 2023-06-01
Add check for label length in linux autoipd action script.

bnielsen1965:
On linux systems with predictable interface names the avahi-autoipd label with the ":avahi" suffix may exceed the maximum length of 15 characters and fail to create an IPv4LL address on the nic with the long interface name.

This pull request adds a check in the linux autoipd action script and will default to using only the interface name for the label if the :avahi suffix will result in a label that exceeds the maximum length.

### DIFF
--- a/avahi-autoipd/avahi-autoipd.action.linux
+++ b/avahi-autoipd/avahi-autoipd.action.linux
@@ -35,20 +35,33 @@ PATH="$PATH:/usr/bin:/usr/sbin:/bin:/sbin"
 
 METRIC=$((1000 + `cat "/sys/class/net/$2/ifindex" 2>/dev/null || echo 0`))
 
+# determine what will be used for the avahi interface label
+# the length of the label is limited to 15 characters
+
+AVAHI_LABEL=
+# check if interface name + :avahi is too long
+if [ $((${#2} + 6)) -gt 15 ]; then
+  # the label would be too long with the avahi suffix, fall back to using just the interface name
+  AVAHI_LABEL="$2"
+else
+  # the interface name is short enough to allow the avahi suffix
+  AVAHI_LABEL="$2:avahi"
+fi
+
 if [ -x /bin/ip -o -x /sbin/ip ] ; then
 
     # We have the Linux ip tool from the iproute package
 
     case "$1" in
         BIND)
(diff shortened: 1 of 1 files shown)


evverx:
Looks like this issue was reported in https://github.com/lathiat/avahi/issues/345 and to judge from https://bugs.launchpad.net/ubuntu/+source/avahi/+bug/1666900 (which that issue links to) the plan was to drop the label eventually:
> We should probably just drop reliance on the label and store state somehow

bnielsen1965:
@evverx that issue was posted in 2021 and the comment about dropping the label is from 2017, seems like this issue is not getting any traction.

The comment about storing the state is concerning, I now wonder if my script change might cause other issues.

I have a debian 11.3 system with the predictable and long interface names that fail to get an IPv4LL address. So I manually replaced the /etc/avahi/avahi-autoipd.action script and so far it appears to be operating as expected.

With a DHCP service available I get an address as expected, if there is no DHCP service then I get an IPv4LL address. And if I disable the DHCP service and wait for the lease to expire then it reverts to an IPv4LL address and I then re-enable the DHCP service and wait for the debian system to try DHCP again then the IPv4LL address is removed once it has the DHCP provided address.

I'll peruse the source code and see if there is a dependency somewhere for the ":avahi" tag.

bnielsen1965:
I had a look at #111, #112 , did a bit more reading of the C code and some script testing.

At this point I think my PR should not be merged.

Not only does the **ip addr flush** depend on the label but the ifconfig alternative is also dependent on the label.

The flush could theoretically be replaced by an **ip addr del** because the autoipd instance knows which link-local address it has bound to the interface (although that doesn't explain why #111 happened).

But ifconfig is problematic because there doesn't appear to be a reliable way to remove the link-local address if it has been assigned to the interface without a label. I did some testing here and I cannot remove an IPv4 address from an interface with ifconfig once it has been assigned.

I can see why this issue has been around since 2017, this is going to be a bear to resolve while maintaining backward compatibility.

I am available to work on this issue but I don't feel comfortable going any further because there are some tough decisions to be made. Short term I think I can work around the issue via configuration, I noticed that avahi-autoipd has a -t option to specify a custom action script.

PR #468, opened 2023-07-24
Emit the ServiceResolver Found/Failure signals at the correct time

hughsie:
The whole idea of the ServiceResolverPrepare method is that it delays the Found/Failure signals until the client has connected up the D-Bus signal handler. The client then signals it's ready for results by calling Start().

Change the avahi_s_service_resolver_prepare() method to call avahi_s_record_browser_prepare() for the SRV and TXT records rather than avahi_s_record_browser_new() which calls avahi_s_service_resolver_start() much too early.

This fixes an intermittent deadlock when using the org.freedesktop.Avahi.Server2 D-Bus interface with GDBus.

Fixes: https://github.com/lathiat/avahi/issues/446

### DIFF
--- a/avahi-core/resolve-service.c
+++ b/avahi-core/resolve-service.c
@@ -438,7 +438,7 @@ AvahiSServiceResolver *avahi_s_service_resolver_prepare(
     AVAHI_LLIST_PREPEND(AvahiSServiceResolver, resolver, server->service_resolvers, r);
 
     k = avahi_key_new(n, AVAHI_DNS_CLASS_IN, AVAHI_DNS_TYPE_SRV);
-    r->record_browser_srv = avahi_s_record_browser_new(server, interface, protocol, k, flags & ~(AVAHI_LOOKUP_NO_TXT|AVAHI_LOOKUP_NO_ADDRESS), record_browser_callback, r);
+    r->record_browser_srv = avahi_s_record_browser_prepare(server, interface, protocol, k, flags & ~(AVAHI_LOOKUP_NO_TXT|AVAHI_LOOKUP_NO_ADDRESS), record_browser_callback, r);
     avahi_key_unref(k);
 
     if (!r->record_browser_srv) {
@@ -448,7 +448,7 @@ AvahiSServiceResolver *avahi_s_service_resolver_prepare(
 
     if (!(flags & AVAHI_LOOKUP_NO_TXT)) {
         k = avahi_key_new(n, AVAHI_DNS_CLASS_IN, AVAHI_DNS_TYPE_TXT);
-        r->record_browser_txt = avahi_s_record_browser_new(server, interface, protocol, k, flags & ~(AVAHI_LOOKUP_NO_TXT|AVAHI_LOOKUP_NO_ADDRESS),  record_browser_callback, r);
+        r->record_browser_txt = avahi_s_record_browser_prepare(server, interface, protocol, k, flags & ~(AVAHI_LOOKUP_NO_TXT|AVAHI_LOOKUP_NO_ADDRESS), record_browser_callback, r);
         avahi_key_unref(k);
 
         if (!r->record_browser_txt) {


hughsie:
I've added an artificial 500ms delay before calling Start() to highlight the issue:

Screenshot from 2023-07-24 10-48-21

With this fix I now get:

Screenshot from 2023-07-24 20-20-33

pemensik:
It seems reasonable. Would you be able to provide minimal reproducer to test this change? Do we need to modify existing tools to check order stays predictable and is not randomly changing? I have tried watching this by ``dbus-monitor``, but that does not seem to show avahi requests. What have you used for your testing?

hughsie:
In my experience, existing tools are all working around the issue using a global signal handler, so this change shouldn't break them. I'm using bustle to test, that even draws line graphs. 

I'm on holiday with limited internet. But I can get to thevseld test next week if needed. I'm back Wednesday. Apologies

hughsie:
> Would you be able to provide minimal reproducer to test this change

I tried this, and it's harder than I thought -- you actually need to register a service that returns the Found signal -- doing it with something random gives you `org.freedesktop.Avahi.ServiceResolver.Failure` and none of the signals that show the issue. What I was using for testing was:

	// gcc -o example example.c `pkg-config --libs -cflags gio-2.0` && ./example

	#include <gio/gio.h>

	static void
	_sr_signal_cb(GDBusConnection *connection,
			 const gchar *sender_name,
			 const gchar *object_path,
			 const gchar *interface_name,
			 const gchar *signal_name,
			 GVariant *parameters,
			 gpointer user_data)
	{
		g_print("GOT %s.%s\n", interface_name, signal_name);
	}

	static gboolean
	_timeout_cb(gpointer user_data)
	{
		GMainLoop *loop = (GMainLoop *) user_data;
		g_main_loop_quit(loop);
		return G_SOURCE_REMOVE;
	}

	int
	main(int argc, char *argv[])
	{
		g_autoptr(GDBusProxy) proxy = NULL;
		g_autoptr(GError) error = NULL;
		g_autoptr(GVariant) value = NULL;
		g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, FALSE);

		proxy = g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SYSTEM,
						      G_DBUS_PROXY_FLAGS_NONE,
						      NULL,
						      "org.freedesktop.Avahi",
						      "/",
						      "org.freedesktop.Avahi.Server2",
						      NULL,
						      &error);
		if (proxy == NULL)
			g_error("%s", error->message);

		g_dbus_connection_signal_subscribe(g_dbus_proxy_get_connection(proxy),
						   g_dbus_proxy_get_name(proxy),
						   "org.freedesktop.Avahi.ServiceResolver",
						   NULL,
						   NULL,
						   NULL, /* argv */
						   G_DBUS_SIGNAL_FLAGS_NONE,
						   _sr_signal_cb,
						   NULL,
						   NULL);
		value = g_dbus_proxy_call_sync(proxy,
					       "ServiceResolverPrepare",
					       g_variant_new("(iisssiu)",
							     -1, //interface,
							     -1, //protocol,
							     "Passim-8AC4", //name,
							     "_cache._tcp", //type,
							     "", //domain,
							     -1, //AVAHI_PROTO_UNSPEC,
							     0), /* flags */
					       G_DBUS_CALL_FLAGS_NONE,
					       5000,
					       NULL, //cancellable
					       &error);
		if (value == NULL)
			g_error("%s", error->message);
		g_timeout_add(100, _timeout_cb, loop);
		g_main_loop_run(loop);
		g_print("DONE\n");
	}

If you call prepare you wouldn't expect to see a `GOT org.freedesktop.Avahi.ServiceResolver.Found` until you've done `.Start()` but you'll see it every time with that code. The actual code is https://github.com/hughsie/passim/blob/main/src/passim-avahi-service-resolver.c#L246 if that's perhaps more interesting.

evverx:
I think the test can be added to https://github.com/lathiat/avahi/blob/master/avahi-client/client-test.c. It isn't run automatically yet but I'm working on a basic smoke test in https://github.com/evverx/avahi/pull/5 where the test is run among other things. Another option would be to add a separate test probably. It could be run but that smoke as well.

hughsie:
@evverx I'd appreciate any help, I'm not super familiar with the Avahi code I'm afraid.

lathiat:
Good catch @hughsie, thanks :)

evverx:
@hughsie I compiled `example.c`, published `Passim-8AC4._cache._tcp` and reproduced the issue:
```
   19   0.003484       :1.130 → :1.116       D-Bus 208 ServiceResolverPrepare() @ /
   20   0.003654       :1.116 → :1.130       D-Bus 94 -> 'o'
   21   0.154203       :1.116 → :1.130       D-Bus 292 * Found() @ /Client7/ServiceResolver1
```
so I thought it could be added to the test suite. Turns out it isn't as straightforward as I expected :-) I still think it should be covered by the test suite but I guess it can be added later when the testing infrastructure is ready.

evverx (approved):
LGTM. Thanks!

I ran this on top of https://github.com/evverx/avahi/pull/5 just in case and the CI is happy. Those code paths are exercised there (https://coveralls.io/builds/62832879/source?filename=avahi-core%2Fresolve-service.c#L441, https://coveralls.io/builds/62832879/source?filename=avahi-core%2Fresolve-service.c#L451). It's just a smoke test so it doesn't check whether signals are emitted at the right time but I think it should be safe to say that there are no side effects. I manually confirmed that the issue is gone.

PR #472, opened 2023-09-06
Fix avahi-browse: Invalid service type

### LINKED ISSUE #212: avahi-browse -a breaks with "Invalid service type" if a device sends a malformed service
**What happens**
If a device on the network sends a service PTR like "esp32.http.tcp.local" then avahi-browse is broken on every computer on the network:
```
avahi-browse -a
avahi_service_browser_new() failed: Invalid service type
```

avahi-discover-standalone is not affected (it lists an empty line)

**What I expect to happen**
avahi-browse -a lists all found services, maybe an error message for the malformed service

**How to reproduce it**
create a service missing the underscores

mbiebl:
See also https://bugs.archlinux.org/task/71781

Fixes: #212

### DIFF
--- a/avahi-core/browse-service.c
+++ b/avahi-core/browse-service.c
@@ -103,7 +103,9 @@ AvahiSServiceBrowser *avahi_s_service_browser_prepare(
     AVAHI_CHECK_VALIDITY_RETURN_NULL(server, AVAHI_PROTO_VALID(protocol), AVAHI_ERR_INVALID_PROTOCOL);
     AVAHI_CHECK_VALIDITY_RETURN_NULL(server, !domain || avahi_is_valid_domain_name(domain), AVAHI_ERR_INVALID_DOMAIN_NAME);
     AVAHI_CHECK_VALIDITY_RETURN_NULL(server, AVAHI_FLAGS_VALID(flags, AVAHI_LOOKUP_USE_WIDE_AREA|AVAHI_LOOKUP_USE_MULTICAST), AVAHI_ERR_INVALID_FLAGS);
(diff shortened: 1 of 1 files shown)


pemensik:
While this change would be enough to not crash on wrong service announced, I think it might be appropriate to log invalid service to some channel. But should not be done each time it is received. I am not sure there is already sufficent mechanism to remember some issue with announcing machine.

evverx (changes requested):
I'm sorry but I don't think it can be merged.

This patch gets around that particular issue but avahi starts sending queries like
```sh
    Queries
        _invalid._tcp.bogus.service.local: type PTR, class IN, "QM" question
            Name: _invalid._tcp.bogus.service.local
            [Name Length: 33]
            [Label Count: 5]
            Type: PTR (domain name PoinTeR) (12)
            .000 0000 0000 0001 = Class: IN (0x0001)
            0... .... .... .... = "QU" question: False
```
where (`_invalid._tcp` is appended to the bogus service name). It's of course possible to forge bogus PTR and SRV resource records to reply to those queries too and that part isn't addressed at all.

I think it should be either handled by `avahi-browser` (where `service_type_browser_callback` can keep track of services it receives and errors avahi returns). Another option would be to fix `record_browser_callback` in `avahi_s_service_type_browser_prepare` to prevent avahi from passing services like that to `avahi-browse` in the first place but it technically changes the semantics of its API so it should be a last resort probably.

evverx:
As far as I understand the idea was to just extract the patch backported to Arch (and Debian apparently). I think it can be closed.

rossburton:
For anyone else chasing patches, I believe this was fixed in avahi-browse in 93b14365c1c1e04efd1a890e8caa01a2a514bfd8.

PR #474, opened 2023-09-08
Add null check for key comparison

cqlmdnk:
Came across a crash for this code piece, fixed by adding null check. Note that this reproduced rarely in my device.

### DIFF
--- a/avahi-core/rr.c
+++ b/avahi-core/rr.c
@@ -327,6 +327,8 @@ int avahi_key_equal(const AvahiKey *a, const AvahiKey *b) {
     if (a == b)
         return 1;
 
+    if (!(a->name) || !(b->name))
+        return 0;
     return avahi_domain_equal(a->name, b->name) &&
         a->type == b->type &&
         a->clazz == b->clazz;


evverx:
> Came across a crash for this code piece

Could you show the backtrace (preferably with the debug symbols included)?

> Note that this reproduced rarely in my device

I wonder what triggers it? Are there any steps to reproduce it?

cqlmdnk:
(gdb) bt
#0  0xb6f28b84 in avahi_key_equal () from libs/libavahi-core.so.7
#1  0xb6f28e08 in avahi_record_equal_no_ttl () from libs/libavahi-core.so.7
#2  0xb6f19fc0 in avahi_cache_update () from libs/libavahi-core.so.7
#3  0xb6f12844 in mcast_socket_event () from libs/libavahi-core.so.7
#4  0xb6f52d1c in avahi_simple_poll_dispatch () from libs/libavahi-common.so.3
#5  0x00013690 in main ()
(gdb) bt full
#0  0xb6f28b84 in avahi_key_equal () from libs/libavahi-core.so.7
No symbol table info available.
#1  0xb6f28e08 in avahi_record_equal_no_ttl () from libs/libavahi-core.so.7
No symbol table info available.
#2  0xb6f19fc0 in avahi_cache_update () from libs/libavahi-core.so.7
No symbol table info available.
#3  0xb6f12844 in mcast_socket_event () from libs/libavahi-core.so.7
No symbol table info available.
#4  0xb6f52d1c in avahi_simple_poll_dispatch () from libs/libavahi-common.so.3
No symbol table info available.
#5  0x00013690 in main ()
No symbol table info available.
Max info I can get is above.
There is no reproduction steps, as I mentioned it occurred during daily usage and have no logs. Sorry for that.

evverx:
Got it. Thanks! With the backtrace it should hopefully be easier to figure out what triggers it.

I wonder if it was built with assertions disabled (`-DNDEBUG`)?

evverx:
Looking at the backtrace it seems expired entries weren't removed from the cache properly for some reason (and that's where NULLs could have come from). It would be great if you could rebuild avahi with assertions enabled and turn on debug logs. I suspect the cache needs fixing.

cqlmdnk:
Thanks for response, I'll try that.

evverx:
That part of the code is responsible for handling responses so it could be that a weird sequence of ttls and cache-flush bits causes it. It would be ideal if it was possible to collect mDNS replies received by the device before the crash.

evverx:
@cqlmdnk could you turn this PR into an issue? It would be great if it was possible to attach configs, logs and pcaps and also mention the version of avahi.

(If this crash is still reproducible with avahi built from the master branch I don't think this PR fixes the underlying issue because the internal state is still messed up. For example there was a segfault reported in https://github.com/lathiat/avahi/issues/251 and it was suggested there that it was possible to fix it by checking NULLs but the actual issue was much deeper and that check just moved the segfault a bit further without fixing it once and for all).

PR #481, opened 2023-09-19
Run codespell on pull requests

pemensik:
Related to issue #167

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -12,14 +12,10 @@ on:
 permissions:
   contents: read
 
-env:
-  CC: "gcc"
-  BUILD_ONLY: "true"
-
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     concurrency:
       group: ${{ github.workflow }}-${{ matrix.language }}-${{ github.ref }}
       cancel-in-progress: true
@@ -36,5 +32,3 @@ jobs:
 
     - name: Do CodeSpell check
       uses: codespell-project/actions-codespell@v2
-      with:
-        only_warn: 1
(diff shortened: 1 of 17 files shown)


evverx:
FWIW the utf8 errors can be fixed by applying https://github.com/lathiat/avahi/pull/425/commits/bad042c006221b8cff38d3079b2ea745d797a626. clang doesn't like it either.

pemensik:
> FWIW the utf8 errors can be fixed by applying [bad042c](https://github.com/lathiat/avahi/commit/bad042c006221b8cff38d3079b2ea745d797a626). clang doesn't like it either.

Is that commit part of any PR? I think we want similar change. I would use \xfc hex encoding myself, but we do not want to rely on file encoded in latin-1 anymore. Especially if it tries garbage more or less.

evverx:
> Is that commit part of any PR?

As far as I can remember it was included in a PR where `-Werror` was passed to clang. There were (and still are) so many errors that I closed it but GitHub keeps those commits.

pemensik:
I think it is ready

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,34 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ matrix.language }}-${{ github.ref }}
```
evverx, line 20: Looks like `${{ matrix.language }}` came from the codeql workflow and can be removed.

evverx (commented):
I think it would be great to run this on PRs but I'm not sure the code should change. Would it be possible to ignore `fo`, `te`, `etxt` and `calback` instead for the time being?

evverx:
The commits related to the utf8 tests were merged in https://github.com/lathiat/avahi/pull/481

PR #483, opened 2023-09-28
fuzz: no longer download public OSS-Fuzz corpora

evverx:
When the fuzz targets were moved upstream they were renamed to avoid running into issues like
https://github.com/ossf/fuzz-introspector/pull/620. The corpora accumulated by OSS-Fuzz couldn't be renamed though so the idea was to download the old corpora and use them as seed corpora. It worked but now that the old corpora are gone the script can no longer download them and CIFuzz is failing.

CIFuzz downloads the new corpora automatically anyway so it's no longer necessary to keep that part.


evverx:
I'll go ahead and merge it. It should also close https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62660&q=label%3AProj-avahi

PR #486, opened 2023-10-11
common: reject single label host names with unescaped trailing dot

### LINKED ISSUE #451: Reachable assertion in avahi_alternative_host_name
```sh
busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server GetAlternativeHostName "s" ').'
```
```sh
dbus-protocol.c: interface=org.freedesktop.Avahi.Server, path=/, member=GetAlternativeHostName
alternative.c:112: char *avahi_alternative_host_name(const char *): Assertion `avahi_is_valid_host_name(r)' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff788ec0c in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install dbus-libs-1.14.4-1.fc36.x86_64 expat-2.5.0-1.fc36.x86_64 glibc-2.35-20.fc36.x86_64 libcap-2.48-4.fc36.x86_64 libdaemon-0.14-23.fc36.x86_64 libgcc-12.2.1-4.fc36.x86_64 libgpg-error-1.45-1.fc36.x86_64 libzstd-1.5.2-2.fc36.x86_64 lz4-libs-1.9.3-4.fc36.x86_64 systemd-libs-250.8-1.fc36.x86_64 xz-libs-5.2.5-9.fc36.x86_64
(gdb) bt
#0  0x00007ffff788ec0c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff783e986 in raise () from /lib64/libc.so.6
#2  0x00007ffff78287f4 in abort () from /lib64/libc.so.6
#3  0x00007ffff782871b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7837536 in __assert_fail () from /lib64/libc.so.6
#5  0x00007ffff7f8ed85 in avahi_alternative_host_name (s=0x611000000404 ").") at alternative.c:112
#6  0x00000000005410c3 in dbus_get_alternative_host_name (c=<optimized out>, m=0x610000000340, error=0x7fffffffde60) at dbus-protocol.c:377
#7  0x000000000053f7d8 in dbus_select_common_methods (c=<optimized out>, m=0x610000000340, userdata=<optimized out>, iface=<optimized out>, error=<optimized out>) at dbus-protocol.c:1040
#8  0x000000000053e6db in msg_server_impl (c=0x612000000040, m=0x610000000340, userdata=<optimized out>) at dbus-protocol.c:1183
#9  0x00007ffff7cf3d11 in dbus_connection_dispatch () from /lib64/libdbus-1.so.3
#10 0x0000000000561f76 in dispatch_timeout_callback (t=t@entry=0x606000000620, userdata=0x603000000610) at ../avahi-common/dbus-watch-glue.c:105
#11 0x00007ffff7fa1c80 in timeout_callback (t=t@entry=0x606000000620) at simple-watch.c:447
#12 0x00007ffff7fa0d37 in avahi_simple_poll_dispatch (s=s@entry=0x60e000000040) at simple-watch.c:567
#13 0x00007ffff7fa1e7d in avahi_simple_poll_iterate (s=0x60e000000040, timeout=timeout@entry=-1) at simple-watch.c:605
#14 0x000000000051e503 in run_server (c=<optimized out>) at main.c:1268
#15 0x0000000000517bdf in main (argc=<optimized out>, argv=<optimized out>) at main.c:1686
```

msekletar:
Previously, alternative host name we created might have ended up being invalid. We consider single label names with trailing dot valid, e.g. "foo.". However, multi label host names are not valid. Hence, for valid host name "foo."  we created alternative "foo.-2" which was invalid and crashed the daemon.

Fixes #451
CVE-2023-38473

### DIFF
--- a/avahi-common/alternative.c
+++ b/avahi-common/alternative.c
@@ -49,15 +49,20 @@ static void drop_incomplete_utf8(char *c) {
 }
 
 char *avahi_alternative_host_name(const char *s) {
+    char label[AVAHI_LABEL_MAX], alternative[AVAHI_LABEL_MAX*4+1];
+    char *alt, *r, *ret;
     const char *e;
-    char *r;
+    size_t len;
 
     assert(s);
 
     if (!avahi_is_valid_host_name(s))
         return NULL;
 
-    if ((e = strrchr(s, '-'))) {
+    if (!avahi_unescape_label(&s, label, sizeof(label)))
+        return NULL;
+
+    if ((e = strrchr(label, '-'))) {
         const char *p;
 
         e++;
@@ -74,19 +79,18 @@ char *avahi_alternative_host_name(const char *s) {
 
     if (e) {
         char *c, *m;
-        size_t l;
         int n;
 
         n = atoi(e)+1;
         if (!(m = avahi_strdup_printf("%i", n)))
             return NULL;
 
-        l = e-s-1;
+        len = e-label-1;
 
-        if (l >= AVAHI_LABEL_MAX-1-strlen(m)-1)
-            l = AVAHI_LABEL_MAX-1-strlen(m)-1;
+        if (len >= AVAHI_LABEL_MAX-1-strlen(m)-1)
+            len = AVAHI_LABEL_MAX-1-strlen(m)-1;
 
-        if (!(c = avahi_strndup(s, l))) {
+        if (!(c = avahi_strndup(label, len))) {
             avahi_free(m);
             return NULL;
         }
@@ -100,7 +104,7 @@ char *avahi_alternative_host_name(const char *s) {
     } else {
         char *c;
 
-        if (!(c = avahi_strndup(s, AVAHI_LABEL_MAX-1-2)))
+        if (!(c = avahi_strndup(label, AVAHI_LABEL_MAX-1-2)))
             return NULL;
 
         drop_incomplete_utf8(c);
@@ -108,6 +112,13 @@ char *avahi_alternative_host_name(const char *s) {
         r = avahi_strdup_printf("%s-2", c);
         avahi_free(c);
     }
+
+    alt = alternative;
+    len = sizeof(alternative);
+    ret = avahi_escape_label(r, strlen(r), &alt, &len);
+
+    avahi_free(r);
+    r = avahi_strdup(ret);
 
     assert(avahi_is_valid_host_name(r));
 
--- a/avahi-common/alternative-test.c
+++ b/avahi-common/alternative-test.c
@@ -31,6 +31,9 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     const char* const test_strings[] = {
         "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
         "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXüüüüüüü",
+        ").",
+        "\\.",
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\\\\",
         "gurke",
         "-",
         " #",


evverx (changes requested):
Thanks for the PR!

It fixes that particular crash but avahi can still be crashed
```sh
busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server GetAlternativeHostName "s" '\\.'
...
#0  0x00007ffff72b0884 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff725fafe in raise () from /lib64/libc.so.6
#2  0x00007ffff724887f in abort () from /lib64/libc.so.6
#3  0x00007ffff724879b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7258187 in __assert_fail () from /lib64/libc.so.6
#5  0x00007ffff7f77ed4 in avahi_alternative_host_name (s=0x604000001254 "\\\\.") at alternative.c:118
#6  0x000000000042e188 in dbus_get_alternative_host_name (error=0x7ffff4ba0f00, m=0x610000000640, c=0x6120000010c0) at dbus-protocol.c:367
#7  dbus_select_common_methods (c=c@entry=0x6120000010c0, m=m@entry=0x610000000640, userdata=userdata@entry=0x0, 
    iface=iface@entry=0x455480 "org.freedesktop.Avahi.Server", error=error@entry=0x7ffff4ba0f00) at dbus-protocol.c:1038
#8  0x0000000000433764 in msg_server_impl (c=0x6120000010c0, m=0x610000000640, userdata=<optimized out>) at dbus-protocol.c:1181
#9  0x00007ffff7ed9314 in dbus_connection_dispatch () from /lib64/libdbus-1.so.3
#10 0x0000000000446c82 in dispatch_timeout_callback (t=<optimized out>, userdata=0x603000002110) at ../avahi-common/dbus-watch-glue.c:105
#11 0x00007ffff7f808fa in timeout_callback (t=t@entry=0x6060000017c0) at simple-watch.c:447
#12 0x00007ffff7f851c5 in avahi_simple_poll_dispatch (s=s@entry=0x60e000000040) at simple-watch.c:563
#13 0x00007ffff7f857ae in avahi_simple_poll_iterate (s=0x60e000000040, timeout=timeout@entry=-1) at simple-watch.c:605
#14 0x0000000000410254 in run_server (c=0x4d8b20 <config>) at main.c:1268
#15 main (argc=<optimized out>, argv=<optimized out>) at main.c:1693
```
I think `is_valid_hostname` should reject labels with dots (which should address https://github.com/lathiat/avahi/issues/453 too) and after that it should ignore everything except the first valid label.

Anyway I think it would be great to add a test case to `avahi-common/alternative-test.c` and also add `avahi_alternative_host_name` to `fuzz-domain` to make sure that the parser can withstand all sorts of weird things. In the meantime I'll go ahead and merge https://github.com/lathiat/avahi/pull/482 to make it possible for the CI to poke things with dfuzzer. Currently the CI doesn't cover a lot of things.

evverx:
> For example in https://github.com/lathiat/avahi/issues/453 the problem is that we split on dot but we really shouldn't because it is actually escaped.

Indeed. Looks like instead of `avahi_normalize_name_strdup` and `hn[strcspn(hn, ".")] = 0` it should call `avahi_unescape_label` to extract the first label. (I'd need to double-check though)

avahi-common/alternative.c
```diff
@@ -57,6 +57,11 @@ char *avahi_alternative_host_name(const char *s) {
     if (!avahi_is_valid_host_name(s))
         return NULL;
 
+    /* We consider multi label host names invalid. If we have single label
+     * name with unescaped trailing dot we would create such invalid name. */
+    if (avahi_is_dot_suffixed_host_name(s))
+        return NULL;
```
evverx, line 63: I've been thinking about this and I think the only way to make this function work is to unescape the label with `avahi_unescape_label`, then let it do its thing and escape the result back with `avahi_escape_label`. It should hopefully fix it once and for all. Right now the whole thing is buggy and can be crashed without any dots: https://github.com/lathiat/avahi/issues/487.

evverx: Either way I have to admit I'm not sure what this function is supposed to do. Looking at https://github.com/lathiat/avahi/issues/478 (where `avahi-publish` passes the hostname along with the domain name to `avahi_alternative_host_name`) it seems either `is_valid_host_name` should be replaced with `is_valid_fqdn` or `avahi-publish` should extract the host name, pass it to `avahi_alternative_host_name` and then join the result with the domain name. I'd go with the latter to avoid any backward incompatible changes but still.

evverx: All in all I'm inclined to say that unescaping the label, turning that unescaped label into its alternative name and escaping that alternative name with `avahi_escape_label`  and returning that is the way to go.

msekletar: I think this would change the behavior quite a bit. The problem is that [RFC](https://datatracker.ietf.org/doc/html/rfc4343#section-2.1) allows for escaping all printable ASCII characters but it is not necessarily what people expect. 

For example, if we implement your proposal (I tried locally as PoC) then for host name like `[server][basement]` the returned alternative is (after escaping) `\091server\093\091basement\093-2` while previously it was `[server][basement]-2`.

Is suppose we could change how escaping works and make sure we only escape characters in ranges 0x00 to 0x20,  0x7F to 0xFF in \DDD style and we would continue escaping dot and backslash as usual. Then the returned string for the example I gave won't change. However, I am not sure what other parts of avahi might break.

Of course, alternative is to accept my proposal and also work on addressing the issue you filed in #487.

evverx: > The problem is that [RFC](https://datatracker.ietf.org/doc/html/rfc4343#section-2.1) allows for escaping all printable ASCII characters but it is not necessarily what people expect

I think if avahi is used it's kind of expected because all the names are usually normalized. For example if `host-name` is set to `[server][basement]` in the config file avahi shows something like
```sh
Server startup complete. Host name is \091server\093\091basement\093.local. Local service cookie is 1167619599.
```
when it starts and `avahi-browse` shows
```sh
+;eth0;IPv6;\091server\093\091basement\093;_ssh._tcp;local
```
It's true that the D-Bus API has never normalized anything (and that's probably why it crashes so much) but names that are successfully passed around end up being escaped one way or another. Other than that I'm not sure this D-Bus API is widely used (I'm not even sure why it exists) so all in all I think it should be fine to change that.

msekletar: I just wanted to point out that would be changing behavior, but I also agree it is likely OK.

avahi-common/alternative.c
```diff
@@ -49,15 +49,20 @@ static void drop_incomplete_utf8(char *c) {
 }
 
 char *avahi_alternative_host_name(const char *s) {
+    char label[AVAHI_LABEL_MAX], alternative[AVAHI_LABEL_MAX*4+1];
```
evverx, line 52: I think it should be `(AVAHI_LABEL_MAX-1)*4+1` but it doesn't matter much probably.

avahi-common/alternative.c
```diff
@@ -109,6 +113,13 @@ char *avahi_alternative_host_name(const char *s) {
         avahi_free(c);
     }
 
+    alt = alternative;
+    len = sizeof(alternative);
+    ret = avahi_escape_label(r, strlen(r), &alt, &len);
+
+    avahi_free(r);
+    r = avahi_strdup(ret);
```
evverx, line 121: `r` can be `NULL` so it should be checked probably.

evverx (commented):
I left a few comments but on the whole it looks good. Let me check that the daemon can handle those names when conflicts occur.

avahi-common/alternative.c
```diff
@@ -109,6 +113,13 @@ char *avahi_alternative_host_name(const char *s) {
         avahi_free(c);
     }
 
+    alt = alternative;
+    len = sizeof(alternative);
+    ret = avahi_escape_label(r, strlen(r), &alt, &len);
```
evverx, line 118: `ret` should probably be checked too.

avahi-common/alternative.c
evverx: Looks like the check isn't necessary here. NULL is either caught by the assertion or just returned if assertions are turned off.

avahi-common/alternative.c
evverx: I guess it can be changed in a separate commit where something like `AVAHI_LABEL_ESCAPED_MAX` can be introduced. It can be used to extract hostnames properly later.

avahi-common/alternative.c
evverx: It can fail only if there isn't enough room and it can always fit into `alt` in its current form. The fuzz target should trigger the assertion if there is anything wrong here.

evverx:
There is an integer overflow there apparently: https://github.com/lathiat/avahi/issues/491 but it's a separate issue too.

PR #488, opened 2023-10-18
core: cap maximum size of DNS packet at AVAHI_DNS_PACKET_SIZE_MAX

### LINKED ISSUE #455: Reachable assertion in avahi_dns_packet_append_record

msekletar:
Fixes #455

CVE-2023-38469

### DIFF
--- a/avahi-core/dns.c
+++ b/avahi-core/dns.c
@@ -49,6 +49,8 @@ AvahiDnsPacket* avahi_dns_packet_new(unsigned mtu) {
 
     if (max_size < AVAHI_DNS_PACKET_HEADER_SIZE)
         max_size = AVAHI_DNS_PACKET_HEADER_SIZE;
+    else if (max_size > AVAHI_DNS_PACKET_SIZE_MAX)
+        max_size = AVAHI_DNS_PACKET_SIZE_MAX;
 
     if (!(p = avahi_malloc(sizeof(AvahiDnsPacket) + max_size)))
         return p;
--- a/fuzz/fuzz-consume-key.c
+++ b/fuzz/fuzz-consume-key.c
@@ -29,6 +29,10 @@ void log_function(AvahiLogLevel level, const char *txt) {}
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     avahi_set_log_function(log_function);
     AvahiDnsPacket* packet = avahi_dns_packet_new(size + AVAHI_DNS_PACKET_EXTRA_SIZE);
+
+    if (size + AVAHI_DNS_PACKET_EXTRA_SIZE > AVAHI_DNS_PACKET_SIZE_MAX)
+        size = packet->max_size - AVAHI_DNS_PACKET_EXTRA_SIZE;
+
     memcpy(AVAHI_DNS_PACKET_DATA(packet), data, size);
     packet->size = size;
     AvahiKey* key = avahi_dns_packet_consume_key(packet, NULL);
--- a/fuzz/fuzz-consume-record.c
+++ b/fuzz/fuzz-consume-record.c
@@ -29,6 +29,10 @@ void log_function(AvahiLogLevel level, const char *txt) {}
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     avahi_set_log_function(log_function);
     AvahiDnsPacket* packet = avahi_dns_packet_new(size + AVAHI_DNS_PACKET_EXTRA_SIZE);
+
+    if (size + AVAHI_DNS_PACKET_EXTRA_SIZE > AVAHI_DNS_PACKET_SIZE_MAX)
+        size = packet->max_size - AVAHI_DNS_PACKET_EXTRA_SIZE;
+
     memcpy(AVAHI_DNS_PACKET_DATA(packet), data, size);
     packet->size = size;
     AvahiRecord* rec = avahi_dns_packet_consume_record(packet, NULL);


evverx:
That one is tricky. I haven't figured out how to deal with it yet.

Looks like ASan isn't happy
```sh
==26==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x63100040c945 at pc 0x00000052e3ec bp 0x7ffd92e7f5e0 sp 0x7ffd92e7edb0
WRITE of size 128959 at 0x63100040c945 thread T0
SCARINESS: 45 (multi-byte-write-heap-buffer-overflow)
    #0 0x52e3eb in __asan_memcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:22:3
    #1 0x56c46b in LLVMFuzzerTestOneInput /src/avahi/fuzz/fuzz-consume-record.c:32:5
    #2 0x43de13 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #3 0x43d5fa in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #4 0x43f464 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:826:7
    #5 0x43f699 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:857:3
    #6 0x42ecff in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #7 0x458352 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #8 0x7f1e4e00[50](https://github.com/lathiat/avahi/actions/runs/6563046959/job/17826378396?pr=488#step:5:51)82 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: e678fe[54](https://github.com/lathiat/avahi/actions/runs/6563046959/job/17826378396?pr=488#step:5:55)a5d2c2092f8e47eb0b33105e380f7340)
    #9 0x41f73d in _start (build-out/fuzz-consume-record+0x41f73d)
```

msekletar:
Service announcing is done through EntryGroup objects, i.e. you create entry group, you add service to it and then we announce it. Although, there is a concept of "failing" the entry group (`AVAHI_ENTRY_GROUP_FAILURE`) it seems to be unused in the daemon implementation and instead we transition entry group to established state even when we failed to announce the service. I think that should be fixed but can be done later as a normal bugfix rather than security fix (i.e. let's make CVE fix as small as possible).

evverx:
Let me think about it. The tests don't cover those corner cases. Assertions like
```
assert(p->size + p->res_size <= p->max_size)
```
in other places were probably added based on the assumption that `max_size` is usually enough to store stuff but given that this PR truncates `max_size` but those huge blobs are still passed around I'm not sure whether those assumptions are still correct.

Generally there are a lot of other ways to crash avahi so I believe this particular crash can wait a bit.

msekletar:
Hmmm, but I think that `max_size` should have always been capped at `AVAHI_DNS_PACKET_SIZE_MAX`..... value bigger than that doesn't make sense as there is no DNS protocol fragmentation so if this assert is in contradiction with that then it really means that assert is incorrect to begin with.

evverx:
> value bigger than that doesn't make sense as there is no DNS protocol fragmentation so if this assert is in contradiction with that then it really means that assert is incorrect to begin with

The problem is that those code paths aren't exercised anywhere so at this point I can't say for sure that this PR doesn't introduce new crashes or buffer overflows. If it was possible to somehow cover those places with tests it would help.

evverx:
I agree that those packets don't make sense but avahi still somehow ends up sending those large probes/records with this PR applied, which means those blobs can overflow `max_size` somewhere in theory and I'm not sure how to check all the places where it can happen.

msekletar:
Actually it seems that even packets as big as `AVAHI_DNS_PACKET_SIZE_MAX` are just too big for mDNS. Reading RFC it says following in section [17](https://datatracker.ietf.org/doc/html/rfc6762#section-17).

```
A Multicast DNS packet larger than the interface MTU, which is sent
using fragments, MUST NOT contain more than one resource record.

Even when fragmentation is used, a Multicast DNS packet, including IP
and UDP headers, MUST NOT exceed 9000 bytes.
```

IOW, trying to construct packets we MUST NOT sent (even as multiple IP fragments) seems like a wasted CPU time.

evverx:
Ideally depending on what avahi sends it should set the truncated bit and split large packets. I don't think it's supported though. Large unicast legacy stuff should go via TCP (I don't think it's supported either).

Either way I'm not saying that I don't agree with the RFC. I'm saying that at this point there are no tests that can show what avahi actually does. It should be possible to add the command from the bug report to the "smoke" test to cover some of those code paths but I don't think that would be enough to be sure that avahi doesn't go sideways anywhere else.

evverx:
I'd go with something like
```diff
@@ -693,7 +694,7 @@ int avahi_record_is_valid(AvahiRecord *r) {
                 if (strlst->size > 255 || strlst->size <= 0)
                     return 0;
 
-            return 1;
+            return avahi_string_list_serialize(r->data.txt.string_list, NULL, 0) <= AVAHI_DNS_RDATA_MAX;
         }
     }
```
for now. With that patch applied the crash is gone and avahi rejects obviously nonsensical TXT resource records with
```sh
avahi-publish -s T _qotd._tcp 22 $(perl -le 'print "A " x 100000')
Failed to add service: Invalid record
```
```sh
...
dbus-entry-group.c: interface=org.freedesktop.Avahi.EntryGroup, path=/Client1/EntryGroup1, member=GetState
dbus-entry-group.c: interface=org.freedesktop.Avahi.EntryGroup, path=/Client1/EntryGroup1, member=IsEmpty
dbus-entry-group.c: interface=org.freedesktop.Avahi.EntryGroup, path=/Client1/EntryGroup1, member=AddService
dbus-util.c: Responding error 'Invalid record' (-9)
dbus-protocol.c: client :1.176 vanished.
```
It doesn't fully address the issue with large packets in general but I guess it's better than nothing (at least until we figure out how to deal with large packets. Looking at https://github.com/lathiat/avahi/issues/23 I'm not sure it's going to happen anytime soon though :-)).

msekletar:
Are you going to prep the PR with the proposed fix?

evverx:
I can send the patch once I've gotten round to it. It needs polishing though because `avahi_record_is_valid` shouldn't go over `data.txt.string_list` twice. It can calculate the total length when it looks at `strlst->size` by analogy with how `avahi_string_list_serialize` does it. If it's urgent it would be faster if it was implemented here probably.

evverx:
The patch is ready
```diff
diff --git a/avahi-core/rr.c b/avahi-core/rr.c
index 2bb8924..9c04ebb 100644
--- a/avahi-core/rr.c
+++ b/avahi-core/rr.c
@@ -32,6 +32,7 @@
 #include <avahi-common/malloc.h>
 #include <avahi-common/defs.h>
 
+#include "dns.h"
 #include "rr.h"
 #include "log.h"
 #include "util.h"
@@ -689,11 +690,17 @@ int avahi_record_is_valid(AvahiRecord *r) {
         case AVAHI_DNS_TYPE_TXT: {
 
             AvahiStringList *strlst;
+            size_t used = 0;
 
-            for (strlst = r->data.txt.string_list; strlst; strlst = strlst->next)
+            for (strlst = r->data.txt.string_list; strlst; strlst = strlst->next) {
                 if (strlst->size > 255 || strlst->size <= 0)
                     return 0;
 
+                used += 1+strlst->size;
+                if (used > AVAHI_DNS_RDATA_MAX)
+                    return 0;
+            }
+
             return 1;
         }
     }
```
it doesn't go over TXT RRs twice and bails out as soon as the total size exceeds 65535. I'll add some tests and open a PR.

I'll go ahead and close this PR. Generally I agree that avahi should limit the size of packets somewhere (and it seems to me that it should return errors instead of proceeding because it makes it less likely that any part of the code even tries to write anything there but I'm reluctant to change anything there as long as there are no tests covering those codepaths).

evverx:
-> https://github.com/lathiat/avahi/pull/500

PR #509, opened 2023-10-31
Drop support for Qt3

mbiebl:
The last release of Qt3 was in 2004. It is no longer supported upstream and no major distro still ships it.

Qt3 is looooong dead


evverx:
Personally I think it should be fine to drop it so LGTM.

I'd wait for the others to chime in though.

pemensik:
I would keep it for 0.9 release and drop it later. If that still compiles with qt3 headers, let's keep it intact. But we should call it obsolete somewhere.

mbiebl:
fwiw, I would probably also drop the Qt4 binaries unless we have some very basic build coverage

evverx:
> If that still compiles with qt3 headers

I'm not sure it does. The CI doesn't build it at all.

Generally I think all those clearly unmaintained parts should be either dropped (like this particular dependency) or EOLed. To judge from feature requests it's somehow still believed that the compat libraries, avahi-autoipd and so on are maintained. Those should be kept of course but nobody should expect that any new features are ever going to end up here. I'm not sure bugs are ever going to be fixed there either.

mbiebl:
> But we should call it obsolete somewhere.

If you want to postpone this to 0.9.1, this would be fine with me. The release notes of 0.9 could be used to document which parts/bindings are declared obsolete and will be removed in (one of) the next releases

pemensik:
RHEL/CentOS 7 contains both qt3 and qt5 development files. I admit it is quite old and probably no-one needs it for a long time. But I would not remove it for upcoming release. Deprecating Qt4 bindings should be done at the same time, even that is old enough.

pemensik:
We can always move it into separate project at avahi organization and leave it there to die or strive. I think we need to simplify main avahi project to focus on core components. Bindings should be best effort, especially for old libraries or unusual languages. But such split might be done at version 0.10 (or 1.0?). For now I would keep it as much compatible as we can afford.

mbiebl:
> Deprecating Qt4 bindings should be done at the same time, even that is old enough.

You're probably right. I've updated the PR accordingly.

mbiebl:
> RHEL/CentOS 7 contains both qt3 and qt5 development files.

I've updated the commit message slightly, thanks for the hint. From what I can see, RHEL/CentOS 7 doesn't build the Qt3 bindings though.

evverx:
This one conflicts with https://github.com/avahi/avahi/pull/704. It should be merged one way or another but the conflicts should be resolved. I'll leave to @pemensik to sort it out.

PR #513, opened 2023-11-02
Stop using embedded copy of sd-daemon.{c,h} and use libsystemd instead

mbiebl:
Fixes: #287
Fixes: #296


mbiebl:
Somewhat related (although not a result of this PR):
If chroot support is enabled, ` avahi-daemon 0.8 starting up.` is the last sd_notify message that is received and no further notifications reach systemd.

evverx:
Looks like the culprit was `AC_DEFINE([HAVE_LIBSYSTEMD],...)`. It all adds up now.

The CI confirmed that avahi reached systemd so let's try again :-)

evverx:
@mbiebl I took a look at the NixOS package and it seems it's built on *BSD systems too. To make it work: `with-systemdsystemunitdir=no` is passed to turn off `systemd` (by flipping `HAVE_SYSTEMD`): https://github.com/NixOS/nixpkgs/blob/c4c6047b3cbc0a8d24f3113fc259658b67fc4776/pkgs/development/libraries/avahi/default.nix#L102-L103. I guess it would make sense to look for `libsystemd` (and fail if it isn't installed) depending on `HAVE_SYSTEMD`. It's different from the autodetection logic proposed in https://github.com/lathiat/avahi/pull/459 but I think it makes sense. @jtojnar what do you think?

evverx:
Even though it's possible to turn it off explicitly with `--disable-libsystemd` I guess it can be annoying on non-linux systems

jtojnar:
Looks like we are passing `--with-systemdsystemunitdir=no` [unconditionally](https://github.com/NixOS/nixpkgs/blob/c4c6047b3cbc0a8d24f3113fc259658b67fc4776/pkgs/development/libraries/avahi/default.nix#L103) because we [create our own service file](https://github.com/NixOS/nixpkgs/blob/c4c6047b3cbc0a8d24f3113fc259658b67fc4776/nixos/modules/services/networking/avahi-daemon.nix#L304) on Linux. This is likely to enable syslog and to use our config. I would rather prefer enabling syslog at build time and put the config to `/etc` so that we can get rid of the custom service file.

As a packager, I generally prefer explicit disable flag over auto detection since the latter makes it easy to accidentally omit a dependency. It can be a bit annoying when the dependency is not available for some platforms but I still consider the annoyance worth it in preventing broken functionality.

As a developer, auto flags can be convenient but unless the build system has an option to force-enable them like [Meson does](https://mesonbuild.com/Build-options.html#features), I would go with packaging reliability over not having to install all dependencies/configuring them out. Developers will probably want to build with all features anyway.

evverx:
Got it. Thanks! I'm not a fan of autodetection either (and that's why I removed that commit from the first PR) but it was based on the assumption that it was built on Linux where it makes sense to fail loudly given that this feature had always been enabled by default by embedding libsystemd.

I guess it should be fine to keep it as is then. Thanks again!

evverx:
In light of https://www.openwall.com/lists/oss-security/2024/03/29/4 it seems the trend is to implement `sd_notify` manually. To judge from a systemd issue where the libsystemd dependencies tree is discussed Luca is planning to add a copy-pastable example soon as far as I understand. @mbiebl what do you think? Or would Debian prefer the explicit dependency? Both approaches are fine by me.

mbiebl:
Embedded code copies have the tendency to get stale, which was the reason for #296 to begin with, so tbh I'm no a huge fan of it and maybe the next vulnerability is because of an outdated code copy. Who knows.
The recommendation to inline / re-implement the code for sd_notify is a bit of a surprise to me as the suggested best practice so far was to use libsystemd.
Also, keep in mind that avahi uses not only the sd_notify API but also `sd_booted`.

That all said, I acknowledge the problem when libraries (excessively) pull in transitive dependencies.
In libsystemd this is specifically the case for sd_journal and its different compression formats.
One way to address this would be, to make an opinionated choice and disable lz4,bzip2,xz,gzip and only keep zstd.
When I discussed this within the debian-systemd team, @bluca mentioned that with the next upstream release of systemd all those libraries will be dlopened, i.e. only be loaded on demand.
So maybe this is good enough.

PR #515, opened 2023-11-03
core: compare RRs with zero-length rdata correctly

evverx:
RRs where types are unknown to avahi can't be passed to
lexicographical_memcmp directly because it doesn't expect NULL pointers
and crashes. This patch prevents it from crashing by treating those NULL
pointers as empty strings and returning -1, 0 or 1. It in turn prevents
avahi from crashing on receiving probes with zero-length rdata if
unusual RRs have ever been added using avahi_entry_group_add_record.

This patch also fixes a bug where RRs with no remaining bytes were
deemed lexicographically later even though according to
https://datatracker.ietf.org/doc/html/rfc6762#section-8.2

The bytes of the raw uncompressed rdata are compared ...
until a byte is found whose value is greater than that of its counterpart
... or one of the resource records runs out of rdata (in which case,
the resource record which still has remaining data first is deemed
lexicographically later)

The test were added to make sure it actually works as expected and the fuzz
target was updated to make sure it no longer crashes anywhere.

Closes https://github.com/lathiat/avahi/issues/493

### DIFF
--- a/avahi-core/dns-test.c
+++ b/avahi-core/dns-test.c
@@ -33,6 +33,7 @@
 
 #include "dns.h"
 #include "log.h"
+#include "rr-util.h"
 #include "util.h"
 
 int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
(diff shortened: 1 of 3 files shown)


evverx (commented):
TBR

I'll go ahead and merge it to let Coverity and OSS-Fuzz start testing this code.

PR #529, opened 2023-11-17
cleanup: rm travis config, fedora/suse init scripts and bespoke fuzzer

evverx:
Travis is no longer used.

Fedora dropped the init scripts 12 years ago in
https://src.fedoraproject.org/rpms/avahi/c/d6e5bc71835e8f7e8238e40c6b489f99e95e9d38?branch=rawhide
so they shouldn't be kept upstream either.

openSUSE dropped the init scripts 8 years ago in
https://build.opensuse.org/request/show/359515 so they shouldn't be kept
upstream either.

The fuzzer doesn't fuzz much. Other than that the scapy API has changed so it
doesn't work anymore.


mbiebl:
Please drop the Debian initscripts as well while you are at it.
https://salsa.debian.org/utopia-team/avahi/-/commit/41a296309334853ed3164544bd39d872a8fc70a4

mbiebl:
since mandriva is dead since 2011, it's probably safe to drop those init scripts as well.

PR #531, opened 2023-11-17
Drop support for HOWL compatibility layer

mbiebl:


pemensik:
I have found that on Fedora there are still 2 packages depending on avahi-compat-howl-devel:
- https://src.fedoraproject.org/rpms/ngircd (https://ngircd.barton.de/)
- https://src.fedoraproject.org/rpms/player (http://playerproject.github.io/)

evverx:
If Fedora is the only distro using this I wonder if Fedora can ship this downstream?

> Just mention clearly it is not recommended and actively maintained part.

The compat layer has been officially deprecated since 2005 so in a way it has been done for almost 20 years.

evverx:
I rebased it on top of the master branch, resolved the conflicts and made the CI green. It fixed a lot of gcc warnings and increased the coverage of avahi. It should be good to go once 0.9 is tagged.

Issue #540, opened 2023-12-07
Repository settings


evverx:
This issue is supposed to keep track of various settings that should be tweaked eventually to get Packit, Coverity, Coveralls and some other things to work:
* https://docs.coveralls.io/index#integrate-coveralls-with-your-codebase
* ~~https://packit.dev/docs/guide#github~~
* ~~a secret named COVERITY_SCAN_TOKEN should be added to the repository: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository to make the Coverity action work. It should match https://scan.coverity.com/projects/avahi-daemon?tab=project_settings~~
* ~~https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository~~
* ~~https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches~~

lathiat:
@evverx Since moving to an org, I seem now able to give you "Maintain" access to specifically the avahi repository, which may let you adjust these settings. Can you check and see if that works for you - and if there is anything you can't adjust but want to then document it here and we can look to adjust it.

Permissions are documented here:
https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization

Neustradamus:
To follow this ticket :)

pemensik:
@lathiat We would like to have enabled *Private vulnerability reporting* in https://github.com/avahi/avahi/settings/security_analysis and access to such advisories. I am not able to see any now in Security tab.

I have played with it at my fork https://github.com/pemensik/avahi/security/advisories. It seems we would need multiple of advisories. They would allow private collaboration when preparing fixes, then publishing it once fixes are ready. Not sure how to define whose people are able to see reported vulnerabilities, I haven't found that in my fork. But I have no collaborators :)

pemensik:
It seems at least Admin rights are needed for Security advisory creation or reading, unless people are given collaborator rights for given advisory. Our commit (*Member*) rights are not enough for it. Unless Trent would take more active role, we need someone in addition to have *Admin* rights. Then those settings could be tuned be such person.

lathiat:
I will sort this out on Monday

lathiat:
I enabled private vulnerability reporting in both the avahi and nss-mdns repositories (just the 1-click enable).

[...]

PR #541, opened 2023-12-08
ci: build avahi on aarch64, i386, ppc64le, s390x, x86_64 on PRs

evverx:
It already works:
https://github.com/evverx/avahi/pull/3/checks?check_run_id=19440388533
https://copr.fedorainfracloud.org/coprs/packit/evverx-avahi-3/builds/

Depends on https://github.com/avahi/avahi/issues/540

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,42 @@
+---
+specfile_path: .packit_rpm/avahi.spec
+files_to_sync:
+  - .packit.yml
+  - src: .packit_rpm/avahi.spec
+    dest: avahi.spec
+upstream_package_name: avahi
+downstream_package_name: avahi
+upstream_tag_template: "v{version}"
+srpm_build_deps: []
+
+actions:
+  post-upstream-clone:
+    # Use the Fedora Rawhide specfile
+    - "git clone https://src.fedoraproject.org/rpms/avahi .packit_rpm --depth=1"
+    # Drop the "sources" file so rebase-helper doesn't think it's a dist-git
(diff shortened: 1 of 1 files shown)


pemensik (approved):
This seems nice. Do we have any avahi COPR repository already, where this can be tried?

pemensik:
From my team I am working in, only tuned [uses packit](https://github.com/redhat-performance/tuned/blob/master/.packit.yaml) in their upstream project. I were asked, why we do not use ``fedora-all`` in our proposal. Is there a specific reason for it?

evverx:
> Do we have any avahi COPR repository already, where this can be tried?

It's built at https://copr.fedorainfracloud.org/coprs/packit/evverx-avahi-3/ regularly but the avahi repository itself hasn't been hooked up yet.

> why we do not use fedora-all in our proposal. Is there a specific reason for it?

rawhide is kind of unstable (for example today one build failed on rawhide-i386: https://copr.fedorainfracloud.org/coprs/packit/evverx-avahi-3/build/6946551/) so it's easier to temporarily turn it off by commenting it out if it's mentioned explicitly. `fedora-all` should work too though.

evverx:
I'll go ahead and merge it to make it easier for me to run those builds using my fork at least. Currently I have to cherry-pick this commit on top of PRs every time. It should still be hooked up but it's tracked in a separate issue. It still builds fine: https://copr.fedorainfracloud.org/coprs/packit/evverx-avahi-3/build/7018789/

Issue #478, opened 2023-09-16
`avahi-publish -a` crashes when conflicts occur


evverx:
```sh
avahi-publish -a C.local 192.168.82.194
Established under name 'C.local'
Name collision, picking new name '(null)'.
avahi-publish: entrygroup.c:738: int avahi_entry_group_add_address(AvahiEntryGroup *, AvahiIfIndex, AvahiProtocol, AvahiPublishFlags, const char *, const AvahiAddress *): Assertion `name' failed.
Aborted (core dumped)
```
```sh
Program received signal SIGABRT, Aborted.
0x00007ffff7c55884 in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install dbus-libs-1.14.10-1.fc38.x86_64 glibc-2.37-5.fc38.x86_64 libcap-2.48-6.fc38.x86_64 libgcc-13.2.1-1.fc38.x86_64 libzstd-1.5.5-1.fc38.x86_64 lz4-libs-1.9.4-2.fc38.x86_64 systemd-libs-253.9-1.fc38.x86_64 xz-libs-5.4.1-1.fc38.x86_64
(gdb) bt
#0  0x00007ffff7c55884 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff7c04afe in raise () from /lib64/libc.so.6
#2  0x00007ffff7bed87f in abort () from /lib64/libc.so.6
#3  0x00007ffff7bed79b in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7bfd187 in __assert_fail () from /lib64/libc.so.6
#5  0x00007ffff7f8e8d2 in avahi_entry_group_add_address (group=<optimized out>, interface=<optimized out>, protocol=<optimized out>, flags=<optimized out>, 
    name=0x0, a=<optimized out>) at entrygroup.c:738
#6  0x00000000004f5a7e in register_stuff (config=<optimized out>) at avahi-publish.c:123
#7  0x00007ffff7f85680 in filter_func (bus=<optimized out>, message=<optimized out>, userdata=<optimized out>) at client.c:219
#8  0x00007ffff7f23f79 in dbus_connection_dispatch () from /lib64/libdbus-1.so.3
#9  0x00007ffff7fa15b8 in dispatch_timeout_callback (t=<optimized out>, userdata=0x603000001450) at ../avahi-common/dbus-watch-glue.c:105
[...]

PR #547, opened 2024-01-08
Altsvc name int overflow

### LINKED ISSUE #526: OSS-Fuzz issue 64170
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5604692931248128 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64170 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

pemensik:
Fixes #526

### DIFF
--- a/avahi-utils/avahi-publish.c
+++ b/avahi-utils/avahi-publish.c
@@ -91,11 +91,15 @@ static void entry_group_callback(AvahiEntryGroup *g, AvahiEntryGroupState state,
                 n = avahi_alternative_host_name(config->name);
             }
 
-            fprintf(stderr, _("Name collision, picking new name '%s'.\n"), n);
-            avahi_free(config->name);
-            config->name = n;
+	    if (n) {
+		fprintf(stderr, _("Name collision, picking new name '%s'.\n"), n);
+		avahi_free(config->name);
+		config->name = n;
 
-            register_stuff(config);
+		register_stuff(config);
+	    } else
+		fprintf(stderr, _("Name collision, failed to pick a new name for '%s'.\n"),
+				config->name);
 
             break;
         }
--- a/avahi-common/alternative-test.c
+++ b/avahi-common/alternative-test.c
@@ -59,6 +59,7 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
         "gurke---",
         "gurke #",
         "gurke ###",
+	"testalt #2147483548", /* MAX_INT-100 */
         NULL
     };
 
@@ -74,7 +75,13 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
             for (i = 0; i <= 100; i++) {
                 char *n;
 
-                n = i == 0 ? avahi_strdup(test_strings[k]) : (j ? avahi_alternative_service_name(r) : avahi_alternative_host_name(r));
+		if (i == 0)
+		    n = avahi_strdup(test_strings[k]);
+		else if (j)
+		    n = avahi_alternative_service_name(r);
+		else
+		    n = avahi_alternative_host_name(r);
+
                 avahi_free(r);
                 r = n;
 
@@ -89,7 +96,6 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     }
 
     avahi_alternative_service_name("\xc1\x0a");
-
     avahi_free(r);
     return 0;
 }
--- a/avahi-common/alternative.c
+++ b/avahi-common/alternative.c
@@ -21,6 +21,7 @@
 #include <config.h>
 #endif
 
+#include <limits.h>
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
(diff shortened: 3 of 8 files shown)


evverx (changes requested):
Thanks for the PR! I think it would be great to add a test case to let the CI test it under ASan/UBSan. Something like
```sh
busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server2 GetAlternativeServiceName "s" 'ab #2147483647'
```
should be enough to exercise that code path.

Other than that I don't think the function should return NULL there to prevent crashes like
```
Established under name 'ab #2147483647'                                                                                                                       
Name collision, picking new name '(null)'.                                                                                                                    
avahi-publish: entrygroup.c:422: avahi_entry_group_add_service_strlst: Assertion `name' failed.                                                               
Aborted (core dumped)
```
and loops like
```
Received conflicting record [ab\032\0352147483647._ssh._tcp.local       IN      SRV 0 0 22 host-3.local ; ttl=120] with local record to be. Withdrawing.      
Service "ab #2147483647" (/etc/avahi/services/ssh.service) successfully established.                                                                          
Service name conflict for "ab #2147483647" (/etc/avahi/services/ssh.service), retrying with "(null)".                                                         
...                                                                               
Received conflicting record [ab\032\0352147483647._ssh._tcp.local       IN      SRV 0 0 22 host-3.local ; ttl=120] with local record to be. Withdrawing.      
Service "ab #2147483647" (/etc/avahi/services/ssh.service) successfully established.                                                                          
Service name conflict for "ab #2147483647" (/etc/avahi/services/ssh.service), retrying with "(null)".    
...
```
when avahi runs into conflicts and chooses new names. In this particular case it should be fine to start counting from 0 again. I think `atoi` should be replaced and the counter should probably be unsigned in the first place.

pemensik:
But that function returns NULL on multiple places. Return value of null therefore must be handled anyway in all code paths. If it is not already, it must be fixed.

pemensik:
Prepared some tests and modified the approach a bit. If we reach INT_MAX, start again at 2. That might make avahi loop forever in case there is attacker present on the network and sending on every name we try to assign a conflicting response. I don't think we have better solution. Just log all attempts and keep trying.

Previous reaction in aborting would allow such attacker to succeed and create DoS. Yes, it would consume a lot of resources, but I think that approach is the best we can do.

pemensik:
Forgot to mention, expanded the same also to alternative hostname generation, where the problem is the same. It would be probably better to use strtoul() function instead of current code, but this should be handled enough.

evverx:
> But that function returns NULL on multiple places.

That's true but there those NULLs are usually caused by allocation failures and it's harder to trigger them intentionally in practice. Generally there are a lot of places where allocation failures aren't handled well unfortunately.

> Return value of null therefore must be handled anyway in all code paths. If it is not already, it must be fixed.

Agreed.

> If we reach INT_MAX, start again at 2

I'm still not sure it should be signed. It's true that the signed integer overflow is gone but the counter itself is semantically unsigned and with unsigned numbers `n++` is well defined without having to check `*_MAX` or anything else.

avahi-utils/avahi-publish.c
```diff
@@ -91,11 +91,15 @@ static void entry_group_callback(AvahiEntryGroup *g, AvahiEntryGroupState state,
...
-            register_stuff(config);
+	    if (n) {
+		fprintf(stderr, _("Name collision, picking new name '%s'.\n"), n);
+		avahi_free(config->name);
+		config->name = n;
+
+		register_stuff(config);
+	    } else
+		fprintf(stderr, _("Name collision, failed to pick a new name for '%s'.\n"),
+				config->name);
```
evverx, line 102: @pemensik could you move the last commit to a separate PR? It changes a few different things in a few different places and I think it would be better to review them separately (one by one probably). For example this change prevents avahi-publish from crashing (https://github.com/avahi/avahi/issues/478) but it doesn't fix the issue in the sense that `avahi-publish` still fails to pick the new name there.

I'm not sure what the test does either. It doesn't trigger the overflow. Could you drop it too? OSS-Fuzz should test it on x86_64 and i386 and I can add a test calling busctl to the smoke-test calling busctl later.

The first commit looks good (even though I think ideally  but I wonder if it should count from 1 instead of 2? I'm not sure where 2 came from. Could you also add "fixes https://github.com/avahi/avahi/issues/491" to the commit message?

evverx: I extracted the first commit, replaced tabs with spaces there and added a smoke test on top: https://github.com/avahi/avahi/pull/568. The other commits can be reviewed here later.

evverx:
I think now that the commit fixing the integer overflows is merged this PR can be closed. Could you open a separate PR with the avahi-daemon changes (and replace the tabs with spaces)?

The examples aren't fully fixed in the sense that they get stuck instead of calling `avahi_simple_poll_quit(simple_poll)`. The avahi-publish change prevents avahi-publish from crashing in https://github.com/avahi/avahi/issues/478 but the underlying bug where it doesn't extract the first label isn't fixed and it still fails when collisions occur. I think it's better to fail loudly for now.

PR #548, opened 2024-01-11
Use nonnull attribute on functions in lookup.h

pemensik:
Declare, which function parameters are expected to have always defined value. Some values can have NULL value, some do not. Issue compiler warnings on function, where they receive NULL but should not.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -30,16 +30,16 @@ jobs:
 - job: copr_build
   trigger: pull_request
   targets:
-  - fedora-all-aarch64
-  - fedora-all-i386
-  - fedora-all-ppc64le
-  - fedora-all-s390x
+  - fedora-latest-stable-aarch64
+  - fedora-latest-stable-i386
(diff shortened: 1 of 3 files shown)


pemensik:
This should be probably used on all functions in all headers, where any pointers are used.

evverx:
I'm on the fence about this. While I agree it's great to let compilers do their thing it should be enforced automatically somewhere because without that it's easy to miss those things. Currently the CI doesn't do that because avahi fails to compile with `-Werror`. I think it would be better to fix those warnings and flip `-Werrror` first (or at least get to the point where those particular warnings can be treated as errors).

pemensik:
oh hello packit! I have enabled staging packit by mistake. I do not think we want to use that.

evverx:
I think this PR can be closed. It should probably be revisited once avahi no longer fails to compile with `-Werror` but it's unlikely to happen anytime soon.

Issue #379, opened 2022-06-04
Wireshark reporting reflected mDNS responses as being malformed


imerin:
I'm trying to separate some IOT devices onto a VLAN.  I've setup avahi-daemon to reflect MDNS traffic between my main LAN and the new VLAN.   I have a terrible issue with resolving mDNS domain names with this setup, and using wireshark to inspect mdns traffic reveals the following:

Original packet:
`0000   01 00 5e 00 00 fb 48 e1 e9 32 ca c7 08 00 45 00   ..^...H..2....E.
0010   00 f9 e5 10 00 00 ff 11 e6 fe c0 a8 4d 40 e0 00   ............M@..
0020   00 fb 14 e9 14 e9 00 e5 2e 4b 00 00 84 00 00 00   .........K......
0030   00 01 00 00 00 05 04 5f 68 61 70 04 5f 74 63 70   ......._hap._tcp
0040   05 6c 6f 63 61 6c 00 00 0c 00 01 00 00 11 94 00   .local..........
0050   0e 0b 4d 53 47 31 30 30 2d 63 61 63 37 c0 0c c0   ..MSG100-cac7...
0060   27 00 10 80 01 00 00 11 94 00 4b 04 63 23 3d 32   '.........K.c#=2
0070   04 66 66 3d 32 14 69 64 3d 46 33 3a 42 45 3a 36   .ff=2.id=F3:BE:6
0080   37 3a 41 42 3a 31 30 3a 45 39 09 6d 64 3d 4d 53   7:AB:10:E9.md=MS
0090   47 31 30 30 06 70 76 3d 31 2e 31 04 73 23 3d 31   G100.pv=1.1.s#=1
00a0   04 73 66 3d 30 04 63 69 3d 34 0b 73 68 3d 52 64   .sf=0.ci=4.sh=Rd
00b0   6f 62 75 51 3d 3d c0 27 00 21 80 01 00 00 00 78   obuQ==.'.!.....x
00c0   00 0f 00 00 00 00 13 92 06 6d 74 37 36 38 37 c0   .........mt7687.
00d0   16 c0 9e 00 01 80 01 00 00 00 78 00 04 c0 a8 4d   ..........x....M
00e0   40 c0 27 00 2f 80 01 00 00 11 94 00 09 c0 27 00   @.'./.........'.
00f0   05 00 00 80 00 40 c0 9e 00 2f 80 01 00 00 00 78   .....@.../.....x
[...]

PR #552, opened 2024-01-27
Expand supported qtypes with some new types

pemensik:
Apple devices have been seen to use HTTPS RR queries even on mDNS. Define that record type to ensure it would be printed nicely in logs.

### DIFF
--- a/avahi-common/defs.h
+++ b/avahi-common/defs.h
@@ -324,7 +324,12 @@ typedef enum {
 
 /** @{ \name DNS RR definitions */
 
-/** DNS record types, see RFC 1035 */
+/** DNS record types, see RFC 1035.
+ *
+ * types registry:
+ * https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4
+ * Any new types should be added to avahi_dns_type_to_string function.
+ */
 enum {
     AVAHI_DNS_TYPE_A = 0x01,
     AVAHI_DNS_TYPE_NS = 0x02,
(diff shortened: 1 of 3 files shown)


avahi-common/defs.h
```diff
@@ -335,7 +340,15 @@ enum {
...
-    AVAHI_DNS_TYPE_SRV = 0x21
+    AVAHI_DNS_TYPE_SRV = 0x21,
+    AVAHI_DNS_TYPE_NAPTR = 0x23,
+    AVAHI_DNS_TYPE_CERT = 0x25,
+    AVAHI_DNS_TYPE_DNAME = 0x27,
+    AVAHI_DNS_TYPE_SSHFP = 0x2c,
+    AVAHI_DNS_TYPE_TLSA = 0x34,
+    AVAHI_DNS_TYPE_OPEPGPKEY = 0x3d,
+    AVAHI_DNS_TYPE_SVCB = 0x40,
+    AVAHI_DNS_TYPE_HTTPS = 0x41,
```
evverx, line 351: > Expand supported qtypes with some new types

> Apple devices have been seen to use HTTPS RR queries even on mDNS. Define that record type to ensure it would be printed nicely in logs.

As far as I can see HTTPS RRs aren't parsed anywhere so they are still dumped according to rfc3597. I'm not sure RRs that aren't fully supported should be included in this list. (At this point only fully supported RRs are listed here).

It could be I'm missing something though. Could you expand on what this PR is trying to improve?

pemensik: I disagree. These changes are enough to print these types without cryptic TYPE65 for known record types. Of course we want more complex decoding for all those types. Especially SVCB and HTTPS have great potential to be very common, so we will need them fixed later.

But that would require non-trivial code. These changes should make a bit better and very safe, without bringing any regressions. But yes, at least mentioning those types do not have avahi_record_to_string implementation in a comment would be nice.

This is how it is decoded in unknown records format:
```
$ echo "IN HTTPS 0 example.org " | named-rrchecker -u 
CLASS1	TYPE65	\# 15 0000076578616D706C65036F726700
```

This is how it should be formatted after my change
```
IN	HTTPS	\# 15 0000076578616D706C65036F726700
```

While this is suboptimal and not fully decoded, it is still much better with at least q-types converted.

evverx: I don't feel strongly about this but to me it just looks inconsistent to add unsupported types to the list of supported types.

I'm not sure how helpful it is either. For example NSEC RRs are used much more often in the context of mDNS and I haven't seen any feature requests asking to prettify them a bit. I saw bug reports like https://github.com/avahi/avahi/issues/379 though where it's necessary to actually parse NSEC RRs to reflect them properly.

evverx: Having taken a look at f6712902a92eb82b6c8d7e7fd0980a20a716fa0d I don't think this PR should be merged at all because the idea behind that commit was to format unknown types according to RFC3597 where
> an unknown RR type is
   represented by the word "TYPE" immediately followed by the decimal RR
   type number, with no intervening whitespace

, move all the supported types to `avahi-common/defs.h` and keep stuff like `AVAHI_DNS_TYPE_TKEY` separately.  It's a NACK from me.

pemensik: Oh, in that case we want to include NSEC to printed types as well. Of course reacting on its presence would be more difficult and not handled now.

pemensik: That is not list of supported types, but list of known types. That is somehow important difference.

evverx: > That is not list of supported types, but list of known types

In the context of avahi it's a list of supported types avahi knows how to parse and display properly. `AVAHI_DNS_TYPE_OPT` isn't in `avahi-common/defs.h` because avahi can't parse it. It is needed to do some things so it's known and kept in `avahi-common/rr.h`.

Other than that
> IN	HTTPS	\# 15 0000076578616D706C65036F726700

violates the RFC.

evverx:
I'm inclined to close this because there aren't a lot of places where RRs are logged. In the several places where they are logged those RRs are unlikely to show up because avahi doesn't support them. They can't be easily queried via, say, `avahi-resolve`. They can't be easily published so they are unlikely to trigger any conflicts either.

evverx:
Just to clarify as mentioned elsewhere I'm all for improving logs but it should be done systematically.  avahi should start logging that it receives packets, whether they are queries/responses or whatever and so on. Once those basics are covered it would probably make sense to prettify RR types.

PR #554, opened 2024-01-27
Expand logging of server name conflicts

pemensik:
Pass address into incoming_probe and use it in logs, when conflict arises. We do not have a good way now to report, with who the conflict happened. If we have lost the probing, include details in info log level.

Related: #117

### DIFF
--- a/avahi-core/server.c
+++ b/avahi-core/server.c
@@ -33,6 +33,7 @@
 #include <assert.h>
 #include <stdlib.h>
 
+#include <avahi-common/address.h>
 #include <avahi-common/domain.h>
 #include <avahi-common/timeval.h>
 #include <avahi-common/malloc.h>
@@ -193,13 +194,14 @@ static void withdraw_rrset(AvahiServer *s, AvahiKey *key) {
         withdraw_entry(s, e);
 }
 
-static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *i) {
+static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *i, const AvahiAddress *a) {
     AvahiEntry *e, *n;
     int ours = 0, won = 0, lost = 0;
 
     assert(s);
     assert(record);
     assert(i);
+    assert(a);
 
     /* Handle incoming probes and check if they conflict our own probes */
 
@@ -225,12 +227,18 @@ static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *
     }
 
     if (!ours) {
+	char address[AVAHI_ADDRESS_STR_MAX] = "(unknown)";
+        const char *iface = i->hardware ? i->hardware->name : "(unknown)";
         char *t = avahi_record_to_string(record);
 
+	avahi_address_snprint(address, sizeof(address), a);
+
         if (won)
-            avahi_log_debug("Received conflicting probe [%s]. Local host won.", t);
+            avahi_log_debug("Received conflicting probe [%s] from %s(%s). Local host won.",
+                            t, address, iface);
         else if (lost) {
-            avahi_log_debug("Received conflicting probe [%s]. Local host lost. Withdrawing.", t);
+            avahi_log_info("Received conflicting probe [%s] from %s(%s). Local host lost. Withdrawing.",
+                           t, address, iface);
             withdraw_rrset(s, record->key);
         }
 
@@ -238,13 +246,14 @@ static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *
     }
 }
 
-static int handle_conflict(AvahiServer *s, AvahiInterface *i, AvahiRecord *record, int unique) {
+static int handle_conflict(AvahiServer *s, AvahiInterface *i, AvahiRecord *record, const AvahiAddress *a, int unique) {
     int valid = 1, ours = 0, conflict = 0, withdraw_immediately = 0;
     AvahiEntry *e, *n, *conflicting_entry = NULL;
 
     assert(s);
     assert(i);
     assert(record);
+    assert(a);
 
     /* Check whether an incoming record conflicts with one of our own */
 
@@ -319,18 +328,21 @@ static int handle_conflict(AvahiServer *s, AvahiInterface *i, AvahiRecord *recor
     }
 
     if (!ours && conflict) {
-        char *t;
+	char address[AVAHI_ADDRESS_STR_MAX] = "(unknown)";
+        const char *iface = i->hardware ? i->hardware->name : "(unknown)";
+        char *t = avahi_record_to_string(record);;
 
         valid = 0;
-
-        t = avahi_record_to_string(record);
+	avahi_address_snprint(address, sizeof(address), a);
 
         if (withdraw_immediately) {
-            avahi_log_debug("Received conflicting record [%s] with local record to be. Withdrawing.", t);
+            avahi_log_info("Received conflicting record [%s] with local record to be from %s(%s). Withdrawing.",
+                           t, address, iface);
             withdraw_rrset(s, record->key);
         } else {
             assert(conflicting_entry);
-            avahi_log_debug("Received conflicting record [%s]. Resetting our record.", t);
+            avahi_log_info("Received conflicting record [%s] from %s(%s). Resetting our record.",
+                           t, address, iface);
             avahi_entry_return_to_initial_state(s, conflicting_entry, i);
 
             /* Local unique records are returned to probing
@@ -639,7 +651,7 @@ static void handle_query_packet(AvahiServer *s, AvahiDnsPacket *p, AvahiInterfac
             if (!avahi_key_is_pattern(record->key)) {
                 if (!from_local_iface)
                     reflect_probe(s, i, record);
-                incoming_probe(s, record, i);
+                incoming_probe(s, record, i, a);
             }
 
             avahi_record_unref(record);
@@ -713,7 +725,7 @@ static void handle_response_packet(AvahiServer *s, AvahiDnsPacket *p, AvahiInter
                 }
             }
 
-            if (handle_conflict(s, i, record, cache_flush)) {
+            if (handle_conflict(s, i, record, a, cache_flush)) {
                 if (!from_local_iface) {
                     if (!avahi_record_is_link_local_address(record))
                         reflect_response(s, i, record, cache_flush);


pemensik:
This would not resolve issue #117, but should help recording with who did the conflict arisen. I think that is important enough to be logged in info log level, not just debug level. If it causes change of our hostname, few log lines is justified for it.

avahi-core/server.c
```diff
@@ -319,18 +326,19 @@
...
+	avahi_address_snprint(address, sizeof(address), a);
 
         if (withdraw_immediately) {
-            avahi_log_debug("Received conflicting record [%s] with local record to be. Withdrawing.", t);
+            avahi_log_info("Received conflicting record [%s] with local record to be from %s(%s). Withdrawing.", t, address, iface);
             withdraw_rrset(s, record->key);
         } else {
             assert(conflicting_entry);
-            avahi_log_debug("Received conflicting record [%s]. Resetting our record.", t);
+            avahi_log_info("Received conflicting record [%s] from %s(%s). Resetting our record.", t);
```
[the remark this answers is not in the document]
pemensik: Ouch. missing parameters. I *love* those automatic check watching my back!

avahi-core/server.c
```diff
@@ -225,26 +227,33 @@ static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *
...
+	avahi_address_snprint(address, sizeof(address), a);
+
         if (won)
-            avahi_log_debug("Received conflicting probe [%s]. Local host won.", t);
+            avahi_log_debug("Received conflicting probe [%s] from %s(%s). Local host won.",
+                            t, address, iface);
         else if (lost) {
-            avahi_log_debug("Received conflicting probe [%s]. Local host lost. Withdrawing.", t);
+            avahi_log_info("Received conflicting probe [%s] from %s(%s). Local host lost. Withdrawing.",
+                           t, address, iface);
```
evverx, line 241: Without this patch avahi logs something like
```
Received conflicting record [C.local    IN      AAAA fe80::3430:b6ff:fe30:94d7 ; ttl=120] with local record to be. Withdrawing.
```
where the address can be seen so the "from" part seems to be redundant. I'm not sure about switching to "info" either. It would just flood the logs until that bug is fixed.

pemensik: That is something completely different address. From address is address of remote computer, which owns the name I am trying to register. Whereas conflicting record contains only my own address with name I am trying to register. That is not useful for debugging. I need information what machine won the registration of the proposed name.

Existing non-debug log shows just my new name, but no information with who I have clashed. I think such information is vital for analysis, why my machine conflicts with itself.

What I have used for testing is:
- Have other machine having avahi with a name, say ``test-a.local``. 
- My machine has name ``test-b.local``.
- call ``sudo avahi-set-host-name test-a`` on my machine and observe, what is logged.
- addresses in record and from should be different in this case.

evverx: > That is completely different address

I'm looking at
```
Received conflicting record [C.local    IN      AAAA fe80::3430:b6ff:fe30:94d7 ; ttl=120] with local record to be from fe80::3430:b6ff:fe30:94d7(ve-TEST-01). 
Withdrawing.
```
and the from address is the same here and it shows the address of the remote machine.

pemensik: What I get is:

```
Received conflicting record [menryzen.local	IN	AAAA 2a03:3b40:296:0:9b07:171c:858a:ffbb ; ttl=60] with local record to be from fe80::be0d:7fb0:1317:9d59(enp0s31f6). Withdrawing.

$ avahi-resolve -a fe80::be0d:7fb0:1317:9d59 
fe80::be0d:7fb0:1317:9d59	menryzen.local
$ avahi-resolve -a 2a03:3b40:296:0:9b07:171c:858a:ffbb
2a03:3b40:296:0:9b07:171c:858a:ffbb	menryzen.local
```

Ah, I did not expect both of those addresses are of the remote machine. They are just different in my case. Okay, you are correct that the another address is not useful. Interface information could be. Since it prints into info severity lines like ``Withdrawing address record for ::1 on lo.`` and ``Registering new address record for ::1 on lo.*.`` after name change, I think more information about conflict should be logged in non-debug environment as well.

pemensik: From address should be the actual source address of a packet, from which packet were received on socket level. I think the record printed contains addresses from that host, which might conflict. Not sure, they could be potentially different. Especially in some reflection scenarios.

evverx: > Registering new address record for ::1 on lo.*.

I think in the context of those messages interface names are helpful because they show what avahi registers and where. In the context of conflicting records I'm not sure interfaces matter that much probably.

> Since it prints into info severity lines like

If it logged that with the debug info there would be no way to figure out what's happening by just looking at the logs so the info level is more or less justified. With name conflicts it's different.

Generally I'm on the fence about this. If I get to the point where I have to debug name conflicts I can probably turn on the debug level. It should be possible to run Wireshark too. I can see where messages like
```
Host name conflict with some machine, retrying with C
```
can be helpful but I think the "conflicting record" messages are detailed enough probably.

pemensik: I am not so sure. I have not seen any checks on the incoming message, that at least on of addresses in incoming DNS packet belongs to the sender of that packet. We may conflict with some proxy, but we need to know address of that proxy in this case. Not addresses inside the message, which might be actually ours.

Unfortunately not enough details are passed to ``Host name conflicts`` message in *server_callback* in avahi/daemon/main.c. It would take a lot of parameters adding. So I have taken a bit shortcut by logging those information earlier, when we still have the information.

By the way, logging of conflicts are quite inconsistent:

```
$ grep 'Host name conflict' -r avahi*
avahi-core/avahi-test.c:        avahi_log_debug("Host name conflict, retrying with <%s>", n);
avahi-core/conformance-test.c:        avahi_log_warn("Host name conflict, retrying with <%s>", n);
avahi-daemon/static-hosts.c:            avahi_log_error("Host name conflict for \"%s\", not established.", h->host);
avahi-daemon/main.c:            avahi_log_warn("Host name conflict, retrying with %s", n);
avahi-daemon/main.c:            sd_notifyf(0, "STATUS=Host name conflict, retrying with %s", n);
avahi-utils/avahi-publish.c:                fprintf(stderr, _("Host name conflict\n"));
```

Notice always a different importance is used. Perhaps we should unify it in different tools?

evverx: > I have not seen any checks on the incoming message, that at least one of addresses in incoming DNS packet belongs to the sender of that packet. We may conflict with some mdns proxy, but we need to know address of that proxy in this case

I still think that if I get to this point I have to use debuggers, run Wrieshark and so on. I can't see how logs can help with anything.

> Notice always a different importance is used. Perhaps we should unify it in different tools?

Those messages are logged in different contexts. The tests don't matter generally. In `static-hosts` it's "error" because avahi doesn't keep retrying with new names as far as I can remember. In `main.c` it's "warning" because conflicts don't prevent avahi from getting another name. `avahi-publish` uses the "verbose" switch so log levels don't matter there and it just crashes when conflicts occur https://github.com/avahi/avahi/issues/478. It would be nice to make logs consistent in general of course but it all depends on the context so switching everything to "info" isn't ideal either.

evverx:
> We do not have a good way now to report

With `--debug` conflicting records are shown.

If you need a reliable way to reproduce the underlying issue you can send two conflicting responses following each other to avahi and watch it violate the following part of the RFC:
> conflicting Multicast
   DNS responses received *before* the first probe packet is sent MUST
   be silently ignored

which is supposed to take care of stale stuff reflected back locally or remotely.
```
    1   0.000000 192.168.17.209 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.17.216
    2   0.101874 192.168.17.209 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.17.216
    3   0.316334 192.168.17.217 → 224.0.0.251  MDNS 132 Standard query 0x0000 ANY 217.17.168.192.in-addr.arpa, "QM" question ANY C-2.local, "QM" question A 192.168.17.217 PTR C-2.local
    4   0.567970 192.168.17.217 → 224.0.0.251  MDNS 132 Standard query 0x0000 ANY 217.17.168.192.in-addr.arpa, "QM" question ANY C-2.local, "QM" question A 192.168.17.217 PTR C-2.local
    5   0.819613 192.168.17.217 → 224.0.0.251  MDNS 132 Standard query 0x0000 ANY 217.17.168.192.in-addr.arpa, "QM" question ANY C-2.local, "QM" question A 192.168.17.217 PTR C-2.local
    6   1.021152 192.168.17.217 → 224.0.0.251  MDNS 120 Standard query response 0x0000 PTR, cache flush C-2.local A, cache flush 192.168.17.217
    7   2.185930 192.168.17.217 → 224.0.0.251  MDNS 120 Standard query response 0x0000 PTR, cache flush C-2.local A, cache flush 192.168.17.217
    8   4.349981 192.168.17.217 → 224.0.0.251  MDNS 120 Standard query response 0x0000 PTR, cache flush C-2.local A, cache flush 192.168.17.217
```
The first packet switches avahi to probing state and the second packet triggers the conflict (even though it MUST be ignored). Instead of the same second response you can send a probe with an IP that is lexicographically later. It would more or less emulate probes coming back locally with a short delay.

Here's what should happen instead:
```sh
    1   0.000000 192.168.229.225 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.229.237
    2   0.103413 192.168.229.225 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.229.237
    3   0.269525 192.168.229.238 → 224.0.0.251  MDNS 83 Standard query 0x0000 ANY C.local, "QM" question A 192.168.229.238
    4   0.270026 192.168.229.238 → 224.0.0.251  MDNS 89 Standard query response 0x0000 AAAA, cache flush fe80::3430:b6ff:fe30:94d7
    5   0.519027 192.168.229.238 → 224.0.0.251  MDNS 83 Standard query 0x0000 ANY C.local, "QM" question A 192.168.229.238
    6   0.769547 192.168.229.238 → 224.0.0.251  MDNS 83 Standard query 0x0000 ANY C.local, "QM" question A 192.168.229.238
    7   0.969737 192.168.229.238 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.229.238
    8   2.187305 192.168.229.238 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.229.238
    9   4.405110 192.168.229.238 → 224.0.0.251  MDNS 77 Standard query response 0x0000 A, cache flush 192.168.229.238
```
The second packet is ignored and avahi defends itself.

I'll go ahead and close this.

pemensik:
I understand we have quite limitations in our current code. But this logging is important in a case of reflector involved, such as https://github.com/vfreex/mdns-reflector. That publishes services on behalf other services. You would never see its address in logs, because that is not what is advertised. Printing address from which the message were received would allow obtaning also MAC address of conflicting device.

But okay, let it closed for now. I think we want ability to print all records from the conflicting set, at least all with conflicting name. Sender address seems useful in any case. But proper style repeat won't be implemented before 0.9.1, it needs more testing.

PR #564, opened 2024-02-10
Adds TTL Ignore Option to Avahi Cache for Enhanced Compatibility

anoxdetox:
In diverse network setups, especially those leveraging Avahi's reflector capability, there can be discrepancies in how TTL values, service advertisements, and queries are managed across different devices and clients. These discrepancies can lead to services being prematurely dropped from the cache, disrupting connectivity and discovery.

The introduction of the ignore-ttl configuration option serves as a workaround for these challenges, accommodating environments where TTL mishandling, be it client-side software or devices on other network segments, may be in cause. It allows for a more robust discovery process by temporarily extending the visibility of services that might otherwise be considered offline.


evverx:
Thank you for the PR.

It could be I misunderstood what `ignore-ttl` does but if the idea is to avoid keeping resource records in the reflector cache it seems to me that `cache-entries-max=0` should work here. Could you expand on what the difference is?

anoxdetox:
Thank you for your insightful remarks.

It seems like `cache-entries-max=0` behaves the same, at least in our reflector setup, as the proposed `ignore-ttl` proposal.

Our intial assumption that an early cache pruning might be better, than a cache block, based on a somewhat old, Stack Overflow article: https://stackoverflow.com/questions/18388334/avahi-hostname-resolution-is-it-caching-somewhere .

Then again, since we are not that familiar on the code base, we did not want to induce unforseen consequences with the submittal.

Then again `
cache-entries-max= Takes an unsigned integer specifying how many resource records are cached per interface. Bigger values allow mDNS work correctly in large LANs but also increase memory consumption.
` is somewhat vague and could be improved.

Cheers!

evverx:
> "However, I should note here that this will also render avahi pretty much useless for normal use. While avahi-daemon will be issuing lookup packets, it will be unable to store the results and every call of avahi-resolve-host-name (as well as other command-line tools, nss-mdns, D-Bus API…) will fail."

That answer is applicable to devices that don't reflect mDNS packets. `avahi-resolve` indeed times out there. It still times out with this PR applied though. Generally stale cache entries aren't handled well and https://github.com/avahi/avahi/pull/229 should help to address that. As far as I can remember it doesn't cover all the use cases but it's based on the RFC and can probably be extended to make it possible to run something like `avahi-resolve --cache=no -n some-name` to bypass the cache altogether (that's what the question on Stack Overflow was about as far as I can tell)

All in all I think that `cache-entries-max=0` already covers this particular reflector use case. I agree it can be documented better but I have to admit I'm not sure how to do that.

PR #565, opened 2024-02-10
smoke-test: cover sync address/hostname/service resolvers

evverx:
record browser, domain browser, dns server browser and some other things.

https://coveralls.io/builds/65623826


pemensik:
I would like such tests outside of .github directory. Those are useful and eventually should be run in a regressions suite. Something, which could be run in distributions on builds and updates. Eventually checking issues with unusual build architecture or something similar. It would be great, if they could be run with ``make check`` or similar target from main repository. *.github* should be just a thin wrapper for existing tests to run them on PR.

PR #17, opened 2015-10-13
Fixed dns_sd segfaults, initialization issues, and added NDEBUGs

tpenguin:
- Corrected a segfault in the code on shutdown.
- Fixed memory initialization error which caused problems with
  shutdown; see above.
- Added NDEBUG calls around debugging fprintfs.

PR #569, opened 2024-02-19
libdns_sd: Use allocator with memory set to zero

nkarstens:
Fixes segfault on shutdown. On client failure, generic_client_callback function simply checks if all the callback functions are non-zero and triggers the callback. This change sets the callback function memory to NULL and prevents the application from thinking that a callback has been installed.

### DIFF
--- a/avahi-compat-libdns_sd/compat.c
+++ b/avahi-compat-libdns_sd/compat.c
@@ -366,7 +366,7 @@ static DNSServiceRef sdref_new(void) {
     if (socketpair(AF_UNIX, SOCK_STREAM, 0, fd) < 0)
         goto fail;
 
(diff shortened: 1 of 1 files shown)


evverx:
@nkarstens thank you for the PR! As far as I understand it fixes one of the issues mentioned in https://github.com/avahi/avahi/pull/17. I wonder if it would be possible to add a test for that?

(On an unrelated note I remember about the reconfirm PR. I'm planning to get back to it soon. The stale cache entries came up recently again)

evverx:
Looks like the compat-libdns-sd tests are in bad shape in general and don't cover much. Other than that they don't seem to be run when avahi is up and running and it needs fixing too.

I'll go ahead and merge this PR.

nkarstens:
Thanks, @evverx!

PR #572, opened 2024-02-27
avahi-core/entry.c: Allow localhost as valid hostname

zdohnal:
Several daemons (PAPPL based printer applications, ipp-usb) register/can register services only locally to prevent remote access to the service due security reasons (using mDNS gives them unified API for autodiscovery, local machine and network), so it would be great if avahi supported localhost as valid hostname.

Currently, if daemon wants to register service locally, it cannot pass localhost as registering hostname - Avahi returns it as invalid hostname until the machine's hostname is renamed to localhost (which is sometimes unwanted). The fix is to allow localhost in the check `AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL` when checking hostname.

### DIFF
--- a/avahi-core/entry.c
+++ b/avahi-core/entry.c
@@ -598,7 +598,7 @@ static int server_add_service_strlst_nocopy(
     AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL(s, avahi_is_valid_service_name(name), AVAHI_ERR_INVALID_SERVICE_NAME);
     AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL(s, avahi_is_valid_service_type_strict(type), AVAHI_ERR_INVALID_SERVICE_TYPE);
     AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL(s, !domain || avahi_is_valid_domain_name(domain), AVAHI_ERR_INVALID_DOMAIN_NAME);
-    AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL(s, !host || avahi_is_valid_fqdn(host), AVAHI_ERR_INVALID_HOST_NAME);
+    AVAHI_CHECK_VALIDITY_SET_RET_GOTO_FAIL(s, !host || avahi_is_valid_fqdn(host) || !strcmp(host, "localhost"), AVAHI_ERR_INVALID_HOST_NAME);
 
     if (!domain)
         domain = s->domain_name;


pemensik:
I think the first problem with this approach is localhost is compared in case-sensitive way. Hostnames should be compared case-insensitive way.

evverx:
> if daemon wants to register service locally, it cannot pass localhost as registering hostname

I'm not sure why it would need "localhost". It can pass NULL there to let avahi pick the current host name and also pass the loopback interface explicitly when publishing (and when resolving it can pass the loopback interface too to resolve that host name to 127.0.0.1). Could you expand on why "localhost" is required there?

evverx (changes requested):
It would be great to figure out why "localhost" is needed when the loopback interface with a `NULL` hostname should be enough but other than that this PR doesn't make "localhost" work because services published like that fail to resolve locally (unless `AVAHI_LOOKUP_NO_ADDRESS` is specified explicitly). With `wide-area` enabled attempts to resolve those services trigger unicast queries like
```sh
07:23:10.775856 IP H.38134 > 10.0.2.3.domain: 46416+ A? localhost. (27)
07:23:10.780747 IP 10.0.2.3.domain > H.38134: 46416 NXDomain 0/1/0 (102)
```
(because "localhost" doesn't end in `.local`). With wide-area disabled multicast queries are sent and then lookups time out if there are no replies from other machines. "localhost" probably shouldn't be allowed on any other interfaces either in the sense that for example `iface=AVAHI_IF_UNSPEC` combined with `host="localhost"` should fail to prevent other machines from seeing things like
```
23:15:16.089  T._qotd._tcp.local. can be reached at localhost.:123 (interface 13)
```
Direct unicast queries and probably some other stuff should support "localhost" too. All in all it should be a patch touching quite a few different places. Then again it would be great to figure out what the use case is.

evverx:
Looking at https://github.com/avahi/avahi/issues/125#issuecomment-302907801 I think I more or less get where "localhost" comes from and it seems Trent was opposed to the idea.

Anyway it certainly can't be addressed with one-line patches like this so I'll go ahead and close this. Could you open a separate issue with all the details?

zdohnal:
Thank you for the review and comments! Tbh passing NULL as hostname with 'lo' interface index occurred to me as well, but I wanted to simplify PAPPL code if possible (PAPPL has mDNSResponder support as well, so I looked for something which would look/behave similar way, so code maintenance difficulty won't get higher - mDNSResponder looks to be ok with passing localhost as hostname), and this draft fix did the trick for me - the service was resolved correctly during printing.
But I will try to suggest hostname NULL as well to PAPPL upstream - the security issue mentioned by Mike in the #125 looks to be fixed by using dnssd name as device URI, so this might be okay. Although I'm not sure about non-local address - the daemon might act as a server and it might be unwanted to let hostname change based on the current network situation. (However it prevents registering a service on hostname it does not belong to the issuer).

So I will be back if we don't solve this other way or somethings else is needed regarding this matter, thanks!

evverx:
> the service was resolved correctly during printing

As far as I understand it works because avahi is used for browsing only there and when hostnames are received they go via the system resolver where `localhost` is resolved. When the following service pointing to localhost is published on the loopback interface:
```sh
dig  @127.0.0.1 -p5353 local-stuff._qotd._tcp.local SRV +noall +answer
local-stuff._qotd._tcp.local. 10 IN     SRV     0 0 123 localhost.
```

`avahi-browse` can browse it:
```sh
avahi-browse -pt _qotd._tcp
+;lo;IPv4;local-stuff;_qotd._tcp;local
```
but  it all breaks once `-r` is passed to fully resolve it using `avahi`:
```sh
avahi-browse -rpt _qotd._tcp
+;lo;IPv4;local-stuff;_qotd._tcp;local
Failed to resolve service 'local-stuff' of type '_qotd._tcp' in domain 'local': DNS failure: NXDOMAIN
```

Anyway my understanding is that the localhost situation was discussed a lot (outside of https://github.com/avahi/avahi/issues/125 too) and it would probably be better to talk to the participants from that issue to figure out what the consensus was at that point and whether is has changed since then.

PR #577, opened 2024-03-04
avahi-daemon.conf: turn off wide-area

evverx:
In its current form it doesn't interact with mDNS well. For example services advertised over mDNS where PTR RRs point "_services._dns-sd._udp.local" to unresolvable wide-area domain names bring down mDNS browsers. DNS isn't supported well either. For example responses larger than 512 bytes aren't supported, all the queries come from the same UDP port and so on. All in all in its current form it isn't something that should be enabled by default or picked up accidentally by packages downstream.

It reverts the part of 2c453196ee040e17e357f3431b0647391c88d616 where that feature was turned on.

### DIFF
--- a/avahi-daemon/avahi-daemon.conf
+++ b/avahi-daemon/avahi-daemon.conf
@@ -39,7 +39,7 @@ ratelimit-interval-usec=1000000
 ratelimit-burst=1000
 
 [wide-area]
-enable-wide-area=yes
+#enable-wide-area=no
 
 [publish]
 #disable-publishing=no


pemensik:
I think primary fix here should be complete split of mDNS and wide area queries. They have differences between them, they need also slightly different behaviour. And I doubt we ever want to follow unicast request with multicast request or vice versa. The only exception should be gateway running in a special mode.

But following CNAME from one to another should be prevented, at least for privacy reasons.

PR #587, opened 2024-03-28
Report fatal error when UI file were not loaded

pemensik:
Include useful message printing the reason of the failure. Useful especially when the file is not in correct path.

This were rejected in #408, but I think this part alone should be merged.

### DIFF
--- a/avahi-discover-standalone/main.c
+++ b/avahi-discover-standalone/main.c
@@ -322,6 +322,7 @@ int main(int argc, char *argv[]) {
     GtkTreeViewColumn *c;
     gint error;
     AvahiGLibPoll *poll_api;
+    GError *gerror = NULL;
 
     gtk_init(&argc, &argv);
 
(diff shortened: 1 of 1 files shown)


evverx:
> when the file is not in correct path

As far as I understand it can happen if the build system is messed up somewhere. Should it be fixed too?

> This were rejected in https://github.com/avahi/avahi/pull/408

 I focused on service names there and I don't think I noticed that commit.

pemensik:
This happened in Fedora package, where the UI file existed, but in wrong subpackage. I had to use debugger to find why nothing were showing. So I am not sure it is built wrong way, but it should be reported if something is missing.

But sure, I included it with unrelated commit, because I found it testing changes in #408. Still this is useful.

pemensik:
Damn. It seems I have broken already working packit integration.

evverx:
> This happened in Fedora package, where the UI file existed, but in wrong subpackage

Assuming it was a packaging issue where the files were installed in the wrong order or something like that I guess the error message should be enough then.

Could you replace the tabs with spaces? Even though it isn't mentioned anywhere explicitly tabs aren't used.

pemensik:
Ah, right! I have to find a nice way to specify editor defaults per project, because common defaults varies across different project I work on. We should try to describe preferred code style anyway.

pemensik:
Related Fedora commit: https://src.fedoraproject.org/rpms/avahi/c/f29dc51363501a541aa4ec6202a45d7e067a5590

evverx:
> We should try to describe preferred code style anyway

It's kind of described at https://web.archive.org/web/20160504114856/http://avahi.org/wiki/CodingStyle (which points to https://web.archive.org/web/20150920162617/http://freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/CodingStyle/). It should probably be revived one way or another.

Looking at https://copr.fedorainfracloud.org/coprs/packit/avahi-avahi-587/build/7231661/ I wonder if the Packit folks hooked it up to the repository?

PR #588, opened 2024-03-28
Write first attempt to security issues reporting

pemensik:
@evverx What do you think? Should be something else mentioned?

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,27 @@
+# Security Policy
+
+## Supported Versions
+
+Security updates are only provided for last released version.
+Previous versions are not fixed.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 0.8.x   | :white_check_mark: |
+| < 0.8.0 | :x:                |
+
+## Reporting a Vulnerability
+
+If any can prevent avahi-daemon from responding or even terminating it
+Use this section to tell people how to report a vulnerability.
+If you discover a way to crash avahi-daemon via network packet or make it non-responsive,
+please do not create normal public issue. Instead report it on
+[Security](https://github.com/avahi/avahi/security) page as a vulnerability.
+Give us enough time to look into it, please do not post it on public places.
+Including our own issue tracker or mailing list.
+If it does not crash but stops responding in reaction to remote event, report it the same way.
+
+If any tool crashes in reaction to network traffic, consider reporting it the same way.
+
+Crash caused by wrong configuration or when it requires local administrator privileges should be reported as
+an ordinary issue only.


SECURITY.md
```diff
@@ -0,0 +1,27 @@
+# Security Policy
+
+## Supported Versions
+
+Security updates are only provided for last released version.
```
evverx, line 5: I think it kind of implies that one "stable" branch is maintained here upstream and that's not accurate. Fixes go to the master branch directly. Backports to released versions (including the latest one) are handled downstream. I think this paragraph can be removed.

SECURITY.md
```diff
@@ -0,0 +1,27 @@
...
+| < 0.8.0 | :x:                |
+
+## Reporting a Vulnerability
+
+If any can prevent avahi-daemon from responding or even terminating it
+Use this section to tell people how to report a vulnerability.
+If you discover a way to crash avahi-daemon via network packet or make it non-responsive,
+please do not create normal public issue. Instead report it on
+[Security](https://github.com/avahi/avahi/security) page as a vulnerability.
+Give us enough time to look into it, please do not post it on public places.
```
evverx, line 20: I think it's too vague. It should say that the project follows 90-day disclosure policy (or whatever) and that response should be expected in a reasonable amount of time (say, a week or two). Without that it looks like it's possible to sit on bugs indefinitely. I think it should be possible to copy-paste https://github.com/google/oss-vulnerability-guide/blob/main/templates/security_policies/github_security_policy.md and adjust it a bit.

SECURITY.md
```diff
@@ -0,0 +1,27 @@
...
+| 0.8.x   | :white_check_mark: |
+| < 0.8.0 | :x:                |
+
+## Reporting a Vulnerability
+
+If any can prevent avahi-daemon from responding or even terminating it
+Use this section to tell people how to report a vulnerability.
+If you discover a way to crash avahi-daemon via network packet or make it non-responsive,
+please do not create normal public issue. Instead report it on
+[Security](https://github.com/avahi/avahi/security) page as a vulnerability.
```
evverx, line 19: There should probably be a way to report things without having a GitHub account.

evverx (changes requested):
I think it would be.better to copy-paste https://github.com/google/oss-vulnerability-guide/blob/main/templates/security_policies/github_security_policy.md  and adjust it a bit.

evverx:
I think it should also be mentioned explicitly that there are no bug bounties, rewards and so on just in case.

evverx:
I think this PR can be closed. I'll try to add `SECURITY.md` pointing to https://github.com/avahi/avahi/security/advisories/new (along with `CODING_STYLE.md` pointing to https://web.archive.org/web/20160504114856/http://avahi.org/wiki/CodingStyle and https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/CodingStyle/) later.

PR #593, opened 2024-04-28
avahi-discover: Fix invalid escape sequences

C0rn3j:
Corrects 
```python
/usr/lib/python3.12/site-packages/avahi/ServiceTypeDatabase.py:98: SyntaxWarning: invalid escape sequence '\.'
  if not re.search('_[a-zA-Z0-9-]+\._[a-zA-Z0-9-]+', key):
```

### DIFF
--- a/avahi-python/avahi/ServiceTypeDatabase.py.in
+++ b/avahi-python/avahi/ServiceTypeDatabase.py.in
@@ -95,9 +95,9 @@ class ServiceTypeDatabase:
     def __iter__(self):
 
         def want_key(key):
-            if not re.search('_[a-zA-Z0-9-]+\\._[a-zA-Z0-9-]+', key):
+            if not re.search(r'_[a-zA-Z0-9-]+\._[a-zA-Z0-9-]+', key):
                 return False
-            if re.search('_[a-zA-Z0-9-]+\\._[a-zA-Z0-9-]+\\[.*\\]', key):
+            if re.search(r'_[a-zA-Z0-9-]+\._[a-zA-Z0-9-]+\[.*\]', key):
                 return False
             return True
 
--- a/avahi-python/avahi/ServiceTypeDatabase.py.in
+++ b/avahi-python/avahi/ServiceTypeDatabase.py.in
@@ -95,9 +95,9 @@ class ServiceTypeDatabase:
     def __iter__(self):
 
         def want_key(key):
-            if not re.search('_[a-zA-Z0-9-]+\._[a-zA-Z0-9-]+', key):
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR!

I think it would be better to use raw strings instead (to me `r'\.'` is a bit easier to read than `'\\.'`)

evverx (commented):
It's unrelated to this PR but looks like this is completely broken in the sense that I added a couple of tests to the smoke test to cover the Python code a bit and with the "invalid escape sequence" warnings gone the test running
```sh
python3 -Werror -Xdev ServiceTypeDatabase.py
```
fails with
```sh
  File "/home/runner/work/avahi/avahi/avahi-python/avahi/ServiceTypeDatabase.py", line 156, in <module>
    print (b.items())
  File "/home/runner/work/avahi/avahi/avahi-python/avahi/ServiceTypeDatabase.py", line 74, in items
    return list(self.iteritems())
  File "/home/runner/work/avahi/avahi/avahi-python/avahi/ServiceTypeDatabase.py", line 135, in <genexpr>
    return ((key, self[key]) for key in self)
  File "/home/runner/work/avahi/avahi/avahi-python/avahi/ServiceTypeDatabase.py", line 112, in __iter__
    if want_key(key):
  File "/home/runner/work/avahi/avahi/avahi-python/avahi/ServiceTypeDatabase.py", line 98, in want_key
    if not re.search('_[a-zA-Z0-9-]+\\._[a-zA-Z0-9-]+', key):
  File "/usr/lib/python3.10/re.py", line 200, in search
    return _compile(pattern, flags).search(string)
TypeError: cannot use a string pattern on a bytes-like object
```
Looks like it needs revisiting.

C0rn3j:
Might be a good time to start using Python's type hints (3.5+) if this is going to be revisited.

I have switched the PR to raw strings.

evverx:
> Python's type hints (3.5+)

I think it would be nice to at least get to the point where the tests were actually run and various basic linters didn't produce a lot of warnings first.

Just out of curiosity I wonder where the warnings originally came from? I'm guessing It's likely to be broken too unless it just imports the module and looks for service names probably.

C0rn3j:
Just running `avahi-discover` makes them pop up immediately on Arch Linux which just migrated to Python 3.12, which I presume is harsher than previous versions.

evverx:
Got it. Thanks! I opened https://github.com/avahi/avahi/issues/596 and https://github.com/avahi/avahi/issues/594 to prevent it from getting lost. I'll go ahead and merge this PR.

PR #599, opened 2024-05-06
Add dbus-1 to avahi-client Requires.private

alyssais:
This fixes static linking of avahi-client using pkg-config, by telling pkg-config to also link with dbus, since static libraries don't carry their own dependency information.


evverx:
The libs emitted by `pkg-config --static` make much more sense now. Thank you!

PR #602, opened 2024-05-19
Initial Hindi translation

Scrambled777:
Adds initial Hindi `hi` language translation.

Cheers.

### DIFF
--- a/po/hi.po
+++ b/po/hi.po
@@ -6,8 +6,8 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: avahi master\n"
-"Report-Msgid-Bugs-To: https://github.com/lathiat/avahi/issues\n"
+"Project-Id-Version: Avahi master\n"
+"Report-Msgid-Bugs-To: https://github.com/avahi/avahi/issues\n"
 "POT-Creation-Date: 2024-05-18 15:24+0000\n"
 "PO-Revision-Date: 2024-05-19 14:21+0530\n"
 "Last-Translator: Scrambled777 <weblate.scrambled777@simplelogin.com>\n"
--- a/po/hi.po
+++ b/po/hi.po
@@ -1,0 +1,864 @@
+# Hindi translation for avahi.
+# Copyright (C) 2024 avahi's COPYRIGHT HOLDER
+# This file is distributed under the same license as the avahi package.
+# Scrambled 777 <weblate.scrambled777@simplelogin.com>, 2024.
+# Scrambled777 <weblate.scrambled777@simplelogin.com>, 2024.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: avahi master\n"
+"Report-Msgid-Bugs-To: https://github.com/lathiat/avahi/issues\n"
+"POT-Creation-Date: 2024-05-18 15:24+0000\n"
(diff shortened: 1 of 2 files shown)


po/hi.po
```diff
@@ -0,0 +1,864 @@
+# Hindi translation for avahi.
+# Copyright (C) 2024 avahi's COPYRIGHT HOLDER
+# This file is distributed under the same license as the avahi package.
+# Scrambled 777 <weblate.scrambled777@simplelogin.com>, 2024.
+# Scrambled777 <weblate.scrambled777@simplelogin.com>, 2024.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: avahi master\n"
```
evverx, line 9: Could you change it to `Avahi` to make it consistent with the other `po` files?

po/hi.po
```diff
@@ -0,0 +1,864 @@
...
+"Report-Msgid-Bugs-To: https://github.com/lathiat/avahi/issues\n"
```
evverx, line 10: Could you change it to https://github.com/avahi/avahi/issues? avahi moved to the avahi organization recently.

evverx (changes requested):
Thank you for the PR!

Looking at the email address I wonder if it would be more convenient if avahi was integrated with Weblate? This way it would be possible to automatically pull changes from there instead of opening PRs. There was a PR where it came up but it didn't go anywhere.

(I'm not a native speaker so I'd need a bit more time to delve into the translation itself)

po/hi.po
Scrambled777: `"Project-Id-Version: Avahi master\n"`

Like this?

po/hi.po
Scrambled777: Sure. Will update `po` file now. Thanks.

po/hi.po
evverx: `"Project-Id-Version: Avahi\n"`

I should have suggested a change instead of that ambiguous comment. Sorry!

Scrambled777: > I should have suggested a change instead of that ambiguous comment. Sorry!

No worries.

Amended `po` file. 

Anyway, I got pot file from [here](https://l10n.gnome.org/vertimus/avahi/master/po/hi/).
Is it possible to have updated POT file with the modifications you mentioned? It will avoid further mistakes in future.

Scrambled777:
> Looking at the email address I wonder if it would be more convenient if avahi was integrated with Weblate? This way it would be possible to automatically pull changes from there instead of opening PRs. There was a PR where it came up but it didn't go anywhere.

Weblate integration would be awesome.

evverx:
> Weblate integration would be awesome.

Got it. I opened https://github.com/avahi/avahi/issues/603 so as not to forget about it.

Scrambled777:
@evverx I have already updated `po` file as per your findings. 

Let me know if something else need to be amended. Cheers.

po/hi.po
evverx: > Is it possible to have updated POT file with the modifications you mentioned?

I'm not sure who has access to https://l10n.gnome.org/vertimus/avahi/master/po/hi/ and how it gets updated so I'm afraid I can't help here.

evverx (approved):
LGTM. Thank you!

Issue #315, opened 2020-12-22
avahi fails if avahi-daemon.conf file has any line that exceeds 255 characters


dkerr64:
avahi daemon fails with an error if any line in the conf file exceeds 255 characters, including comment lines.  I traced the problem to [ini-file-parser.c](https://github.com/lathiat/avahi/blob/d1e71b320d96d0f213ecb0885c8313039a09f693/avahi-daemon/ini-file-parser.c#L53) where 256 bytes is allocated on the stack to hold each line.  As a quick fix I bumped it up to 1024 and that solved my problem.

The reason I exceeded 255 characters is because I use `reflect-filters` and with that it is quite easy to have a long line length if you have several different types of devices you want to reflect across LAN subnets.

Thanks
DAK.

abelbeck:
Nice find!  Your solution looks good for a minimal impact fix.

PR #609, opened 2024-07-06
Increase length limit on input lines.

dennypage:
This is particularly important for reflect-filters, which can be lengthy in real-world use.

### DIFF
--- a/avahi-core/log.c
+++ b/avahi-core/log.c
@@ -33,7 +33,7 @@ void avahi_set_log_function(AvahiLogFunction function) {
 }
 
 void avahi_log_ap(AvahiLogLevel level, const char*format, va_list ap) {
-    char txt[256];
+    char txt[1280];
 
     vsnprintf(txt, sizeof(txt), format, ap);
 
--- a/avahi-core/log.c
+++ b/avahi-core/log.c
@@ -33,7 +33,7 @@ void avahi_set_log_function(AvahiLogFunction function) {
 }
 
 void avahi_log_ap(AvahiLogLevel level, const char*format, va_list ap) {
-    char txt[1280];
+    char txt[256];
 
     vsnprintf(txt, sizeof(txt), format, ap);
 
--- a/avahi-daemon/ini-file-parser.c
+++ b/avahi-daemon/ini-file-parser.c
@@ -50,7 +50,7 @@ AvahiIniFile* avahi_ini_file_load(const char *fname) {
 
     line = 0;
     while (!feof(fo)) {
-        char ln[256], *s, *e;
+        char ln[1024], *s, *e;
         AvahiIniFilePair *pair;
 
         if (!(fgets(ln, sizeof(ln), fo)))


evverx:
Thank you for the PR!

It was reported in https://github.com/avahi/avahi/issues/315 and as far as I can see there it was bumped to 1024 as well so it seems 1024 should be good enough. That being said it isn't the only place where 256 is hard-coded (https://github.com/avahi/avahi/issues/402) and at this point I'm not sure how this patch can affect that part. Error messages are likely to be cut off as far as I understand.

(I don't think the reflector/reflect filters/ini parser are covered by any tests so I added a fuzz target in https://github.com/avahi/avahi/pull/610. I'll try to figure out how to exercise the reflect filters)

dennypage:
I did see the other issue but it did not appear that a PR was done.

Regarding the other 256 instances, I took a look at some of them, such as host and interface names, and didn't see an intrinsic issue. Either they were not lists (such as hostnames), or the list was very limited in size/number (such as interface names). For the reflector list, the size of individual elements in the list is large, and the length of the list can be rather long as well.

evverx:
> I did see the other issue but it did not appear that a PR was done.

I was looking at https://github.com/astlinux-project/astlinux/commit/4e5a506e32735915da0caf27a587e823683b5ae4. There have been no PRs because it's kind of a stopgap. Ideally it shouldn't be necessary to bump it and recompile avahi every time the limit is hit so it should read lines and allocate memory properly instead.

> Either they were not lists (such as hostnames), or the list was very limited in size/number (such as interface names). For the reflector list, the size of individual elements in the list is large, and the length of the list can be rather long as well

Agreed. Normally it shouldn't matter. I was talking about places like https://github.com/avahi/avahi/blob/master/avahi-daemon/ini-file-parser.c#L89 where error messages are shown because of typos, missing equal signs or something like that.

dennypage:
> I was talking about places like https://github.com/avahi/avahi/blob/master/avahi-daemon/ini-file-parser.c#L89 where error messages are shown because of typos, missing equal signs or something like that.

Good point. Given that they currently have the same buffer limit, it's kinda intrinsically broken. I've just made it worse. :)

I'll push another a fix to expand the error buffer.

evverx:
> it's kinda intrinsically broken

Agreed. It's broken. The log limit is kind of tricky because `set_log_function` allows passing external functions accepting log levels and log messages. It's public API and there can be places outside of the avahi repository where log messages longer than the current default can cause undesired side effects if, say, those log functions copy log messages into their internal buffers without checking their lengths or something like that. That's why https://github.com/avahi/avahi/pull/409 was closed.

All in all I think it should be OK to bump the limit in the ini parser only to somewhat address the immediate issue but let's wait for the others to chime in.

dennypage:
May we proceed with the PR?

evverx:
Could you drop the second commit (where the log limit was bumped)? It can't be fixed that way unfortunately.

The first commit looks good to me. It would be great to add a link to https://github.com/avahi/avahi/issues/315 to the commit message (but it shouldn't be "fixes/closes #315" or anything like that to prevent it from being closed automatically. I think "partially addresses #315" should be fine)

dennypage:
> Could you drop the second commit (where the log limit was bumped)? 

I must have misunderstood you previously? I thought you had asked me to address the situation with logging of longer ini file input errors. Was that not correct?

evverx:
> I thought you had asked me to address the situation with logging of longer ini file input errors. Was that not correct?

It should be fixed but it can't be bumped like that because log functions can be set via `set_log_function` and those external functions can rely on the length of that buffer for whatever reason. It was discussed in https://github.com/avahi/avahi/pull/409#discussion_r1465020040.

dennypage:
Understood that it doesn't fix external loggers, but doesn't it address the situation when external log functions are not in use? Such as when the avahi daemon itself is trying to log errors with the ini file?

evverx:
avahi provides the avahi-core library that can be used to embed the mDNS/DNS-SD stuff into bespoke mDNS responders and `avahi_set_log_function` is included in its public API. Once it's updated it affects all the clients (including `avahi-daemon`).

dennypage:
How should we fix it then?

evverx:
I think it should look like https://github.com/avahi/avahi/pull/409#issuecomment-2043726081.

Just to clarify I don't expect it to be fixed in this PR. I don't expect https://github.com/avahi/avahi/issues/315 to be fully fixed here either. The first commit should be enough to make reflect-filters more or less work in most cases.

dennypage:
Okay, I reverted the log change.

Issue #615, opened 2024-07-26
sendmsg() to ff02::fb failed: Invalid argument


ceggers-arri:
If running the `avahi-daemon` with `--debug`, I see the following messages after attaching the network cable:

```
sendmsg() to ff02::fb failed: Invalid argument
sendmsg() to ff02::fb failed: Invalid argument
sendmsg() to ff02::fb failed: Invalid argument
sendmsg() to ff02::fb failed: Invalid argument
```
My main concern is, that some probably important messages are dropped by the kernel so that other nodes on the network cannot receive them (as long as sending them is not further repeated by Avahi).

The Avahi call stack is:
```
sendmsg_loop()
avahi_send_dns_packet_ipv6()
avahi_interface_send_packet_unicast()
avahi_interface_send_packet()
elapse_callback()
```

I found out that the `-EINVAL` is returned concretely by the kernel here (note that `-EINVAL` is also used at many other locations in the UDPv6 tx path):
```
int ip6_datagram_send_ctl()
{
...
		switch (cmsg->cmsg_type) {
		case IPV6_PKTINFO:
			...
			if (addr_type != IPV6_ADDR_ANY) {
				int strict = __ipv6_addr_src_scope(addr_type) <= IPV6_ADDR_SCOPE_LINKLOCAL;
				if (!ipv6_can_nonlocal_bind(net, inet_sk(sk)) &&
				    !ipv6_chk_addr_and_flags(net, &src_info->ipi6_addr,
							     dev, !strict, 0,
							     IFA_F_TENTATIVE) &&
				    !ipv6_chk_acast_addr_src(net, dev,
							     &src_info->ipi6_addr))
					err = -EINVAL;  <-- this one
				else
					fl6->saddr = src_info->ipi6_addr;
			}
			...
		}
...
}
```
The problem seems to be, that avahi-daemon starts sending datagrams with `IPV6_PKTINFO` as soon as an address (e.g. IPv6 link local) is added to the interface. But during the _IPv6 Duplicate Address Detection (DAD)_ phase, the `IFA_F_TENTATIVE` flag is set for that address and the kernel refuses to use it as source IP.

A possible exception from this rule is if _Optimistic Duplicate Address Detection_ ([RFC4429](https://tools.ietf.org/rfc/rfc4429.txt)) would be active, in this case additionally the `IFA_F_OPTIMISTIC` flag could be set and the address may be used from the early beginning:
```
__ipv6_chk_addr_and_flags()
{
...
[...]

PR #589, opened 2024-04-19
core: Mark new addresses with a lifetime of 0 as deprecated

dan-eero:
When Linux receives an RA for an on-link, autonomous prefix with a preferred lifetime of 0, it will assign itself an address and mark it as deprecated. This results in two Netlink messages being sent: one for the new address, and another updating the address as deprecated. Avahi responds to these messages respectively by first registering the address, then withdrawing it. Subsequent RAs for the deprecated prefix will result in repeated registers and withdraws until the address's valid lifetime expires.

This patch uses the IFA_CACHEINFO attribute of the RTM_NEWADDR message to get the lifetime of the address and uses this to preemptively deprecated it.

Fixes https://github.com/avahi/avahi/issues/41

PR #616, opened 2024-07-26
iface: Check for IFA_F_TENTATIVE

ceggers-arri:
Skip (IPV6) addresses during Duplicate Address Detection (DAD) phase. Using them as source address when sending datagrams would be rejected by the kernel (#615).

### DIFF
--- a/avahi-core/iface-linux.c
+++ b/avahi-core/iface-linux.c
@@ -183,6 +183,7 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
         size_t l;
         AvahiAddress raddr, rlocal, *r;
         int raddr_valid = 0, rlocal_valid = 0;
+        uint32_t flags = ifaddrmsg->ifa_flags;  /* may be overridden by IFA_FLAGS */
 
         /* We are only interested in IPv4 and IPv6 */
         if (ifaddrmsg->ifa_family != AF_INET && ifaddrmsg->ifa_family != AF_INET6)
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR!

I'll take a look next week.

I'll add the "needs-test" label because that part isn't tested by the CI at all currently. It isn't prepared to bring up/down interfaces either. Either way the label is just a reminder that it should be done eventually.

evverx:
Sorry for the delay! I haven't looked at this PR closely yet (because I think it would make sense to wait for a reply in https://github.com/avahi/avahi/pull/617) but I rolled it out on my test machines the other day and as far as I can see the issue is gone.

(In the meantime it reminded me of https://github.com/avahi/avahi/pull/589 somehow and it seems the IPv6 part needs to be looked into in general)

AhmTaher (approved):

evverx:
@arrowd I've just tested this patch on FreeBSD 15.0 with https://github.com/avahi/avahi/pull/799 and as far as I can see avahi can no longer pick up link-local IPv6 addresses when the interfaces are restarted with `service netif restart`/brought up with `service netif start` (but they can be picked up when avahi itself is restarted). Do you know what's going on there?

evverx:
With https://github.com/avahi/avahi/commit/aa0214dacb5d4f79eba445ac889ee1eacc264aa3 reverted link-local IPv6 addresses are picked up on FreeBSD but the bug described in https://github.com/avahi/avahi/issues/615 is present. It manifests itself differently
```sh
sendmsg() to ff02::fb (iface #2) failed: Can't assign requested address
sendmsg() to ff02::fb (iface #2) failed: Can't assign requested address
sendmsg() to ff02::fb (iface #2) failed: Can't assign requested address
sendmsg() to ff02::fb (iface #2) failed: Can't assign requested address
```
The result is the same though in terms of not actually sending probes.

arrowd:
Nope, no idea.

evverx:
On FreeBSD avahi receives netlink messages with the IFA_F_TENTATIVE flag set and that's it. They aren't followed by anything else.

PR #636, opened 2024-08-24
Recognize optimistic IPV6 address as deprecated

pemensik:
Optimistic should be treated as deprecated address. The difference is it starts with optimistic flag and unless conflict is detected, it loses optimistic flag to become normal primary address.

https://www.rfc-editor.org/rfc/rfc4429#section-2.1


Neustradamus:
@pemensik: More one year after, your PR is ready?

Issue #347, opened 2021-07-07
init_rand_seed(): consider failing if real random data is not available


mgerstner:
Currently the `init_rand_seed()` function (existing twice in avahi-daemon and avahi-autoipd) silently ignores failures to open /dev/urandom. If this actually fails then seeding `srand()` only with `time(NULL)` makes random numbers used within Avahi pretty predictable, which could prove problematic now or in the future.

Using /dev/urandom or even `getrandom()` is state of the art nowadays and if it fails it should be considered a critical error.

Issue #254, opened 2019-11-27
avahi-daemon use random port besides 5353(mdns)


realduke2000:
Hi avahi experts,
The avahi-daemon seems to use a random port for udp listening(not 5353). Does anyone know what this port used for? Can we disable or use a fixed port for it? Is it configurable?

Thanks!!

my-test-service:/etc/avahi/services # netstat -plna|grep avahi 
udp        0      0 0.0.0.0:46974           0.0.0.0:*                           98440/avahi-daemon: 
udp        0      0 0.0.0.0:5353            0.0.0.0:*                           98440/avahi-daemon: 
unix  2      [ ACC ]     STREAM     LISTENING     34627    1/systemd            /var/run/avahi-daemon/socket
unix  3      [ ]         STREAM     CONNECTED     255589   98441/avahi-daemon:  
unix  3      [ ]         STREAM     CONNECTED     255593   98440/avahi-daemon:  
unix  3      [ ]         STREAM     CONNECTED     255588   98440/avahi-daemon:  
unix  2      [ ]         DGRAM                    59717    7280/avahi-autoipd:  
unix  3      [ ]         STREAM     CONNECTED     254845   98440/avahi-daemon:  
unix  2      [ ]         DGRAM                    255586   98440/avahi-daemon:  
my-test-service:/etc/avahi/services # systemctl restart avahi-daemon 
my-test-service:/etc/avahi/services # netstat -plna|grep avahi 
udp        0      0 0.0.0.0:34954           0.0.0.0:*                           98781/avahi-daemon: 
udp        0      0 0.0.0.0:5353            0.0.0.0:*                           98781/avahi-daemon: 
unix  2      [ ACC ]     STREAM     LISTENING     34627    1/systemd            /var/run/avahi-daemon/socket
unix  3      [ ]         STREAM     CONNECTED     261174   98781/avahi-daemon:  
unix  2      [ ]         DGRAM                    255811   98781/avahi-daemon:  
unix  3      [ ]         STREAM     CONNECTED     255814   98782/avahi-daemon:  
unix  3      [ ]         STREAM     CONNECTED     255818   98781/avahi-daemon:  
unix  3      [ ]         STREAM     CONNECTED     255813   98781/avahi-daemon:  
unix  2      [ ]         DGRAM                    59717    7280/avahi-autoipd:  
my-test-service:/etc/avahi/services #

lrodriguez14:
Hello!

Any updates or info about this? I am having the same issue. @lathiat 

[...]

Issue #118, opened 2017-05-11
Handle resolv.conf file changes


lathiat:
We should monitor and reload /etc/resolv.conf if changes are detected, because either 

 (a) The servers may change, or
 (b) The file may not yet be available during early startup as the network connection is not yet established

This information is used in two main areas, 
 (1) publish-resolv-conf-dns-servers,
 (2) as the list of servers to query for wide area DNS queries.

Ideally this would be done with inotify and not polled, but we also need to consider the chroot helper.

This would also avoid errors being logged on startup currently even if neither of these features are actually used.  Refer to #63 and #59.

lathiat:
When fixing the errors about loading resolv.conf, we also need to note that errors are produced both by load_resolv_conf and avahi_chroot_helper_get_file

aragilar:
As a suggestion, adding a dbus API which allows setting nameservers/search domains would mean the resolv.conf reading could be done in a separate, low-privileged service (as it would only need to read the file, and send the dbus messages), and could reduce the need to have different paths to handle the chroot and non-chroot cases. That would also mean that avahi could be told about DNS-SD domains independent of "normal" search domains.

ktamp:
Since [#159](https://github.com/lathiat/avahi/issues/159) is marked as a duplicate of this, I have to write here that what is really needed is that avahi-daemon.service restarts/refreshes e.g. every 10 seconds.

This way Avahi would always find network printers / scanners /multifunctionals that had just been switched on. It would also find computers which had just been turned on and which have enabled e.g. the sftp service in order for the home folders to be browsable using Nautilus.

In order to facilitate the above without editing avahi-daemon.service, one has to set a cron job or create and enable a systemd timer. For the systemd timer, creating avahi-daemon-timer.service and avahi-daemon-timer.timer in /usr/lib/systemd/system and enabling avahi-daemon-timer.timer would do the trick.

avahi-daemon-timer.service:
> [Unit]
> Description=avahi-daemon timer
> Requires=avahi-daemon.sevice
> 
> [Service]
> Type=oneshot
> ExecStart=/usr/bin/systemctl restart avahi-daemon.service
> 
> [Install]
> WantedBy=multi-user.target

avahi-daemon-timer.timer:

> [Unit]
> Description=avahi-daemon timer timer
> 
> [Timer]
> OnBootSec=30s
> OnUnitActiveSec=10s
> Unit=avahi-daemon-timer.service
> 
> [Install]
> WantedBy=multi-user.target

[...]

PR #659, opened 2024-11-16
Randomize transaction IDs in wide area queries

pemensik:
Use proper randomization for next transaction id. Do not use naive IDs.

### DIFF
--- a/avahi-core/wide-area.c
+++ b/avahi-core/wide-area.c
@@ -40,6 +40,13 @@
 #include "addr-util.h"
 #include "rr-util.h"
 
+#ifdef HAVE_SYS_RANDOM_H
+#include <sys/random.h>
+#endif
+#ifndef HAVE_GETRANDOM
+#  define getrandom(d, len, flags) (-1)
+#endif
+
 #define CACHE_ENTRIES_MAX 500
 
 typedef struct AvahiWideAreaCacheEntry AvahiWideAreaCacheEntry;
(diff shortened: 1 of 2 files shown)


pemensik:
@msekletar The CVE entry were made public by RH, there is not reason to continue in privacy, I guess.

pemensik:
This should fix CVE-2024-52616.

msekletar (commented):
LGTM

I think that `get_random_uint16()` could be more robust but I guess this is good enough for start.

evverx (commented):
I'm not sure why it was rushed in. I think it would actually be better to document how to point avahi to a localhost DNS resolver properly for now. As mentioned in https://github.com/avahi/avahi/issues/254#issuecomment-2481051549 it isn't straightforward due to things like https://github.com/avahi/avahi/issues/118.

Regarding random numbers in general I think it would be nice to address https://github.com/avahi/avahi/issues/347 first for example.

@pemensik @msekletar if you are planning to resurrect this feature eventually I think it would be great to start covering it with tests. Currently it isn't exercised by the CI at all. (At a minimum it should be possible to turn it on on the CI (where all the configs are edited with `sed` to flip various features and make it all compatible with ASan/UBsan and Valgrind) and add a smoke test talking to, say, `dns-sd.org` to make sure it doesn't crash anywhere or anything like that). Ideally the CI would set up a DNS server returning known responses to be able to make sure avahi actually gets them right.

(Either way it shouldn't be considered a full-blown review or anything like that)

evverx:
FWIW @pemensik in light of https://github.com/avahi/nss-mdns/issues/85#issuecomment-2465868418 and things like that personally I think it should be OK to invite @msekletar to the avahi organization (if @lathiat is OK with that and @msekletar is interested in that of course).

msekletar:
@evverx I am interested in joining the org.

evverx:
@msekletar I went ahead and invited you to the avahi repository. As far as I understand @pemensik and you are colleagues so it should be possible for Petr to invite you to the other repositories and bump the role if necessary. It would be great if more people reviewed upstream PRs like https://github.com/avahi/avahi/pull/676 :-)

PR #660, opened 2024-11-19
Initialize srand from better random source if supported

### LINKED ISSUE #347: init_rand_seed(): consider failing if real random data is not available

pemensik:
Fixes #347

### DIFF
--- a/avahi-autoipd/main.c
+++ b/avahi-autoipd/main.c
@@ -169,32 +169,12 @@ typedef struct CalloutEventInfo {
     int ifindex;
 } CalloutEventInfo;
 
-#define RANDOM_DEVICE "/dev/urandom"
-
 #define DEBUG(x)                                \
     do {                                        \
         if (debug) {                            \
             x;                                  \
         }                                       \
     } while (0)
-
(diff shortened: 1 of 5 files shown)


evverx:
Looks like it got abandoned so I'll go ahead and close it.

> Fixes https://github.com/avahi/avahi/issues/347

It doesn't fix it because it still silently falls back to `time` in some cases. Other than that with f8710bdc8b29ee1176fe3bfaeabebbda1b7a79f7 merged it's now badly sprinkled all over the place.

By the way that issue is reported by a member of the SUSE security team and they generally participate in code reviews with helpful feedback so next time it would be great to request a review from the reporter.

PR #661, opened 2024-11-21
Use `-flto=auto` as GCC flag in OpenScanHub scans

siteshwar:
... submitted via Packit.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -8,7 +8,7 @@ upstream_package_name: avahi
 downstream_package_name: avahi
 upstream_tag_template: "v{version}"
 srpm_build_deps: []
-csmock_args: --cppcheck-add-flag=--enable=style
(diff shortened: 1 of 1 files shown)


siteshwar:
Enabling `lto` may produce more [accurate results](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/NWTU2DSEQDMS7FMDEMXP62EMATA2UD5M/) in certain cases. But it is still in early [development phase](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117662#c4). Please check the [scan results](https://openscanhub.fedoraproject.org/task/25029/) and merge if it suppresses false positives. Thanks!

siteshwar:
Looks like it suppresses all cppcheck warnings. So, I am going to close this.

PR #667, opened 2024-12-10
Add a shell completion function for zsh

okapia:
This is PR #152 again but with integration to the build system added. (I don't have permissions to reopen the previous PR).

The existing build system seems to be generous in the use of configure --disable/--enable--with etc arguments so this adds `--disable-zsh-completion` for users that want to prevent the installation of the completion files. Let me know if you want a different approach, or different naming or other tweaks to that.

I also checked that the completion file is still valid and up-to-date for the current state of avahi and didn't see any changes that might need to be reflected in the completion function.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,7 +79,8 @@ SUBDIRS = \
 	avahi-ui \
 	avahi-ui-sharp \
 	avahi-libevent \
-	po
+	po \
+	completion
 
 DX_INPUT = \
 	$(srcdir)/avahi-common/address.h \
--- a/completion/Makefile.am
+++ b/completion/Makefile.am
@@ -1,0 +1,25 @@
+# This file is part of avahi.
+#
+# avahi is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# avahi is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public
+# License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with avahi; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
+# USA.
+
+if ENABLE_ZSH_COMPLETION
+
+zshcompdir = $(datadir)/zsh/site-functions
+dist_zshcomp_DATA = _avahi
+
+endif
+
+EXTRA_DIST = _avahi
--- a/completion/_avahi
+++ b/completion/_avahi
@@ -1,0 +1,54 @@
+#compdef avahi-browse avahi-browse-domains avahi-resolve avahi-resolve-address avahi-resolve-host-name
+
+# Completion function for zsh
+
+local curcontext="$curcontext" pri ret=1
+local -A opt_args
+local -a state line args serv desc disp
+
+args=(
+  '(- 1)'{-h,--help}'[display usage information]'
+  '(- 1)'{-V,--version}'[display version information]'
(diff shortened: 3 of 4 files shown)


evverx:
Also all those binaries are installed with `HAVE_DBUS` so it should probably be taken into account as well. It shouldn't be installed if the binaries aren't installed. Either way I'll take a closer look a bit later.

okapia:
> Packit fails with

It looks like that relies on a spec file checked out from `https://src.fedoraproject.org/rpms/avahi`. That probably only needs one line adding under `%files tools` for `${_datadir}/share/zsh/site-functions/_avahi` (I've experience of rpm .spec files but not of packit).

> I'm not sure how to structure it. The "completion" directory is kind of generic. Would it maybe make sense to name it `shell-completion` (or something like that) and put the zsh completion to the `zsh` directory. It should make it easier to add the bash completion eventually (if it's ever added of course).

Those questions are more a matter of personal taste. `shell-completion` is certainly less ambiguous. As for further subdirectories, I don't normally like deep structures of directories each containing only one or two files. However, whereas for zsh you where it is a single file named `_avahi`, for bash you may end up with `avahi-browse.bash`, `avahi-resolve.bash` and so on.

Renaming the directories should be a fairly trivial change, however.

evverx:
> It looks like that relies on a spec file checked out from https://src.fedoraproject.org/rpms/avahi

That's correct. It just clones it, tweaks the spec file a bit and then builds the package: https://github.com/avahi/avahi/blob/master/.packit.yml. It can all be fixed later once the details are settled.

By the way is there a reason why `avahi-publish`, `avahi-publish-address`, `avahi-publish-service` and `avahi-set-hostname` are missing?

okapia:
I've pushed additions to:

1. make the completion directory configurable via `--with-zsh-completion-dir=path`. This still allows for `--without-zsh-completion-dir` to disable the installation. I think Ubuntu have their own different convention for naming here so agree that this is useful.
2. Support `avahi-publish` and `avahi-set-host-name` in the zsh completion. I can't remember why I left them out seven years ago. Probably didn't need them myself.

It may be that it can do something more useful for compieting subtypes etc for `avahi-publish`. I'm not an expert on avahi. And I'm not completely sure that `_ports` is the right thing for ports, it completes TCP/TDP ports.

Feel free to squash the three commits (or not if you prefer). Or I can do a squash and force-push if you like.

evverx:
I'll take a closer look later. Maybe the other maintainers will take a look too.

When I pulled this PR I actually accidently found a bug in the build system thanks to the completion. Long story short if the type database is built with python3.13  it can't be read by avahi-browse so the completion gets stuck until `-` is typed and `tab` is then pressed to complete the options.

evverx:
In the latest version of this PR the completion isn't installed by default and I think it makes sense. This feature is optional and should probably be enabled explicitly with `--with-zsh-completion-dir` (I'd drop the `dir` part but it doesn't matter much). It can also prevent it from being automatically enabled when it doesn't make sense (for example when `dbus` is disabled). I think it would be better to make it explicit by adding something like
```diff
 AC_ARG_WITH([zsh-completion-dir],
-       AS_HELP_STRING([--with-zsh-completion-dir=DIR],[Directory for zsh completion file]))
+       AS_HELP_STRING([--with-zsh-completion-dir=DIR],[Directory for zsh completion file]),
+        [],
+        [with_zsh_completion_dir=no])
```
With that the build system would print something like `zsh completion: no` instead of `zsh completion:`.

(It would be great to replace the tabs with spaces in that part to make it consistent with `configure.ac` in general)

It would probably make sense to enable it in https://github.com/avahi/avahi/blob/1dade81cbf4d3e3f49784e44e543cd77046c79df/.github/workflows/build.sh#L81-L90 to let the CI make sure it's buildable and installable with that flag enabled.

I've installed the completion and am planning to use it a bit to figure out whether anything should be changed there.

okapia:
So is there anything further you would like changed here?

evverx:
Apart from the things mentioned in https://github.com/avahi/avahi/pull/667#issuecomment-2536214748 it looks good to me.

Issue #243, opened 2019-08-27
RFE: Always advertise link-local addresses


srd424:
Avahi doesn't advertise a link-local address if a global address is available. This was queried many years ago on the mailing list: https://lists.freedesktop.org/archives/avahi/2010-March/001863.html

It does in fact cause problems in some circumstances, where some machines have e.g. static global addresses assigned, and others have link local only. For example, Pidgin's Bonjour module is unable to match the source address on an incoming connection with the address that was originally advertised: https://bugs.launchpad.net/ubuntu/+source/pidgin/+bug/1841621

There's an old patch in Launchpad to change this behaviour, but a quick review suggests it is not quite correct: https://bugs.launchpad.net/ubuntu/+source/avahi/+bug/1102906

KaiJan57:
I can confirm that this is a problem. In my configuration, I have avahi configured to advertise a squid server on IPv6 that only accepts connections from link-local addresses. As avahi always chooses the global address, clients on lan cannot connect to the squid server by the address avahi advertises. To work around avahi not preferring link-local addresses, I configured avahi to only advertise the squid server's IPv4 address.

lathiat:
I think we can do this but first we need to implement support for "ordering" the list of addresses given by both API calls and nss-mdns - as right now the order is random and it's possible to give the 'less preferred' address. Which can be a problem particularly for link local addresses if applications don't support setting the sin6_scope.

nss-mdns now sets sin6_scope so that helps a lot - it's part of the reason it was disabled in the past. but I think we should still order the lookups first.

gabekassel:
@lathiat any updates on this enhancement? Thanks so much!

EskoDijk:
Most discussion on this are around which address to use; e.g. link-local or global address.  But what about the requirements for mDNS RFC 6762 [section 6.2](https://datatracker.ietf.org/doc/html/rfc6762#section-6.2) ?

> When a Multicast DNS responder sends a Multicast DNS response message
>    containing its own address records, it MUST include all addresses
>    that are valid on the interface on which it is sending the message
...

> For example, if an interface has both an IPv6 link-
>    local and an IPv6 routable address, both should be included in the
>    response message 

So this says both addresses must be included in case global/link-local addresses are available.  Then an ordering of addresses seems not needed, right?

TownCube:
[...]

PR #673, opened 2025-01-06
Add config option to always allow publishing of link-local address 

notthetup:
Currently, `avahi-daemon` will only publish a link-local address on a interface, IFF there are no other "global" addresses assigned to the same interface. This was widely discussed in https://github.com/avahi/avahi/issues/243, and it seems that the current behavior is counter to RFC6762. 

This PR adds a new config option to `avahi-daemon.conf`, `always-publish-linklocal` which defaults to `no`, but when set to `yes`, will always publish a link-local address regardless other addresses on the same interface.

Hopefully this approach maintains backward compatibility and yet allows users to configure behavior.

P.S. This is my first PR to avahi, let me know if I'm missing something, I'll try to fix asap.

### DIFF
--- a/avahi-core/core.h
+++ b/avahi-core/core.h
@@ -70,6 +70,7 @@ typedef struct AvahiServerConfig {
     unsigned n_cache_entries_max;     /**< Maximum number of cache entries per interface */
     AvahiUsec ratelimit_interval;     /**< If non-zero, rate-limiting interval parameter. */
     unsigned ratelimit_burst;         /**< If ratelimit_interval is non-zero, rate-limiting burst parameter. */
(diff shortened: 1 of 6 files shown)


evverx (commented):
Thank you for the PR!

Unfortunately this patch doesn't address the issues where clients have to get all the IP addresses. The D-Bus API (used by most clients) and the simple protocol (used by nss-mdns) still return one IPv6 address and for example the only way to see what avahi advertises is to send USR1 to avahi to dump its caches. The other issue is that with link-local addresses avahi should properly work with zone identifiers.

I of course agree that avahi MUST include all addresses that are valid on the interface but the change would be much more involved to get it to work.

PR #680, opened 2025-02-10
Replace shebang with a portable one

Awoonyaa:
@arrowd

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eux
 set -o pipefail
--- a/.github/workflows/smoke-tests.sh
+++ b/.github/workflows/smoke-tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eux
 set -o pipefail
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This file is part of avahi.
 #


evverx:
Thank you for the PR. It looks good to me on the whole but this "sponsored-by" thing doesn't look right in the sense that it doesn't match the author mentioned in the commit message (I don't think the email address should be included there either).

Generally it's a FreeBSD thing (and it has never been used in the avahi repository) so I wonder if it can be dropped entirely? Is is required for you to include it everywhere for some reason even when trivial patches are submitted?

arrowd:
> it doesn't match the author mentioned in the commit message

It is the same person.

> Generally it's a FreeBSD thing (and it has never been used in the avahi repository) so I wonder if it can be dropped entirely? Is is required for you to include it everywhere for some reason even when trivial patches are submitted?

I'm baffled why you make so much attention to this. Yes, it can be dropped, but first you said that it is OK to have it. This tag does not impose any legal consequences. We're adding this purely to have something to brag on. Some upstream projects also love when they see patches coming from paid developers. Anyways, we can remove them from all our future commits.

evverx:
> I'm baffled why you make so much attention to this

Because it's a downstream thing. Upstream it's usually possible to use email address to show where things come from and this commit actually shows where it comes even with "sponsored-by" removed.

> first you said that it is OK to have it

I did say that (and also asked to add `(author)` and gave the link to the documentation where it's described how it should be written). Either way can you either drop it or make it consistent with the author? Other than that I think it can be merged.

> Some upstream projects also love when they see patches coming from paid developers

I'm 99% sure that most commits here come from companies.

arrowd:
> I did say that (and also asked to add (author) and gave the link to the documentation where it's described how it should be written).

The docs talks about adding an author in parentheses, quoting, "If only a portion of the work was sponsored, or different amounts of sponsorship were provided to different authors". Which is not the case here. Mary is the sole author of these changes and all her work was sponsored. It is enough to have only "Sponsored-by: Our company" for this case.

> Either way can you either drop it or make it consistent with the author?

We lost so much time on this pointless argument that I'd better just remove it everywhere.

evverx:
> We lost so much time on this pointless argument that I'd better just remove it everywhere

Agreed.

> If only a portion of the work was sponsored

I understand that it looks like the code review (where it turned out that only this change is mergeable) appeared out of thin air.

Issue #678, opened 2025-02-04
Reachable assertion in `lookup_start` caused by `CNAME` resource records


evverx:
```sh
avahi-daemon: browse.c:323: lookup_start: Assertion `!l->wide_area && !l->multicast' failed.                                                                  
==426592==                                                                                                                                                    
==426592== Process terminating with default action of signal 6 (SIGABRT)                                                                                      
==426592==    at 0x4ADA834: __pthread_kill_implementation (in /usr/lib64/libc.so.6)                                                                           
==426592==    by 0x4A888ED: raise (in /usr/lib64/libc.so.6)                                                                                                   
==426592==    by 0x4A708FE: abort (in /usr/lib64/libc.so.6)                                                                                                   
==426592==    by 0x4A7081A: __assert_fail_base.cold (in /usr/lib64/libc.so.6)                                                                                 
==426592==    by 0x4A80C56: __assert_fail (in /usr/lib64/libc.so.6)                                                                                           
==426592==    by 0x488AEC0: lookup_start (browse.c:323)                                                                                                       
==426592==    by 0x488B295: lookup_go (browse.c:392)                                                                                                          
==426592==    by 0x488B458: lookup_handle_cname (browse.c:422)                                                                                                
==426592==    by 0x488AC4E: lookup_multicast_callback (browse.c:278)                                                                                          
==426592==    by 0x489C483: avahi_multicast_lookup_engine_notify (multicast-lookup.c:317)                                                                     
==426592==    by 0x48811A7: avahi_cache_update (cache.c:363)                                                                                                  
==426592==    by 0x4876202: handle_response_packet (server.c:720)                                                                                             
==426592==                                                             
```

The crash can be triggered reliably by sending unsolicited replies 2 seconds apart when an A/AAAA record browser sends its queries. Here's what it looks like
```sh
    1   0.000000 192.168.56.3 → 224.0.0.251  MDNS 82 Standard query response 0x0000 CNAME H.local
    2   2.016982 192.168.56.3 → 224.0.0.251  MDNS 82 Standard query response 0x0000 CNAME H.local
    3   2.988520 192.168.56.4 → 224.0.0.251  MDNS 91 Standard query 0x0000 A H.local, "QM" question A X.local, "QM" question A 192.168.56.4
    4   3.994824 192.168.56.4 → 224.0.0.251  MDNS 91 Standard query 0x0000 A H.local, "QM" question A X.local, "QM" question A 192.168.56.4
[...]

PR #551, opened 2024-01-26
WIP: Attempt to limit CNAME loops

pemensik:
Limit number of chained CNAMEs. Do not allow to query also following to cname key, which we already have in cname chains.

Fixes #501

Issue #501, opened 2023-10-26
Uncontrolled recursion in `lookup_handle_cname` caused by recursive CNAMEs


evverx:
When avahi receives recursive CNAMEs it can crash depending on how the resolver has been created. It can be triggered remotely (https://github.com/lathiat/avahi/issues/391) or locally (https://github.com/lathiat/avahi/issues/78). It can be reproduced (by unprivileged local users) by applying the following patch
```diff
diff --git a/avahi-client/client-test.c b/avahi-client/client-test.c
index da0e43a..f3da437 100644
--- a/avahi-client/client-test.c
+++ b/avahi-client/client-test.c
@@ -24,6 +24,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <assert.h>
+#include <unistd.h>
 
 #include <avahi-client/client.h>
 #include <avahi-client/lookup.h>
@@ -275,8 +276,14 @@ int main (AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     assert(error == AVAHI_ERR_INVALID_RECORD);
     avahi_string_list_free(txt);
 
+    error = avahi_entry_group_add_record (group, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, 0, "Z.local", 0x01, AVAHI_DNS_TYPE_CNAME, 120, "\1Z\5local\0", 9);
+    assert(error == AVAHI_OK);
+
     avahi_entry_group_commit (group);
 
+    hnr = avahi_host_name_resolver_new (avahi, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, "Z.local", AVAHI_PROTO_UNSPEC, AVAHI_LOOKUP_USE_MULTICAST, avahi_host_name_resolver_callback, (char*) "omghai4u");
+
+
     domain = avahi_domain_browser_new (avahi, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, NULL, AVAHI_DOMAIN_BROWSER_BROWSE, 0, avahi_domain_browser_callback, (char*) "omghai3u");
 
     if (domain == NULL)
```
building the test, running `avahi-daemon` and running the test. avahi goes down with
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7866607 in __interceptor_strlen.part.0 () from /lib64/libasan.so.8
[...]

PR #682, opened 2025-02-22
Fix multiple CNAME lookup issues in avahi-daemon

friendlyhugo:
This PR supersedes https://github.com/avahi/avahi/pull/551.

The changes proposed fix the following issues:
- https://github.com/avahi/avahi/issues/501: fixed by checking for CNAME loops in `lookup_handle_cname` using a simple algorithm. If a loop is detected while handling a new record, the record is ignored.
- https://github.com/avahi/avahi/issues/678: fixed by changing the relevant `assert` to an `if-return` statement.

The loop detection algorithm finds the paths to a lookup and makes sure the desired child isn't on them.

New smoke-tests are added to prevent future regressions, see `avahi-core/cname-lookup-test.c`.

### DIFF
--- a/.github/workflows/smoke-tests.sh
+++ b/.github/workflows/smoke-tests.sh
@@ -156,7 +156,7 @@ should_fail systemctl is-failed avahi-daemon
 should_fail avahi-daemon -c
 should_fail avahi-dnsconfd -c
 
-for test_case in self_loop retransmit_cname one_normal one_loop two_normal two_loop two_loop_inner two_loop_inner2 three_normal three_loop diamond; do
+for test_case in self_loop retransmit_cname one_normal one_loop two_normal two_loop two_loop_inner two_loop_inner2 three_normal three_loop diamond cname_answer_diamond cname_answer; do
     run ./avahi-core/cname-test $test_case
 done
 


siteshwar:
It is just an issue with styling, but you may want to fix what is reported through OpenScanHub [scan](https://openscanhub.fedoraproject.org/task/39188/log/added.html):

```
Error: CPPCHECK_WARNING ([CWE-398](https://cwe.mitre.org/data/definitions/398.html)): [[#def1]](https://openscanhub.fedoraproject.org/task/39188/log/added.html#def1)
avahi-0.9.rc2-build/avahi-0.9.rc2/avahi-core/browse.c:406: style[variableScope]: The scope of the variable 'r' can be reduced.
#  404|   static int lookup_exists_in_path(AvahiSRBLookup* lookup, AvahiSRBLookup* from, AvahiSRBLookup* to) {
#  405|       AvahiRList* rl;
#  406|->     int r;
#  407|       if (from == to)
#  408|           return 0;
```

avahi-core/core.h
```diff
@@ -70,6 +70,7 @@ typedef struct AvahiServerConfig {
     unsigned n_cache_entries_max;     /**< Maximum number of cache entries per interface */
     AvahiUsec ratelimit_interval;     /**< If non-zero, rate-limiting interval parameter. */
     unsigned ratelimit_burst;         /**< If ratelimit_interval is non-zero, rate-limiting burst parameter. */
+    int abort_on_cname_loop;          /**< Abort on CNAME loops. If this is enabled Avahi will abort when a CNAME loop is detected */
```
evverx, line 73: Can you expand on why it should be configurable?

evverx (commented):
Thank you for the PR!

As mentioned the code paths responsible for CNAMEs are kind of broken and unfortunately the full fix should take more things into account.

I've just opened https://github.com/avahi/avahi/issues/683 (which is another DoS caused by CNAMEs).

avahi-core/core.h
evverx: > This assertion should be reachable only while testing and avoids runtime DoS.

I'm not sure I understand why it should be reachable. For example the `self_loop` test reproduces https://github.com/avahi/avahi/issues/501 so it seems it should be enough to just run it to make sure that the issue is gone and prevent regressions.

evverx: @friendlyhugo before proceeding with this I wonder if this PR is AI-generated?

evverx: > No... why?

I asked that because these days it's not uncommon for PRs to be AI-generated (or more precisely AI-assisted if you will) and reviewing PR like that takes a lot of time and doesn't necessarily lead to anything in the end. My question was prompted by patterns used in this PR and the approach towards fixing specific assertions without addressing the fundamental issues. In its current form CNAMEs just don't work properly. For example it isn't possible to subscribe to CNAME RR because avahi always follows them. The events clients receive are messed up as well (I opened https://github.com/avahi/avahi/issues/684 where it can be seen in action). The crashes are just a symptom and that part of avahi should be implemented properly.

evverx: > it is logically the right one

I don't think it's accurate. https://github.com/avahi/avahi/issues/683 is directly related to the events clients receive and those things aren't taken into account (and aren't tested either).

> My PR currently fixes (and tests) three different bugs

It doesn't fix them. Have you looked at queries avahi sends with the patch hiding https://github.com/avahi/avahi/issues/501 using Wireshark? Do you think it makes sense for avahi to do that?

> especially when considering upcoming distro releases

Distros are free to backport these patches if they are interested in that.

> it shouldn't delay this PR

I don't think anyone in a hurry here.

evverx: @friendlyhugo I already said that I think that this part of avahi should be implemented properly and it doesn't make much sense to play whack-a-mole. I'll leave it to @pemensik if @pemensik is interested in reviewing, testing and merging this.

Neustradamus:
Dear @avahi team,

Any progress on this PR?

pemensik:
Ah, perfect, code with a test itself! For the test, I would like to have it runnable just by ./cname-test, running all test cases, unless I choose some. But that could be solved anytime later. We need easily runnable tests not only from github PR, but also when testing packages.

I miss at least debug message whe cname loop is detected, printing the failing query identification for analysis.

The test has obsolete GNU address btw, but that is common for all existing files. That test just ensures it will not crash, right? It does not verify following CNAMEs works, if I am correct.

.github/workflows/smoke-tests.sh
```diff
@@ -156,4 +156,8 @@ should_fail systemctl is-failed avahi-daemon
 should_fail avahi-daemon -c
 should_fail avahi-dnsconfd -c
 
+for test_case in self_loop retransmit_cname one_normal one_loop two_normal two_loop two_loop_inner two_loop_inner2 three_normal three_loop diamond cname_answer_diamond cname_answer; do
```
pemensik, line 159: I think we would like to have such tests, if they are reasonable self-contained, runnable separately. Eventually some target like ``make check`` should run tests we have. This running per test case is somehow difficult to work with.

avahi-core/cname-test.c
```diff
@@ -0,0 +1,272 @@
...
+int main(int argc, char *argv[]) {
+    int error;
+    struct timeval tv;
+
+    AvahiSHostNameResolver *hnr;
+    AvahiServerConfig config;
+    AvahiSimplePoll *simple_poll;
+
+    if (argc < 2) {
+        printf("Usage: %s test_name", argv[0]);
```
pemensik, line 235: Add \n to the end please.

avahi-core/cname-test.c
```diff
@@ -0,0 +1,272 @@
...
+}
+
+#define CHECK_TEST_CASE(test_name, function) \
+    if (!strcmp(test_case, test_name)) \
+        avahi_test_case_function = &function;
+
+static void avahi_test_initialize(char *test_case) {
+    assert(test_case);
+
+    CHECK_TEST_CASE("self_loop", self_loop);
```
pemensik, line 209: I think this could have been normal table and if no test case is specified, it should ideally run one test after another. with some parameter for listing test cases, maybe "--list"?

avahi-core/cname-test.c
```diff
@@ -0,0 +1,272 @@
...
+        const AvahiAddress *a,
+        AVAHI_GCC_UNUSED AvahiLookupResultFlags flags,
+        AVAHI_GCC_UNUSED void *userdata) {
+    char t[AVAHI_ADDRESS_STR_MAX];
+
+    if (a)
+        avahi_address_snprint(t, sizeof(t), a);
+
+    avahi_log_debug("HNR: (%i.%i) <%s> -> %s [%s]", iface, protocol, hostname, a ? t : "n/a",
+                    resolver_event_to_string(event));
```
pemensik, line 199: I think we want to have expected event result stored for each test case. Return 1 from main if it does not match. This prints failure for many cases, but still return 0. But compared to other tests we have, it is enough for now. Thank you!

evverx:
> I would like to have it runnable just by ./cname-test, running all test cases, unless I choose some. But that could be solved anytime later

It would be great not to skip this especially when DoSes triggered by incoming packets are fixed.

> We need easily runnable tests

That's what `make check` is for and it was made to run the basic tests recently (and because of that Packit was able to catch those crashes with `_FORTIFY_SOURCE` for example). It would be great not to neglect that again.

> not only from github PR, but also when testing packages

 It's true that the smoke test is run on GitHub only (and it's not ideal) but given that it involves a bunch of things like ASan, Valgrind and so on it's safe to say that it's too high a bar for most packages. Even projects with fancy integration test suites preparing images for running their tests still can't cover all those things.

> It does not verify following CNAMEs works

avahi doesn't handle CNAMEs correctly in general so the test testing that would fail regardless of this PR.

avahi-core/browse.c
```diff
@@ -417,6 +453,12 @@ static void lookup_handle_cname(AvahiSRBLookup *l, AvahiIfIndex interface, Avahi
         return;
     }
 
+    if (cname_would_create_loop(l, n)) {
+        /* CNAME loops are not allowed */
```
pemensik, line 457: I would insert at least debug log printing offending query. We do not respond anything sensible in case of error to clients. At least some kind of log should explain we did not respond to this query on purpose.

pemensik:
What I have seen in BIND9 code is they mostly just track number of CNAME/DNAME redirections and stop when they are too many of them. We have something similar kind of implemented in AVAHI_LOOKUPS_PER_BROWSER_MAX, but not directly refusing it. But we lack good reporting of failed query anyway, we just keep it ignored at best.

pemensik:
Btw. I think at least in unicast DNS cname cannot lead to two different targets. It must lead from single name to single target only, so cname_answer_diamond probably should have been refused in the middle also. There can be multiple redirections, but number of targets should stay at 1.

friendlyhugo:
Thanks for taking the time to review my changes!
@pemensik I'll look at your new PR ASAP :)

PR #687, opened 2025-03-08
Tidy up the dependencies section of README

### LINKED ISSUE #681: systemd requirement - should be optional
**Avahi version**
avahi-0.9-rc2

**Distribution**
Slackware64-current

**Describe the bug**
Please omit the requirement for libsystemd or make it optional. Not every Linux operating system is infested with systemd, particularly Slackware.

**To Reproduce**
`./autogeh.sh` configure fails "No package 'libsystemd' found"

**Expected behavior**
Create a Makefile

**Additional context**
README contains no mention of this systemd outrage. At least add this hidden requirement up front instead of ambushing unsuspecting sysadmins.

arrowd:
Fixes #681

### DIFF
--- a/docs/README
+++ b/docs/README
@@ -13,15 +13,16 @@ You have a chance to meet the developers on
 	#avahi on irc.libera.chat
 
 Avahi has the following requirements:
-	- glib2
+	- glib2 + gobject-introspection
 	- expat
 	- libdaemon (http://0pointer.de/lennart/projects/libdaemon/)
-	- Linux 2.4 or 2.6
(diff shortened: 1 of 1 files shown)


arrowd:
Can we get this in?

evverx:
Thank you for the PR. I'm conflicted about removing the Linux requirement because avahi is actually tested on Linux only but given that in practice it's built on macOS and so on I think it should be OK to drop it for now.

evverx:
The CI failure at https://github.com/avahi/avahi/runs/40471072772 is obviously unrelated. I opened https://github.com/avahi/avahi/issues/699 so as not forget about this.

PR #690, opened 2025-03-11
gobject: fixed getting warning invalid prop id after set value prop flags

GermanAizek:
@evverx, this PR fixed inconspicuous misprint due to which continues to be executed in `default:` block after execution `case PROP_FLAGS:` block.


Neustradamus:
Dear Avahi team, @lathiat, @evverx, @pemensik,

@GermanAizek has done several PRs, time to look?

PR #691, opened 2025-03-11
core/daemon/gobj: align structures (8 byte) for 64bit platforms

GermanAizek:
@evverx,
Smaller size structure or class, higher chance putting into cpu cache, changes require checking using yours benchmark, I haven't figured out how to run bench in your project yet.

Most processors are already 64 bit, so the change won't make it any worse.

Info about technique:

https://stackoverflow.com/a/20882083
https://zijishi.xyz/post/optimization-technique/learning-to-use-data-alignment/
https://en.wikipedia.org/wiki/Data_structure_alignment

Affected structs:

- AvahiRecordListItem 48 -> 40 bytes
- AvahiClientStates 24 -> 16 bytes
- EntryGroupInfo 56 -> 48 bytes
- _GaEntryGroupPrivate 40 -> 32 bytes
- AvahiRecord 40 -> 32 bytes
- StaticHost 64 -> 56 bytes
- AvahiKey 24 -> 16 bytes
- _GaEntryGroupService 56 -> 48 bytes

### DIFF
--- a/avahi-core/rr.h
+++ b/avahi-core/rr.h
@@ -53,8 +53,8 @@ enum {
     reference counter. The structure is intended to be treated as "immutable", no
     changes should be imposed after creation */
 typedef struct AvahiKey {
-    int ref;           /**< Reference counter */
     char *name;        /**< Record name */
+    int ref;           /**< Reference counter */
     uint16_t clazz;    /**< Record class, one of the AVAHI_DNS_CLASS_xxx constants */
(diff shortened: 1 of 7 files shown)


GermanAizek:
Updated to master branch.

pemensik:
I am not sure we can safely merge these changes. I think they would break API and ABI, if it modifies public structures. I think it should be split into change modifying only struct definitions in c files, which are private. Those can be merged right away. But structures from header used by any client must not change, unless we also increase avahi libraries SO version. I would like to avoid changing ABI now.

Issue #654, opened 2024-11-06
avahi-core/server: incoming_probe: Erroneous conflict detection if TXT data is changed


ceggers-arri:
**Avahi version**
v0.8+x

**Distribution**
Yocto Linux

**Describe the bug**
Sometimes (during automated power-on testing) I get a

`Received conflicting probe [MyService._MyServiceType._tcp.local	IN	TXT "MyAlternatingServiceData"; ttl=4500]. Local host lost. Withdrawing.`

message when I update my service's TXT data (via the EntryGroups's "UpdateServiceTxt" D-Bus call). 

Normally the Avahi daemon is able to detect that an incoming probe was originally sent from the local system. But if the local TXT entry is updated shortly after a previous probe request (containing the previous TXT value) has been sent out, the received echo frame (due to multicast communication) doesn't match the stored value anymore and is detected as a conflict:

```C
static void incoming_probe(AvahiServer *s, AvahiRecord *record, AvahiInterface *i) {
    AvahiEntry *e, *n;
    int ours = 0, won = 0, lost = 0;
    ...

    /* Handle incoming probes and check if they conflict our own probes */

    for (e = avahi_hashmap_lookup(s->entries_by_key, record->key); e; e = n) {
        int cmp;
        n = e->by_key_next;

        if (e->dead)
            continue;

        /* !!! doesn't match ... !!! */
        if ((cmp = avahi_record_lexicographical_compare(e->record, record)) == 0) {
            ours = 1;
            break;
        } else {

            /* !!! this depends on the timing (current state of the record's announcer on this interface) !!! */
            if (avahi_entry_is_probing(s, e, i)) {
                if (cmp > 0)
                    won = 1;
                else /* cmp < 0 */
                    lost = 1;
            }
        }
    }

    if (!ours) {
        char *t = avahi_record_to_string(record);

        if (won)
            avahi_log_debug("Received conflicting probe [%s]. Local host won.", t);
        else if (lost) {
            /* !!! Usually I end up here ... !!! */
            avahi_log_debug("Received conflicting probe [%s]. Local host lost. Withdrawing.", t);
            withdraw_rrset(s, record->key);
        }

        avahi_free(t);
    }
}
```

**To Reproduce**
I can reproduce this only by automated power cycle testing. Additionally it is necessary that the _TXT data_ is updated shortly after the service has been _comitted_.

**Expected behavior**
1. Avahi should detect that the incoming _probe_ frame originated from the local system.
[...]

PR #692, opened 2025-03-13
Avoid conflicting probes after TXT data has been updated

ceggers-arri:
RFC6762, section 8.2, states that own probe packages can erroneously be detected as conflicting probe if the internal copy of the probe message has been updated meanwhile before receiving the echoed message:

> The logic ... is to
> guard against stale probe packets on the network (possibly even stale
> probe packets sent moments ago by this host itself, before some
> configuration change, which may be echoed back after a short delay by
> some Ethernet switches and some 802.11 base stations).

This is exactly the problem I described in #654 some time ago. The current
reaction of Avahi after receiving an _lexicographically earlier_ conflicting probe
is to immediately withdraw the own records (which doesn't match the behavior
requested by the RFC):

> If the host finds that its
> own data is lexicographically earlier, then it defers to the winning
> host by waiting one second, and then begins probing for this record
> again.

If I understand the RFC correctly, the actual withdrawing of our own records
(only in case of a _real_ conflict) shall happen after receiving a conflicting response
packet:

> If the
> winning simultaneous probe was from a real other host on the network,
> then after one second it will have completed its probing, and will
> answer subsequent probes.

I assume that this is already done by calling `handle_conflict()` from `handle_response_packet()`.

### DIFF
--- a/avahi-core/announce.c
+++ b/avahi-core/announce.c
@@ -420,7 +420,7 @@ static void send_goodbye_callback(AvahiInterfaceMonitor *m, AvahiInterface *i, v
     avahi_record_unref(g);
 }
 
-static void reannounce(AvahiAnnouncer *a) {
+static void reannounce(AvahiAnnouncer *a, unsigned msec) {
     AvahiEntry *e;
     struct timeval tv;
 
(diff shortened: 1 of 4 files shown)


pemensik:
This seems nice change, thank you for this attempt.

I have not tested it, but I miss obvious difference in behaviour when the probing is resubmitted already. It seems to me maybe it should store time of last conflict into entry. If the retry after 1s gets the same response and we still conflict, no more retry should be done but we try alternative name. I miss different handling when reannounce were tried recently with the same name.

If the second retry does not conflict anymore, seems like it were just some old packet and we succeed this time. It would just timeout, right? I understand this tries to reprobe only records where the conflict occurred. But how it does ensure we won't reprobe until the winning owner disconnects from the network? e->lost is resetted on incoming probe, cannot inform us this were just initial probe or retried attempt. Do I miss how it should work?

I do not see where did withdraw_rrset(s, record->key); go from original incoming_probe function. If it is not needed anymore, not even withdraw_entry replaces it? Will handle_conflict remaining call be used on reprobe?

This would ideally need relatively simple test. This is long standing issue and not trivial to fix. Verifying this just manually is asking for regressions. We should have some kind of automated test for it, maybe with two containers running on common network? But have no similar test already.

stevenschlansker:
Hi all, I picked this work up as part of #982 - using this PR as partial inspiration, I believe I've fixed the core of this problem, and my work includes a regression test to demonstrate.

Issue #289, opened 2020-05-26
Use of dbm is problematic for cross-compilation


smcv:
The service type database `service-types.db` compiled from `service-type-database/service-types` is installed as a GNU gdbm database, or some other dbm-derived "hash-on-disk" database. Unfortunately, the format of these databases is architecture-specific: it depends on at least endianness and word size, and also appears to differ between some of the architectures supported in Debian even though their endianness and word size is the same, for example amd64 and ppc64el.

This means that if you are cross-compiling - for example, if you have an amd64 build system and you want Avahi binaries for a mips host system - you have to be able to run a mips program to generate a mips-compatible `service-types.db`. In practice this is currently done by running the host system python, for example `python3:mips` in Debian notation. In a cross-compilation scenario, this is not always possible.

There are currently only 107 non-blank, non-comment lines in `service-types`, so perhaps using dbm is overkill, and it would be sufficient to install `service-types` into `${datadir}` and adapt the `avahi-python`, `avahi-ui` and `avahi-utils` modules to parse it line-by-line?

Downstream bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=873596#30

pemensik:
Yes, I agree using any external dbm database backend for just 107 string key to string value database is overkill. Especially when we can store it sorted already and use some kind of faster searching, using sorted data in advance. If we have python building script, we could prepare it by few python functions to be easier consumable. I proposed trivial ini style for this in #721. But because key is expected to be without spaces and special characters, it could be even plain text file with format:

``_example._tcp: Name of service``

The only features we want is getting ideally localised Name of service from simple key. Current code is duplicated once in python in avahi-python/avahi/ServiceTypeDatabase.py.in, second time native C code in avahi-utils/stdb.c. Both very minimalistic implementations.

If we have cache implementation, we should have some searching tree algorithm implemented already. External database is used here to allow python and native use common data. I think it should be done simpler way, yes.

pemensik:
I agree with Helmut on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=873596#41. We could make these services built-in into avahi-browse by generating simple C array. The only question is how to share those types between native and python. If we make it built-in in native, then python would need some way to get that data into python.

I think there should be some way to make it extensible by user. That is why I proposed INI style parsing, which are implemented in both.

pemensik:
If we have it built-in somehow, it should be exposed by some public API too.

Issue #260, opened 2020-01-20
Python3 issue with dbm


pyhalov:
https://github.com/lathiat/avahi/commit/63750f1be96ad08c407193b08bf3b9ee74310e2d moved build-db and ServiceTypeDatabase.py to Python3 dbm, but python 3 dbm will always select gdbm over ndbm. This breaks --disable-gdbm --enable-dbm build.

lathiat:
Hi @pyhalov thanks for the report. Seems we need to try and standardize on DBM in some way. Can you tell me your specific usecase for --disable-gdbm / --enable-dbm ?

pyhalov:
I'm not sure, perhaps, just a bit of history - https://lists.freedesktop.org/archives/avahi/2006-August/000840.html

lathiat:
Are you actually using solaris?

lathiat:
Also does solaris now owrk if you just use the default, e.g. gdbm? I would think for modern solaris this would work.

pyhalov:
No, but I support avahi on OpenIndiana and I'm a bit concerned about consequences (possible incompatibilities) for our users if we rebuild it with gdbm.

lathiat:
Ok perfect thanks I’ll see what I can do here

smcv:
Relatedly, since 63750f1, if Avahi is compiled for Python **2**, the service type database will preferentially be compiled with dbhash (bsddb), which is neither GNU gdbm nor traditional Unix (n)dbm, and can't necessarily be loaded by the other database libraries.

I ran across this because dbhash doesn't support looking up Unicode keys, which avahi-discover relies on (it appears they're transparently encoded to bytes by the other backends). The C code might also fail to look up service types if the database is in an unexpected format.

I think the solution is for this script to force a particular database backend, so that the service type database is always compiled to the expected format. I'll send a PR.

(With hindsight, this database could probably have been a simple text file on disk and a hash table in memory, since it's not exactly large... but that would be an incompatible change now.)

evverx:
Looks like it got worse. Python3.13 switched the default backend to `sqlite3` (https://docs.python.org/3/whatsnew/3.13.html) so `build-db` builds it with the default backend and then, for example, `avahi-browse -b` can't read it.

pemensik:
It seems dbm interface does not even allow us to choose gdbm as preferred format. The problem with this is avahi-browse tries to use the same file too, but does not follow the same changes as python dbm interface.

[...]

PR #698, opened 2025-04-13
Fix issue with dbm.sqlite3 on readonly directories

### LINKED ISSUE #670: Use of dbm causes problems with Python 3.13
**Avahi version**

    extra/avahi  1:0.8+r194+g3f79789-3

**Distribution**

    Arch Linux

**Describe the bug**

Arch now has Python 3.13, and Python 3.13 has switched its default `dbm` backend to SQLite. On the one hand, this would kind of solve #289 since the SQLite file format is architecture-agnostic. On the other hand, unfortunately, it leads to a non-functional avahi-discover:

    $ avahi-discover
    Found service 'frost [00:00:00:00:00:00]' of type '_workstation._tcp' in domain 'local' on 1.0.
    ERROR:dbus.connection:Exception in handler for D-Bus signal:
      File "/usr/lib/python3.13/dbm/sqlite3.py", line 79, in _execute
        return closing(self._cx.execute(*args, **kwargs))
                       ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
    sqlite3.OperationalError: attempt to write a readonly database

...as the new `dbm.sqlite3` module always sets the underlying SQLite database to WAL mode upon creation, and it seems that WAL-mode databases cannot be opened read-only, no matter what.

Since the database only has ~100 entries, it would probably be more robust to have a flat file parsed on startup...

mickael9:
Python dbm.sqlite3 module (which is used by default) sets journal_mode
to WAL when creating the database. Unfortunately this makes the database
unable to be opened again by the dbm.sqlite3 module if it is stored on a
read-only directory.

The behavior is documented here:

https://sqlite.org/wal.html#read_only_databases

> Even though it is possible to open a read-only WAL-mode database, it
> is good practice to converted to PRAGMA journal_mode=DELETE prior to
> burning an SQLite database image onto read-only media.

This change applies the recommendation if sqlite3 is used.

Fixes #670

### DIFF
--- a/service-type-database/build-db
+++ b/service-type-database/build-db
@@ -19,8 +19,10 @@
 
 try:
     import anydbm as dbm
+    from whichdb import whichdb
 except ImportError:
     import dbm
+    from dbm import whichdb
 
 import sys
 
@@ -47,3 +49,19 @@ for ln in open(infn, "r"):
     db[t.strip()] = n.strip()
 
 db.close()
+
+if whichdb(outfn) == "dbm.sqlite3":
+    # Python dbm.sqlite3 module (which is used by default) sets journal_mode
+    # to WAL when creating the database. Unfortunately this makes the database
+    # unable to be opened again by the dbm.sqlite3 module if it is stored on a
+    # read-only directory.
+    #
+    # The behavior is documented here:
+    # https://sqlite.org/wal.html#read_only_databases
+    # > Even though it is possible to open a read-only WAL-mode database, it
+    # > is good practice to converted to PRAGMA journal_mode=DELETE prior to
+    # > burning an SQLite database image onto read-only media.
+    import sqlite3
+    conn = sqlite3.connect(outfn)
+    conn.execute("PRAGMA journal_mode = DELETE;")
+    conn.close()


evverx:
Thank you for the PR!

The avahi build system in its current form shouldn't even try to generate the sqlite database because it isn't actually supported in all the places where it can be read. Currently the build system is aware of `dbm` and `gdbm` only
https://github.com/avahi/avahi/blob/337c94fce7da7659ae6371f6d56a237abcc4a784/configure.ac#L764-L780
https://github.com/avahi/avahi/blob/337c94fce7da7659ae6371f6d56a237abcc4a784/avahi-utils/stdb.c#L21-L27

so it should respect that instead of using the default. It's https://github.com/avahi/avahi/issues/260.

mickael9:
Thanks for the feedback

This is what debian is doing, perhaps you should adopt their patch instead ?

https://sources.debian.org/src/avahi/0.8-16/debian/patches/build-db-Use-the-same-database-format-that-the-C-code-exp.patch/

evverx:
That approach certainly makes sense. As far as I can see it also looks for the python modules so it's more robust (there was a build failure on FreeBSD where `gdbm` was installed without `py311-gdbm` and it all fell apart). I'm not sure why it was decided not to send it upstream though. It could be that it's related to https://github.com/avahi/avahi/issues/289.

pemensik:
These files are ridiculous and unnecessary complicated. The basic should be just key->value dictionary mapping service type into English text description. Translations should be moved into pot template for normal translation. I think if we generated just dictionary into python file from it, it would be better for everyone.

And provide maybe just normal ini file format conf file for possible user customization if needed. This never had enough data to deserve fiddling with database engines. Since the original registration seems finished, I doubt it will ever have more.

We might obtain services from https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xml and parse CSV file to build and updated set. But since this were ever used only from python libraries, I think generating python dict for read-only values would be better idea than fixing this database in future.

pemensik:
Okay, this fixes python tools to read that file, created by python build tool. That fixes python tools only, which is definitely an improvement. But the mess with service types should be fixed even more. But this is an improvement, thank you for that!

evverx:
With this patch applied downstream packages doing the right thing have to revert it and apply their patches. I don't think it makes sense. @pemensik can you revert this?

The latest merges really baffle me. I just revert them downstream.

pemensik:
@evverx This is not complete, but at least partial improvement. Yes, a good fix needs a new feature, new format. You have requested in trivial get hostname feature that should be reverted too. This fixed python tools only. And you wrote above this change makes sense. I am confused what is the alternative. Do we have any competing PR to this one I have missed?

Yes, I think this should be replaced by native function with generated code database. But to use it from python, it needs new API addition. That I thought that is what you refused. This merged would allow at least GUI python tool to have the service types database working, even though ``avahi-browse`` will remain unsupported.

Reverting this won't make it less broken. Less broken needs more drastic change.

evverx:
> You have requested in trivial get hostname feature that should be reverted too

That's a new feature and I don't think `rc-*` tags are suitable for new features. Other than that it doesn't pass basic checks, comes with no tests and invalidates the pot file used by translators. If you think it should go in those things should be addressed.

> I am confused what is the alternative. Do we have any competing PR to this one I have missed?

There are no PRs upstream. There are patches downstream though and they have to be adjusted to apply cleanly by essentially reverting this commit.

> I think this should be replaced by native function with generated code database. But to use it from python, it needs new API addition. That I thought that is what you refused

Changes like that aren't suitable for `rc-*` tags. There is nothing wrong with redesigning the database and it's already discussed in several places.

mickael9:
From my limited understanding, the python code was never meant to use sqlite because the C code won't be able to read it anyway (I don't know what are the implications of that).

If fixing the bug now is desired, I think applying the debian patch would be the best solution as it would make things work as intended and other distributions can drop their patches while a better solution is being worked on

PR #704, opened 2025-05-30
Enable building with Qt6

antlarr:


antlarr:
I tested it builds in [my home project at openSUSE's obs](https://build.opensuse.org/package/show/home:alarrosa:branches:GNOME:Factory/avahi) (if that link doesn't work, it's probably because the change was accepted in the [GNOME:Factory project](https://build.opensuse.org/package/show/GNOME:Factory/avahi) ).

I'm not sure how to fix the build tests here in github for fedora, so some help with that would be great.

nigels-com:
The failing build logs seem expired, can this branch be rebased and fresh checks triggered?

evverx:
I rebased it recently elsewhere. It fails because the CI should be adjusted in various places like https://github.com/avahi/avahi/blob/63859cf4b50e0e123ab6cfa25bfd019c3efe3a9c/.github/workflows/build.sh#L101 to either install and use it on Ubuntu/Alpine/FreeBSD instead of qt5 or disable it explicitly in places like https://github.com/avahi/avahi/blob/63859cf4b50e0e123ab6cfa25bfd019c3efe3a9c/fuzz/oss-fuzz.sh#L75-L77. Packit needs a sed command probably in https://github.com/avahi/avahi/blob/master/.packit.yml adding `Build-Requires` to the spec file until the package is updated downstream

I saw that the patch has been shipped for some time by openSUSE in https://build.opensuse.org/projects/openSUSE:Factory/packages/avahi/files/0001-Enable-building-with-Qt6.patch?expand=1.

Issue #708, opened 2025-06-24
avahi-autoipd ignores RTM_NEWLINK with IFF_LOWER_UP in Linux, failing Bonjour Conformance Test (BCT) 1.5.4


Altair-Bueno:
**Avahi version**

v0.8

**Distribution**

Custom image based of Yocto embedded Linux

**Describe the bug**

During Bonjour Conformance Tests 1.5.4, on `CABLE CHANGE HANDLING`, the test suite tasks us with unpluging in the network cable, waiting some time, and pluging the cable again. As far as known, the software in charge of zeroconf shall send a probe upon link establishment so that BCT can continue the test. See the following log for reference:

```txt
Press enter to continue.
START (CABLE CHANGE HANDLING)
Please unplug the device's network cable for ten seconds and then reconnect.
For a device being tested over its wireless interface, cause the interface to
disassociate from the access point and then re-associate with the access point.


WARNING 2025-06-24 09:28:26.272579+0200: Have not received a probe from the device.  Was it properly disconnected and reconnected to the network?
WARNING 2025-06-24 09:28:59.298794+0200: Have not received a probe from the device.  Was it properly disconnected and reconnected to the network?
```

However, avahi-autoipd does not send a probe unless restarted. Using `avahi-autoipd --refresh eth0` does not send the probe either. 

**To Reproduce**

Follow Bonjour Conformance Test instructions

**Expected behavior**

`avahi-autoipd` should pass Bonjour Conformance Test 1.5.4

**Additional context**

We already run the debugger on `avahi-autoipd` and we found out that this function lacks some sort of mechanism that changes the state back to PROBING.

https://github.com/avahi/avahi/blob/25911be0093c161f6effe9327910efe791912ba5/avahi-autoipd/iface-linux.c#L105

In particular, this function should take into account the following information:

- If we unplug the ethernet cable, we receive the following flags: ['IFF_UP', 'IFF_BROADCAST', 'IFF_MULTICAST']
- IF we plug the ethernet cable, we receive the following flags: ['IFF_UP', 'IFF_BROADCAST', 'IFF_RUNNING', 'IFF_MULTICAST', 'IFF_LOWER_UP']

More information about the meaning of these flags can be found on `netdevice(7)`. We think adding something like this:

```c
extern State state;
if (ifi->ifi_flags & IFF_LOWER_UP) {
  state = STATE_PROBING
}
```

Should suffice, although we haven't tried it yet. Would you be open for a PR that fixes this?

**Alternatives**

[...]

Issue #467, opened 2023-07-14
autoipd: Please support dhcpcd


bdrung:
ISC has [announced](https://www.isc.org/dhcp/) the end of life for ISC DHCP as of the end of 2022. Kea is the recommended successor of the DHCP server. This bug report is about replacing the DHCP client `dhclient`. autoipd ships [dhclient-enter-hook.in](https://github.com/lathiat/avahi/blob/master/avahi-autoipd/dhclient-enter-hook.in) and [dhclient-exit-hook.in](https://github.com/lathiat/avahi/blob/master/avahi-autoipd/dhclient-exit-hook.in) for it.

Please provide similar hooks for [dhcpcd](https://roy.marples.name/projects/dhcpcd) to allow user to use dhcpcd as replacement for dhclient. Debian and Ubuntu want to migrate from dhclient to dhcpcd.

pemensik:
Primary way to handle DHCP on Fedora and RHEL is Network Manager. Alternative is systemd-networkd. dhcpcd is possible as well, but pull request would be welcome.
I happen to be dhcpcd maintainer on Fedora, but I have never used it myself. It would be better if anyone actually using it could test the implementation. I am not against, but not sure how much this is required.

evverx:
I've just taken a look at https://roy.marples.name/projects/dhcpcd and it supports IPv4LL out of the box so there is no need to hook it up to avahi-autoipd in the first place.
```sh
host0: soliciting a DHCP lease                                                                                                                                
host0: sending DISCOVER (xid 0x190c595a), next in 4.8 seconds                                                                                                 
host0: ARP probing 169.254.132.242 (2 of 3), next in 1.3 seconds                                                                                              
host0: Router Advertisement from fe80::34d2:13ff:fed2:dd9e                                                                                                    
host0: executing: /usr/libexec/dhcpcd-run-hooks ROUTERADVERT                                                                                                  
Dropped protocol specifier '.ra' from 'host0.ra'. Using 'host0' (ifindex=2).                                                                                  
host0: ARP probing 169.254.132.242 (3 of 3), next in 2.0 seconds                                                                                              
host0: using IPv4LL address 169.254.132.242           
```
Closing.

pemensik:
I think such support should be done on dhcpcd project, if needed. But I agree, this should be better done not by avahi itself, but whatever network configuration service is used. If dhcpcd can configure it directly, it may have ability to notify us about that, unless existing interface address changes can already catch everything needed.

[...]

PR #709, opened 2025-06-25
fix(avahi-autoipd): Switch state to STATE_START if IFF_LOWER_UP is detected

abueno-airzonesl:
For Bonjour Conformance Test CABLE CHANGE HANDLING it is mandatory that the zeroconf stack sends a probe upon link stablishment.

Fixes: https://github.com/avahi/avahi/issues/708

### DIFF
--- a/avahi-autoipd/main.c
+++ b/avahi-autoipd/main.c
@@ -1066,7 +1066,7 @@ static int loop(int iface, uint32_t addr) {
     int iface_fd = -1;
     Event event = EVENT_NULL;
     int retval_sent = !daemonize;
-    State st;
+    State st = STATE_START;
     FILE *dispatcher = NULL;
     char *address_fn = NULL;
     const char *p;


Altair-Bueno:
I'm happy to announce these changes were enough to pass BCT 1.5.4. We are currently ironing out our implementation of HomeKit, and we expect to pass the certification in the following weeks. At Airzone, are happy to contribute these changes back to the OSS community.

evverx:
avahi-autoipd isn't actively maintained anymore (it was discussed in https://github.com/avahi/avahi/issues/467#issuecomment-2511280295). Can you expand on why it's used instead of the alternatives?

Altair-Bueno:
> avahi-autoipd isn't actively maintained anymore (it was discussed in [#467 (comment)](https://github.com/avahi/avahi/issues/467#issuecomment-2511280295)). Can you expand on why it's used instead of the alternatives?

As pointed by #708, we found that systemd-networkd implementation of zeroconf had more blockers. We couldn't make it pass the BCT, at least the systemd versions we have access to. And in systemd fashion, upgrading versions isn't easy, specially for our embedded Linux.

We also tried mDNSResponder, built from Apple sources. We found that it didn't work well with the rest of our system, and the lack of any sort of documentation makes it impossible (or rather not worth it) to continue further with BCT.

For dhcpd, we already have systemd-networkd as a critical component. And again, in systemd fashion, you either use systemd components or good luck with your Frankenstein.

To our knowledge, there are no other well known alternatives out there for zeroconf support on Linux.

evverx:
> As pointed by https://github.com/avahi/avahi/issues/708, we found that systemd-networkd implementation of zeroconf had more blockers.

It isn't mentioned there what those blockers are.

> We couldn't make it pass the BCT, at least the systemd versions we have access to. 

It should be possible to report it. If there are bugs there they can be fixed there. Unlike avahi-autoipd it's actively maintained, comes with tests and so on.

> For dhcpd, we already have systemd-networkd as a critical component. And again, in systemd fashion, you either use systemd components or good luck with your Frankenstein.

I'm not sure why networkd combined with avahi-autoipd isn't Frankenstein then.

Altair-Bueno:
> It isn't mentioned there what those blockers are.

I didn't list all systemd-networkd blockers on an Avahi issue because they were not relevant to it. In any case, the blockers were

- At least the `PROBING: RATE LIMITING` test fails with systemd-networkd 237, 
  - Fixed in https://github.com/systemd/systemd/issues/14299
- We couldn't upgrade just systemd-networkd without breaking our system
 
> It should be possible to report it. If there are bugs there they can be fixed there. Unlike avahi-autoipd it's actively maintained, comes with tests and so on.

Allegedly, systemd-networkd already passes BCT, so it won't be necessary.

> I'm not sure why networkd combined with avahi-autoipd isn't Frankenstein then.

Avahi autoipd maintains its own separate interfaces from systemd-networkd. It is also a standalone component that doesn't conflict with the rest of the system. On the other hand, dhcpd conflicts with our systemd-networkd which we depend on.

evverx:
> I didn't list all systemd-networkd blockers on an Avahi issue because they were not relevant to it

Agreed. They are relevant in the context of deprecating avahi-autoipd though. As far as I understand if it was possible to update networkd there would be no need to roll out avahi-autoipd in this particular case.

> Avahi autoipd maintains its own separate interfaces from systemd-networkd. It is also a standalone component that doesn't conflict with the rest of the system. On the other hand, dhcpd conflicts with our systemd-networkd which we depend on.

dhcpcd is a standalone component as well. I fuzz it from time to time and for example in https://github.com/NetworkConfiguration/dhcpcd/pull/404#pullrequestreview-2526952064 and https://github.com/NetworkConfiguration/dhcpcd/issues/415 the interface where the packets were received was managed by dhcpcd and all the other interfaces were managed by networkd.

Altair-Bueno:
> Agreed. They are relevant in the context of deprecating avahi-autoipd though. As far as I understand if it was possible to update networkd there would be no need to roll out avahi-autoipd in this particular case.

Yes, it should pass BCT. 

> dhcpcd is a standalone component as well. I fuzz it from time to time and for example in https://github.com/NetworkConfiguration/dhcpcd/pull/404#pullrequestreview-2526952064 and https://github.com/NetworkConfiguration/dhcpcd/issues/415 the interface where the packets were received was managed by dhcpcd and all the other interfaces were managed by networkd.

We'll give it another shoot using the links you provided to see if we manage to get both running. However, we still would prefer to use `avahi-dhcpd` as it's much smaller than `dhcpd`: 206.1KB against 32KB without including the `udev` dynamic library needed for `dhcpd`! That's massive for our platform.

<details>

<summary>Build command</summary>

```sh
. path/to/cross/toolchain
# We even set LTO!!
export CFLAGS+=" -flto -Os" LDFLAGS+=" -flto"
./configure ${CONFIGURE_FLAGS:-} \
  --small  --disable-auth \
  --disable-arp --disable-arping \
  --disable-dhcp6  --disable-privsep
DESTDIR="$PWD/rootfs" make -j
${STRIP:-strip} rootfs/**/*
```

We are using these flags https://github.com/NetworkConfiguration/dhcpcd/blob/master/BUILDING.md

</details>

> They are relevant in the context of deprecating avahi-autoipd though

As a side note, if `avahi-autoipd` gets deprecated and critical fixes like this one do not get through, the Avahi project should drop the "compatible with Bonjour" slogan. One could argue that the software is compatible and not certified, but without `avahi-autoipd` (and a fix for this test) or similar software, zeroconf won't work correctly. I think that reason is enough to either stop the deprecation or at least give some sort of guidelines on how to set up zeroconf in conjunction with third party software. It may be wise to create a separate issue on this matter.

evverx:
> it's much smaller than dhcpd: 206.1KB against 32KB 

I guess the size would be the only downside but in general given that these days networkd, NetworkManager and dhcpcd support IPv4LL out of the box and are used to manage network interfaces anyway it would be better to trim those 32KB.

I agree that all the places where it's implied that IPv4LL is implemented should be changed (though I'm not sure it should turn into a tutorial on how to set up network interfaces. It would probably be better to point to the official documentation of the other projects or something like that).

Neustradamus:
@abueno-airzonesl: Nice PR!

PR #719, opened 2025-10-11
Pemensik cname additions

pemensik:
My additions on top of #682. Improves new test a bit to run simpler way.

### DIFF
--- a/avahi-core/cname-test.c
+++ b/avahi-core/cname-test.c
@@ -49,9 +49,9 @@ static void quit_timeout_callback(AVAHI_GCC_UNUSED AvahiTimeout *timeout, void *
 }
 
 typedef struct {
-	const char *name;
-	void (*test_case)(void);
-	AvahiResolverEvent expected;
+    const char *name;
+    void (*test_case)(void);
+    AvahiResolverEvent expected;
 } test_cases_db_t;
 
 #define avahi_test_cache_flush() avahi_cache_flush(cache)
@@ -173,13 +173,13 @@ static void cname_toomuch(void) {
 
     avahi_test_add_cname("X.local", "C1.local");
     for (i=1; i<25; i++) {
-	snprintf(from, sizeof(from), "C%u.local", i);
-	snprintf(to, sizeof(to), "C%u.local", i+1);
-	avahi_test_add_cname(from, to);
+        snprintf(from, sizeof(from), "C%u.local", i);
+        snprintf(to, sizeof(to), "C%u.local", i+1);
+        avahi_test_add_cname(from, to);
     }
     snprintf(from, sizeof(from), "C%u.local", i);
-    avahi_test_add_cname(from, "Y.local");
-    avahi_test_add_a("Y.local", "192.168.50.99", 2);
+    avahi_test_add_cname(from, "A.local");
+    avahi_test_add_a("A.local", "192.168.50.99", 2);
 }
 
 static void server_callback(AvahiServer *s, AvahiServerState state, AVAHI_GCC_UNUSED void *userdata) {
@@ -248,7 +248,7 @@ static void test_list(void) {
     const test_cases_db_t *c;
 
     for (c=test_cases; c->name; c++) {
-	    printf("%s\n", c->name);
+        printf("%s\n", c->name);
     }
 }
 
@@ -258,9 +258,9 @@ static const test_cases_db_t * find_test(const char *test_case) {
     assert(test_case);
 
     for (c=test_cases; c->name; c++) {
-	if (!strcmp(test_case, c->name)) {
-	    return c;
-	}
+        if (!strcmp(test_case, c->name)) {
+            return c;
+        }
     }
     return NULL;
 }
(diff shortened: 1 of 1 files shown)


pemensik:
We want cmocka coded tests instead. But at least make the test similar to normal testing now, try to guess success or failure and whether is expected.

siteshwar:
OpenScanHub reports several [warnings](https://openscanhub.fedoraproject.org/task/75221/log/added.html) in the tests code.

avahi-core/cname-test.c
```diff
@@ -197,46 +225,61 @@ static void hnr_callback(
...
+
+    for (c=test_cases; c->name; c++) {
+	    printf("%s\n", c->name);
+    }
 }
 
-#define CHECK_TEST_CASE(test_name, function) \
-    if (!strcmp(test_case, test_name)) \
-        avahi_test_case_function = &function;
+static const test_cases_db_t * avahi_test_initialize(const char *test_case) {
```
friendlyhugo: I think `avahi_find_test` would be a more suitable name, since it just finds the correct test case now.

avahi-core/cname-test.c
```diff
@@ -259,14 +302,56 @@ int main(int argc, char *argv[]) {
...
+    const test_cases_db_t *c;
+    int r = 0;
+    int i = 0;
+
+    for (c=test_cases; c->name; c++) {
+	    i++;
+	    if (run_single(c) != 0)
+		r++;
+    }
+    avahi_log_info("# All %d test cases finished, status: %d", i, r);
```
friendlyhugo: Printing `failed tests: %d` instead of `status: %d` would be more informative, since `r` is now the number of failed tests.

avahi-core/cname-test.c
```diff
@@ -259,14 +302,56 @@ int main(int argc, char *argv[]) {
...
+    return r;
+}
+
+
+int main(int argc, char *argv[]) {
+    const test_cases_db_t *test;
+
+    if (argc < 2) {
+        printf("Usage: %s [--list] [test_name]\n", argv[0]);
```
friendlyhugo: Usage is always printed, even when attempting to run all test cases.

avahi-core/cname-test.c
```diff
@@ -81,6 +87,11 @@ static void avahi_test_add_cname(const char *src, const char *dst) {
     avahi_log_debug("Added CNAME record to cache: %s -> %s", src, dst);
 }
 
+static const char *myself(void) {
```
friendlyhugo: CI is failing due to `myself` not being resolved to any IP. It looks like Avahi doesn't record its own hostname, so a new A record to `myself` should be added, just like in the original testcases.

avahi-core/cname-test.c
```diff
@@ -259,14 +302,56 @@ int main(int argc, char *argv[]) {
...
+
+    if (!strcmp(argv[1], "--list")) {
+	test_list();
+	return 0;
+    }
+    test = avahi_test_initialize(argv[1]);
+    if (test) {
+        return run_single(test);
+    } else {
+	printf("Test not found: %s\n", argv[1]);
```
friendlyhugo: Maybe run `clang-format` on this file? Indentation looks wrong in multiple places.

avahi-core/cname-test.c
pemensik: yes, I left it there, because it does not implement any --help or -h. I did not want to make proper argument parsing at the moment.

avahi-core/cname-test.c
pemensik: ah, it worked on my machine, where also avahi-daemon from system were running. I thought it would export it from the test as well, but did not try it in a separate namespace. Yes, we lack decent running of tests in a reproducible and reliable manner. Relying on already running avahi is certainly bad idea.

avahi-core/cname-test.c
pemensik: sure, but will omit avahi_ prefix. It is static function anyway.

avahi-core/cname-test.c
pemensik: good idea!

avahi-core/cname-test.c
pemensik: Ok, let's drop myself function and insert just A.local as final host.

avahi-core/cname-test.c
pemensik: That is likely caused by mixed tabs and spaces. Maybe you have different settings than me, I see it ok. avahi does not have any style definition for clang-format. I have never defined that. Default styles I have tried looked different. It would be great to define .clang-format file and add it into repository. Something similar to current style ideally, maybe refining it a bit? I think it should reformat all files once we define it. But I think code formatting is nice for future, but I would remain at somehow sane manual formatting for now.

friendlyhugo:
Looks like CI is failing again, seems to be caused by `server = NULL` on line 196. After removing it the tests passed on my machine.

evverx:
As far as I can see it conflicts with the master branch (where the NULL thing was fixed), doesn't change anything substantially and got stalled overall. Closing.

It would be great to improve the test by exercising more codepaths (for example it strategically uses AVAHI_LOOKUP_USE_MULTICAST which prevents it from discovering bugs in `self_loop` because the loop detection doesn't fully work without https://github.com/avahi/avahi/pull/729 but it can be discussed in that PR)

PR #720, opened 2025-10-11
Get hostname from avahi-set-host-name tool

pemensik:
Provide simple way of getting current Avahi hostname from shell. It might have changed because conflict, make it simple to check it.

### DIFF
--- a/avahi-utils/avahi-set-host-name.c
+++ b/avahi-utils/avahi-set-host-name.c
@@ -116,7 +116,10 @@ static int parse_command_line(Config *c, int argc, char *argv[]) {
     }
 
     if (c->command == COMMAND_UNSPEC) {
-        if (optind != argc-1) {
+	if (argc == 1) {
+	    c->command = COMMAND_GET;
+	    return 0;
+	} else if (optind != argc-1) {
             fprintf(stderr, _("Invalid number of arguments, expecting exactly one.\n"));
             return -1;
         }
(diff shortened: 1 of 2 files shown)


evverx:
@pemensik can you revert this? I don't think `rc-*` tags are suitable for new features.

If you think that it should go in at this point can you reformat your changes properly, add tests running the command and regenerate the `pot` file used by translators?

evverx:
The links to the coding style can be found in https://github.com/avahi/avahi/issues/722 by the way.

> No tabs please! Please indent exclusively with spaces.
> Indentation is 4 characters.

pemensik:
I found multiple times when testing something, that obtaining current avahi name needs to be simpler. I want this tiny feature in 0.9 release. But regenerating template might be a good idea. I have to find how to do that. My system complains about gettext version mismatch on po/Makefile.in.in file.

evverx:
It would still be necessary to wait for the translators to translate that to sync https://github.com/avahi/avahi/pull/657 with that and that would delay the release even further.

> I found multiple times when testing something, that obtaining current avahi name needs to be simpler

On Fedora it's possible to run `systemctl status avahi-daemon` to see the hostname avahi advertises. It should show something like
```sh
...
     CGroup: /system.slice/avahi-daemon.service
             ├─1208829 "avahi-daemon: running [my-hostname.local]"
             └─1208830 "avahi-daemon: chroot helper"
```

Either way there is absolutely no need to rush things like that. New features can be added once the release is cut.

evverx:
I opened https://github.com/avahi/avahi/pull/720. It's a draft for now. Either the issues should be addressed or it should be reverted.

evverx:
@pemensik https://github.com/avahi/avahi/pull/657 is out of sync with the avahi repository now and it shouldn't happen before the release. Are you planning to get it in sync in time?

Issue #684, opened 2025-02-27
avahi-browse: reachable assertion in service_type_browser_callback


evverx:
It's triggered by CNAMEs pointing `_services._dns-sd._udp.local` to `_services._dns-sd._udp.local`
```sh
avahi-browse: avahi-browse.c:100: void check_terminate(Config *): Assertion `n_all_for_now >= 0' failed.
```
```sh
#0  0x00007ffff7bd8114 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff7b7ef9e in raise () from /lib64/libc.so.6
#2  0x00007ffff7b66942 in abort () from /lib64/libc.so.6
#3  0x00007ffff7b6685e in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff7b770c7 in __assert_fail () from /lib64/libc.so.6
#5  0x00000000004e9148 in check_terminate (c=0x7ffff5b00030) at avahi-browse.c:100
#6  0x00000000004ea6a6 in service_type_browser_callback (b=0x506000001700, interface=-1, protocol=-1, event=AVAHI_BROWSER_ALL_FOR_NOW, type=0x0, domain=0x0, 
    flags=0, userdata=0x7ffff5b00030) at avahi-browse.c:425
#7  0x00007ffff7f6f12d in avahi_service_type_browser_event (client=0x50e000000200, event=AVAHI_BROWSER_ALL_FOR_NOW, message=0x510000000840) at browser.c:555
#8  0x00007ffff7f5b8fa in filter_func (bus=0x512000000040, message=0x510000000840, userdata=0x50e000000200) at client.c:240
#9  0x00007ffff7f0c081 in dbus_connection_dispatch () from /lib64/libdbus-1.so.3
#10 0x00007ffff7f7fcb3 in dispatch_timeout_callback (t=0x5060000011c0, userdata=0x503000001450) at ../avahi-common/dbus-watch-glue.c:105
#11 0x00007ffff7eb60d9 in timeout_callback (t=0x5060000011c0) at simple-watch.c:447
#12 0x00007ffff7eb4f9d in avahi_simple_poll_dispatch (s=0x50e000000120) at simple-watch.c:563
#13 0x00007ffff7eb614b in avahi_simple_poll_iterate (s=0x50e000000120, timeout=-1) at simple-watch.c:605
#14 0x00007ffff7eb639b in avahi_simple_poll_loop (s=0x50e000000120) at simple-watch.c:646
[...]

PR #729, opened 2025-10-29
Fix multiple issues in `lookup_drop_cname` (supersedes #286)

friendlyhugo:
This PR supersedes #286 (as it cherry-picks @szenker's and @pemensik's commits).

This PR fixes a couple of issues with `lookup_drop_cname`, which cause lookups to be retained even after their corresponding CNAME records are removed. It also corrects the lookup flags comparison in `lookup_drop_cname` which never matched the `flags` value.

At the time, #286 was blocked on #551 being merged - #682, which superseded #551 was merged a couple weeks back.

Thanks to @szenker and @pemensik for their work on #286!

### DIFF
--- a/avahi-core/browse.c
+++ b/avahi-core/browse.c
@@ -36,8 +36,6 @@
 #include "domain-util.h"
 #include "rr-util.h"
 
-#define AVAHI_LOOKUPS_PER_BROWSER_MAX 15
-
 struct AvahiSRBLookup {
     AvahiSRecordBrowser *record_browser;
 
--- a/avahi-core/browse.c
+++ b/avahi-core/browse.c
@@ -153,6 +153,19 @@ static AvahiSRBLookup* lookup_ref(AvahiSRBLookup *l) {
     return l;
 }
 
+static int avahi_lookup_match(const AvahiSRBLookup *l, AvahiIfIndex interface, AvahiProtocol protocol, AvahiLookupFlags flags, const AvahiKey *key) {
+    if (l->interface != AVAHI_IF_UNSPEC && l->interface != interface)
+        return 0;
+
+    if (l->protocol != AVAHI_PROTO_UNSPEC && l->protocol != protocol)
+        return 0;
+
+    if (flags != 0 && l->flags != flags)
+        return 0;
+
+    return avahi_key_equal(l->key, key);
+}
+
 static AvahiSRBLookup *lookup_find(
     AvahiSRecordBrowser *b,
     AvahiIfIndex interface,
@@ -166,11 +179,7 @@ static AvahiSRBLookup *lookup_find(
 
     for (l = b->lookups; l; l = l->lookups_next) {
 
-        if ((l->interface == AVAHI_IF_UNSPEC || l->interface == interface) &&
-            (l->protocol == AVAHI_PROTO_UNSPEC || l->protocol == protocol) &&
-            l->flags == flags &&
-            avahi_key_equal(l->key, key))
-
+        if (avahi_lookup_match(l, interface, protocol, flags, key))
             return l;
     }
 
@@ -290,7 +299,7 @@ static void lookup_multicast_callback(
             if (r->key->clazz == AVAHI_DNS_CLASS_IN &&
                 r->key->type == AVAHI_DNS_TYPE_CNAME)
                 /* It's a CNAME record, so let's drop that query! */
-                lookup_drop_cname(l, interface, protocol, 0, r);
+                lookup_drop_cname(l, interface, protocol, b->flags, r);
             else {
                 /* It's a normal record, so let's call the user callback */
 
@@ -478,10 +487,7 @@ static void lookup_drop_cname(AvahiSRBLookup *l, AvahiIfIndex interface, AvahiPr
 
         assert(n);
 
-        if ((n->interface == AVAHI_IF_UNSPEC || n->interface == interface) &&
-            (n->protocol == AVAHI_PROTO_UNSPEC || n->protocol == protocol) &&
-            n->flags == flags &&
-            avahi_key_equal(n->key, k))
+        if (avahi_lookup_match(n, interface, protocol, flags, k))
             break;
     }
 
--- a/avahi-core/browse.c
+++ b/avahi-core/browse.c
@@ -167,7 +167,7 @@ static AvahiSRBLookup *lookup_find(
     for (l = b->lookups; l; l = l->lookups_next) {
 
         if ((l->interface == AVAHI_IF_UNSPEC || l->interface == interface) &&
-            (l->interface == AVAHI_PROTO_UNSPEC || l->protocol == protocol) &&
+            (l->protocol == AVAHI_PROTO_UNSPEC || l->protocol == protocol) &&
             l->flags == flags &&
             avahi_key_equal(l->key, key))
 
@@ -479,7 +479,7 @@ static void lookup_drop_cname(AvahiSRBLookup *l, AvahiIfIndex interface, AvahiPr
         assert(n);
 
         if ((n->interface == AVAHI_IF_UNSPEC || n->interface == interface) &&
-            (n->interface == AVAHI_PROTO_UNSPEC || n->protocol == protocol) &&
+            (n->protocol == AVAHI_PROTO_UNSPEC || n->protocol == protocol) &&
             n->flags == flags &&
             avahi_key_equal(n->key, k))
             break;
--- a/.github/workflows/smoke-tests.sh
+++ b/.github/workflows/smoke-tests.sh
@@ -228,8 +228,9 @@ fi
 run ./avahi-core/avahi-test
 run ./avahi-core/querier-test
 
-for test_case in self_loop retransmit_cname one_normal one_loop two_normal two_loop two_loop_inner two_loop_inner2 three_normal three_loop diamond cname_answer_diamond cname_answer; do
-    run ./avahi-core/cname-test $test_case
+for test_case in self_loop self_loop_answer retransmit_cname one_normal one_loop two_normal two_loop two_loop_inner two_loop_inner2 three_normal three_loop diamond cname_answer_diamond cname_answer cname_answer_removed cname_max_reached cname_answer_triangle_removed; do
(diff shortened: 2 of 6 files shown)


evverx:
The third commit just says
```
avahi-core: fix flags comparison in lookup_find and lookup_drop_cname
Closes https://github.com/avahi/avahi/pull/286".
```
Can you expand on what it fixes? One use case where avahi did the wrong thing with the flags should suffice.

avahi-core/browse.c
```diff
@@ -155,6 +155,19 @@ static AvahiSRBLookup* lookup_ref(AvahiSRBLookup *l) {
     return l;
 }
 
+static int avahi_lookup_match(AvahiSRBLookup *l, AvahiIfIndex interface, AvahiProtocol protocol, AvahiLookupFlags flags, AvahiKey *key) {
```
siteshwar: Please review these [warnings](https://openscanhub.fedoraproject.org/task/78960/log/added.html) reported by OpenScanHub:
```
Error: CPPCHECK_WARNING ([CWE-398](https://cwe.mitre.org/data/definitions/398.html)): [[#def1]](https://openscanhub.fedoraproject.org/task/78960/log/added.html#def1)
avahi-0.9.rc2/avahi-core/browse.c:158: style[constParameterPointer]: Parameter 'key' can be declared as pointer to const
#  156|   }
#  157|   
#  158|-> static int avahi_lookup_match(AvahiSRBLookup *l, AvahiIfIndex interface, AvahiProtocol protocol, AvahiLookupFlags flags, AvahiKey *key) {
#  159|       if (l->interface != AVAHI_IF_UNSPEC && l->interface != interface)
#  160|           return 0;

Error: CPPCHECK_WARNING ([CWE-398](https://cwe.mitre.org/data/definitions/398.html)): [[#def2]](https://openscanhub.fedoraproject.org/task/78960/log/added.html#def2)
avahi-0.9.rc2/avahi-core/browse.c:158: style[constParameterPointer]: Parameter 'l' can be declared as pointer to const
#  156|   }
#  157|   
#  158|-> static int avahi_lookup_match(AvahiSRBLookup *l, AvahiIfIndex interface, AvahiProtocol protocol, AvahiLookupFlags flags, AvahiKey *key) {
#  159|       if (l->interface != AVAHI_IF_UNSPEC && l->interface != interface)
#  160|           return 0;
```

friendlyhugo:
> Can you expand on what it fixes? One use case where avahi did the wrong thing with the flags should suffice.

I experienced this bug when I used CNAME records with a low TTL. When said TTL expires before the record is resolved, avahi keeps sending requests and even handles them if they are answered.

I've added a more technical explanation in the commit message.

avahi-core/browse.c
friendlyhugo: Fixed.

friendlyhugo:
@evverx, I updated the commit message. WDYT?

friendlyhugo:
I've opened #797 to track this.

evverx:
> I've opened https://github.com/avahi/avahi/issues/797 to track this.

Let's keep discussing it there. I'll add the needs-reporter-feedback label here too.

evverx:
While the thing where avahi keeps sending queries after CNAMEs expire is being discussed can you update the commit messages of the other commits and expand on what exactly they fix? Basically it should be possible to include that in the release notes (or whatever) to make it clear what changes. Things like

> [Fix another place with the same issue](https://github.com/avahi/avahi/pull/729/commits/c4bf836bfa043a630b8dc0aeff3dd790455f2652)
> https://github.com/friendlyhugo/avahi/commit/6d40075c8d258f01637ceb8fae07b4d64b19f64a have not noticed the same thing
is repeated multiple times. Fix that too.

aren't clear enough.

friendlyhugo:
I updated the commit messages, but kept the original authors of #286 for proper credit.

evverx:
> I updated the commit messages, but kept the original authors of https://github.com/avahi/avahi/pull/286 for proper credit.

Makes sense (though I don't think those commits make sense separately so they should probably be turned into one commit with a bunch of `Co-authored-by` but those details can be ironed out later)

evverx:
Looking at the commit messages they just say what they do. Can you add stuff like "prevents avahi from sending queries when CNAMES expire" to clarify specific issues they address?

evverx:
It turns out with your patch applied the DoS mentioned in https://github.com/avahi/avahi/issues/684 is no longer reproducible so it affects the events positively in that particular case (though I haven't looked at that closely). With that I'll add the "important" label. Either way can you update the commit messages?

friendlyhugo:
> It turns out with your patch applied the DoS mentioned in #684 is no longer reproducible

Great, I added a "Fixes" line to the commit message.

> Either way can you update the commit messages?

I updated them and squashed the two commits from #286.

evverx:
I'll fetch this PR next week to run some tests to see whether anything else blows up. As mentioned in https://github.com/avahi/avahi/issues/797 I'm concerned about the events related to CNAMEs and how they can affect the existing clients. avahi-browse is fine but it's not the only client consuming those events. I'll add it to the 0.9 milestone in the meantime.

evverx:
Actually it's still possible to mess up the events and crash avahi-browse by sending another CNAME so "Fixes https://github.com/avahi/avahi/issues/684" can be removed. I'll remove the important label.

evverx:
Looking at the following commit message
```
avahi-core: fix flags comparison in lookup_find and lookup_drop_cname

Previously avahi would only accept flags if they matched l->flags exactly.
This never actually happened, as flags was always 0 and l->flags is always
non-0 (initialized in lookup_new using transport_flags_from_domain).
```
it makes it look like `lookup_find` and `lookup_drop_cname` are equally affected and that's not accurate. Can you adjust that part?

Also why do you keep `flags != 0` in `avahi_lookup_match`?

friendlyhugo:
> why do you keep `flags != 0` in `avahi_lookup_match`?

This is required in case `b->flags` is initialized to `0` by a client (`0` means "any transport").

Do you think we should close #797 for now? Let me know if you wanna keep working on this.

evverx:
> This is required in case b->flags is initialized to 0 by a client (0 means "any transport").

That's correct. If `avahi-browse` used `AVAHI_LOOKUP_USE_MULTICAST` instead of `0` it wouldn't crash in https://github.com/avahi/avahi/issues/684 with that particular CNAME. What I was trying to say is that the commit message currently focuses on the thing where lookups aren't removed correctly but it doesn't say that they aren't added correctly either in some cases preventing the loop detection added in https://github.com/avahi/avahi/pull/682 from working.

> Do you think we should close https://github.com/avahi/avahi/issues/797 for now? Let me know if you wanna keep working on this

I don't think it should be closed. I think the cname test need tweaking to cover various protocols, flags and so on.

friendlyhugo:
I updated the commit message to include the changes to `lookup_find`.

> I think the cname test need tweaking to cover various protocols, flags and so on.

Anything apart from running the test with `flags=0`? Should I add it to this branch or open a new PR?

evverx:
> Anything apart from running the test with flags=0?

The test doesn't actually fail when it's run with `flags=0`. It just emits
```
browse.c: Failed to create SRBLookup.
browse.c: Failed to create SRBLookup.
```
and keeps going so in its current form it can't catch that. I'd need to think about how that part can be covered better.

> Should I add it to this branch or open a new PR?

I think the tests should go along with the code to make sure it actually works so it would be better to add those things here.

friendlyhugo:
I decided to add a new test to ensure `lookup_find` functions correctly. It essentially checks whether the final lookup has been added, which, as you've mentioned, would fail before this change.

evverx:
> It essentially checks whether the final lookup has been added

I have to say it's a really creative way to make sure that the CNAME pointing to itself isn't added a gazillion times.

I've been looking at 78eab31128479f06e30beb8c1cbf99dd921e2524 and wondering why `lookup_exists_in_path` is needed at all. If `avahi_lookup_match` can look for lookups correctly now `lookup_handle_cname` can detect loops by looking for lookups before adding them. If they already exists it's a loop. Or is there a reason that it can't be handled that way?

evverx:
Also with the new test added the code path where `AVAHI_LOOKUPS_PER_BROWSER_MAX` is hit is no longer exercised. I think it should be covered as well by adding a bunch of different CNAMEs or something like that.

friendlyhugo:
I encountered a weird crash on this branch. I'm looking into it, converting to draft for now.

evverx:
What's about https://github.com/avahi/avahi/commit/78eab31128479f06e30beb8c1cbf99dd921e2524 though? It looks convoluted for no reason. I'm planning to revert https://github.com/avahi/avahi/pull/682 and then reintroduce those things one by one with proper commit messages and links to advisories.

friendlyhugo:
> Also with the new test added the code path where AVAHI_LOOKUPS_PER_BROWSER_MAX is hit is no longer exercised.

I added a test for that.

> Or is there a reason that it can't be handled that way?

I tried implementing this idea (along with similar ones) a while back, when I worked on #682. From what I can recall, I ran into several edge cases that all of these simpler algorithms didn't handle well. For example, I added the `cname_answer_triangle_removed` test which the algorithm you suggested will incorrectly detect as a loop, which results in the entire lookup failing.

If you'd like, I can add some more comments in `lookup_exists_in_path` which will clarify it.

> I'm planning to revert #682 and then reintroduce those things one by one with proper commit messages and links to advisories.

Do you want me to reword the messages for #682?

evverx:
Why do you think `cname_answer_triangle_removed` is correct? mDNSResponder isn't going to resolve that for example.

evverx:
> Do you want me to reword the messages for https://github.com/avahi/avahi/pull/682

Generally the idea is to revert the PR, reopen the issues and then address them in separate PRs with proper code review and things like that (and that would include rewording the commit messages). Let's discuss this in 2026 though.

evverx:
> Why do you think cname_answer_triangle_removed is correct? mDNSResponder isn't going to resolve that for example.

Just to be absolutely precise here mDNSResponder resolves that but it does that because (unlike avahi in this PR) it doesn't remove CNAMEs that expire. It doesn't actually follow the `X` -> `Z` -> `Y` chain.

PR #731, opened 2025-10-29
Revert "Get hostname from avahi-set-host-name tool"

evverx:
Reverts avahi/avahi#720

It's a draft for now. If the issues aren't addressed it should be reverted.

### DIFF
--- a/avahi-utils/avahi-set-host-name.c
+++ b/avahi-utils/avahi-set-host-name.c
@@ -42,8 +42,7 @@
 typedef enum {
     COMMAND_UNSPEC,
     COMMAND_HELP,
-    COMMAND_VERSION,
-    COMMAND_GET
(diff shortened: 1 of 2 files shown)


pemensik:
Do we know anything this change has broken? I am not sure we have the code base polished enough to be so strict. But I am open to opinions.

evverx:
> Do we know anything this change has broken?

https://github.com/avahi/avahi/pull/657

> I am not sure we have the code base polished enough to be so strict

It doesn't mean it should be made even sloppier.

I'll go ahead and merge this. If you are planning to bring it back please address all the things mentioned in https://github.com/avahi/avahi/pull/720 and mention how you are planning to get https://github.com/avahi/avahi/pull/657 in the releasable state again.

Issue #724, opened 2025-10-27
avahi and FreeBSD


evverx:
Currently it isn't tested on FreeBSD at all, there are patches in https://github.com/freebsd/freebsd-ports/tree/main/net/avahi-app that should be integrated upstream and so on.

The first step would be to setup the CI but the latest attempt to run the CI introduced in https://github.com/avahi/avahi/pull/676 wasn't particularly successfull:
```sh
libtool: link: clang -I.. "-DDEBUG_TRAP=__asm__(\"int \$3\")" -DAVAHI_DAEMON_RUNTIME_DIR=\"/run/avahi-daemon/\" -DAVAHI_SOCKET=\"/run/avahi-daemon/socket\" -DAVAHI_SERVICE_DIR=\"/etc/avahi/services\" -DAVAHI_CONFIG_FILE=\"/etc/avahi/avahi-daemon.conf\" -DAVAHI_HOSTS_FILE=\"/etc/avahi/hosts\" -DAVAHI_DBUS_INTROSPECTION_DIR=\"/usr/local/share/dbus-1/interfaces\" -DAVAHI_CONFIG_DIR=\"/etc/avahi\" -D_REENTRANT -I/usr/local/include -DUSE_EXPAT_H -I/usr/local/libdata/pkgconfig/../../include/dbus-1.0 -I/usr/local/libdata/pkgconfig/../../lib/dbus-1.0/include -DDBUS_VERSION_MAJOR=1 -DDBUS_VERSION_MINOR=16 -DDBUS_VERSION_MICRO=2 -DDBUS_API_SUBJECT_TO_CHANGE -DDBUS_SYSTEM_BUS_DEFAULT_ADDRESS=\"unix:path=/var/run/dbus/system_bus_socket\" -I/usr/local/include -g -O0 -fstack-protector -std=c99 -Wall -W -Wextra -pedantic -pipe -Wformat -Wold-style-definition -Wdeclaration-after-statement -Wfloat-equal -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wshadow -Wendif-labels
ld: error: undefined symbol: inotify_init
>>> referenced by main.c:1231
>>>               avahi_daemon-main.o:(run_server)
ld: error: undefined symbol: inotify_add_watch
>>> referenced by main.c:948
>>>               avahi_daemon-main.o:(add_inotify_watches)
>>> referenced by main.c:953
[...]

PR #723, opened 2025-10-19
Debian dbm fixes

pemensik:
Upgrades #698, solves #270 in debian way. Includes my additions.

Source of first patch is Debian @smcv: https://sources.debian.org/src/avahi/0.8-16/debian/patches/build-db-Use-the-same-database-format-that-the-C-code-exp.patch/

PR #732, opened 2025-10-30
Get rid of bootstrap.sh and use autogen.sh instead

arrowd:


evverx:
Thank you for the PR.

Can you expand on why it's removed? I'm guessing it's because of https://github.com/avahi/avahi/issues/724#issuecomment-3460265399 but it would be great if it was mentioned how `make distcheck` failed or something like that.

`bootstrap.sh` added `-g -O0` to `CFLAGS` on the CI. Without it it should be done elsewhere.

arrowd:
> Can you expand on why it's removed? I'm guessing it's because of https://github.com/avahi/avahi/issues/724#issuecomment-3460265399 but it would be great if it was mentioned how make distcheck failed or something like that.

The main reason is that it is simply useless. In introduces an alternative entry point for building the project in addition to `./autogen.sh`, hardcodes a lot of host info and does not bring anything to the table.

> bootstrap.sh added -g -O0 to CFLAGS on the CI. Without it it should be done elsewhere.

Right, I'll fix that.

evverx:
> The main reason is that it is simply useless. In introduces an alternative entry point for building the project in addition to ./autogen.sh, hardcodes a lot of host info and does not bring anything to the table.

I don't think this rationale is convincing enough in terms of removing it completely. It isn't useful for packages (because it adds the things useful for testing/development only like `-g -O0` among other things) but I don't think it was supposed to be used by packages. Personally I'm fine with removing it completely but the commit message should say why it's removed (and it isn't enough to say that it's useless) and it should also say why it's unlikely to break anything useful downstream somewhere.

evverx:
Since `bootstrap.sh` is kept I think it would be better to keep using it in the coverity workflow to make sure the script doesn't obviously fail. Other than that it looks good to me.

(Just to clarify I was OK with removing it completely. I just thought that the commit message wasn't descriptive enough)

evverx:
Thank you! I'll merge it once the CI is green.

FWIW if `make check` already works on FreeBSD it would be great to integrate it into the CI. There are PRs like https://github.com/avahi/avahi/pull/723 affecting the build system and it would be great to exercise them on FreeBSD automatically. `make distcheck` can be introduced separately/in parallel.

PR #733, opened 2025-10-30
Second try on FreeBSD CI

arrowd:

PR #741, opened 2025-11-05
FreeBSD CI: actually remove the avahi dep

arrowd:
This is a follow-up to #733


arrowd:
Nice catch, thanks @evverx ! It is correctly removed now:

> [1/1] Deleting files for avahi-app-0.8_6: .......... done

evverx:
Thanks! I'll merge it once the CI is green.

@arrowd is there any chance you could take a look at https://github.com/dbus-fuzzer/dfuzzer/pull/163? It's dfuzzer but the things discussed there are relevant in the context of the avahi smoke tests as well. They are likely to be decoupled from systemd similarly to what dfuzzer does.

(I'm slowly porting the test poking the simple protocol (it's now completely broken because radamsa doesn't even build on FreeBSD without patching its makefile and things like that) but it's just a small part)

PR #753, opened 2025-11-13
FreeBSD CI: Use dfuzzer from official packages

arrowd:


.github/workflows/build.sh
```diff
@@ -78,10 +78,9 @@ case "$1" in
             gobject-introspection pkgconf expat libdaemon dbus-glib dbus gdbm \
             libevent glib automake libtool libinotify qt5-core qt5-buildtools \
             gtk3 py311-pygobject py311-dbus py311-gdbm mono git meson socat \
-            valgrind
+            valgrind dfuzzer
```
evverx, line 81: Can you remove `meson` as well? It's a dfuzzer build dependency so it can be omitted.

(It isn't run upstream yet though but because of the FreeBSD CI I'm slowly decoupling the test suite from systemd and even ran it on Alpine (with duct tape and popsicle sticks for now) to see whether Valgrind still complains about `if_nametoindex` coming from musl. It still does. FreeBSD plugged that a long ago to judge from commits where PURIFY is mentioned: https://cgit.freebsd.org/src/log/lib/libc/net/if_nametoindex.c and `glibc` rejects stuff causing that)

evverx:
Looks like the s390x Packit jobs got stuck but it doesn't matter because it affects the FreeBSD action only and it's green.

PR #765, opened 2025-11-18
Fix 5-second use-ipv6=no hostname resolution delay

danielzgtg:
My ISP and my computers are IPv4-only. This PR eliminates the 5-second SSH lag waiting for an IPv6 response that will never come.

# Before

```bash
home@daniel-desktop3:/run/avahi-daemon$ time getent hosts daniel-tablet1.local
192.168.3.91    daniel-tablet1.local

real    0m5.023s
user    0m0.000s
sys     0m0.001s
home@daniel-desktop3:/run/avahi-daemon$ time <<<'RESOLVE-HOSTNAME-IPV6 daniel-tablet1.local' nc -U socket
-15 Timeout reached

real    0m5.003s
user    0m0.000s
sys     0m0.001s
```

# After

```bash
home@daniel-desktop3:/run/avahi-daemon$ time getent hosts daniel-tablet1.local
192.168.3.91    daniel-tablet1.local

real    0m0.122s
user    0m0.000s
sys     0m0.002s
home@daniel-desktop3:/run/avahi-daemon$ time <<<'RESOLVE-HOSTNAME-IPV6 daniel-tablet1.local' nc -U socket
-28 Invalid protocol specification

real    0m0.001s
user    0m0.000s
sys     0m0.001s
```


evverx:
Can you expand on what the use case is? If IPv6 isn't supposed to be used `mdns4*` should be in `nsswitch.conf` and as far as I understand it doesn't seem to be the case here. I don't think the patch is correct because aprotocol affects whether A/AAAA queries are sent and there is nothing wrong with sending AAAA queries over IPv4. avahi itself publishes its AAAA records over IPv4 for example.

danielzgtg:
I recently [changed](https://github.com/danielzgtg/linuxenv/commit/5c914bb619fa0677d63f56afc52f30facbffb8db) `mdns4_minimal` to `mdns_minimal` to fix a information leak security problem. `mdns4_minimal` returns UNAVAIL instead of NOTFOUND for apps that still try IPv6. That request reaches systemd-resolved, which is leaking it to DNS possibly because I added [`Domains=~`](https://wiki.archlinux.org/title/Systemd-resolved#Manually) to keep settings consistent when switching WiFi. I would not rather Google or Cloudflare see all my `daniel-computerN.local` internal-only hostnames, and it's worse for other people who use unencrypted DNS.

evverx:
Got it. I think it's more of an nss-mdns issue (and it's probably tracked there somewhere). The issue is that it calls `RESOLVE-HOSTNAME-IPV4` and `RESOLVE-HOSTNAME-IPV6` sequentially. avahi actually provides `RESOLVE-HOSTNAME` where it can send A and AAAA queries at the same time and return whatever comes first. It's possible to trigger it via D-Bus as well with something like
```sh
busctl call org.freedesktop.Avahi / org.freedesktop.Avahi.Server ResolveHostName "iisiu" -- -1 -1 some-hostname.local -1 0
```
It seems it would help in this particular case.

(The mDNS way is to use NSEC RRs (https://www.rfc-editor.org/rfc/rfc6762#section-6.1) but it isn't supported currently by some mDNS responders (including avahi) so it should probably still fall back to the "whatever comes first" method in some cases)

Either way this PR changes the API to get around another issue. I think it can be closed probably.

evverx:
I think it's tracked in https://github.com/avahi/nss-mdns/issues/83

danielzgtg:
I'll investigate nss-mdns.

> changes the API to get around another issue

Do you mean I should use `AVAHI_ERR_TIMEOUT` instead of `AVAHI_ERR_INVALID_PROTOCOL`? When using the socket, I see no reason why it should waste time waiting for something that will never come.

danielzgtg:
This is [unfixable](https://github.com/avahi/nss-mdns/issues/83#issuecomment-3549923111) in nss-mdns due to glibc. The only solution there is changing [UNAVAIL](https://github.com/avahi/nss-mdns/blob/3709e2023f0bf3ea756330dcd562c714b7e260a1/src/nss.c#L114) to NOTFOUND upon detecting `.local`, but that's an actual breaking change affecting the behavior of `/etc/nsswitch.conf` for others whose intranet DNS server responds to `.local`. Also, [resistance](https://github.com/avahi/nss-mdns/pull/98#issuecomment-3550102598) emerged there opposing "return whatever comes first". Whereas here in avahi, I just skip 5 seconds.

evverx:
Currently `-1` means that avahi should send both A and AAAA queries, 0 means that avahi should send A queries and 1 means that avahi should send AAAA queries. With your patch applied it's no longer possible to send AAAA queries over IPv4 for example and that's not correct.

> I see no reason why it should waste time waiting for something that will never come.

It doesn't come in your particular case but there is nothing wrong with sending AAAA queries over IPv4 in general. With the normal DNS for example it's perfectly fine to run
```sh
dig @1.1.1.2 google.com AAAA
```

evverx:
Either way I'll go ahead and close this PR. Let's keep discussing it in the nss-mdns repository.

pemensik:
Yes, this needs to be configured by nss-mdns plugin. That is responsible for generating two separate A and AAAA requests, which are executed serially.

Different thing which should make it much faster would be generating NSEC response, when the machine has registered this name as its own, but does not have any record to answer. It does not have to timeout. I think we have some other issue for that already.

pemensik:
> I recently [changed](https://github.com/danielzgtg/linuxenv/commit/5c914bb619fa0677d63f56afc52f30facbffb8db) `mdns4_minimal` to `mdns_minimal` to fix a information leak security problem. `mdns4_minimal` returns UNAVAIL instead of NOTFOUND for apps that still try IPv6. That request reaches systemd-resolved, which is leaking it to DNS possibly because I added [`Domains=~`](https://wiki.archlinux.org/title/Systemd-resolved#Manually) to keep settings consistent when switching WiFi. I would not rather Google or Cloudflare see all my `daniel-computerN.local` internal-only hostnames, and it's worse for other people who use unencrypted DNS.

Hmm, but this is a good point. Please fill issue on nss-mdns project to fix that. It should not be fixed here I think, but we may want to address it anyway.

PR #766, opened 2025-11-19
Fix encoding bug on architectures with unsigned char types

friendlyhugo:
This PR fixes issue #169.

Previously, `avahi_string_list_to_string` assumed `char` is signed, which is not true on all architectures (e.g. arm). This caused weird encoding issues on some platforms.

This PR fixes the issue by using `uint8_t` as the char type, since it is also used in `AvahiStringList` for char storage.

### DIFF
--- a/avahi-common/strlst.c
+++ b/avahi-common/strlst.c
@@ -157,25 +157,26 @@ AvahiStringList* avahi_string_list_reverse(AvahiStringList *l) {
  *
  *   - escapes >>> " <<< to >>> \" <<<
  *   - escapes >>> \ <<< to >>> \\ <<<
- *   - escapes bytes less than 32 to backslash-prefixed 3-digit DECIMAL form
+ *   - escapes bytes less than 32 or more than 126 to backslash-prefixed 3-digit DECIMAL form
  */
 char* avahi_string_list_to_string(AvahiStringList *l) {
     AvahiStringList *n;
     size_t s = 0;
-    char *p, *t, *e;
+    uint8_t *p;
+    char *t, *e;
 
     for (n = l; n; n = n->next) {
         if (n != l)
             s ++; /* for the inter-string separating space */
 
-        for (p = (char*) n->text; ((size_t) (p - (char*) n->text) < n->size); p++) {
+        for (p = n->text; ((size_t) (p - n->text) < n->size); p++) {
             switch (*p) {
               case '"':
               case '\\':
                   s += 2;
                   break;
               default:
-                  if (*p < 32) {
+                  if (*p < 32 || *p >= 127) {
                       s += 4;
                   } else {
                       s ++;
@@ -196,18 +197,18 @@ char* avahi_string_list_to_string(AvahiStringList *l) {
             *(e++) = ' ';
 
         *(e++) = '"';
-        for (p = (char*) n->text; ((size_t) (p - (char*) n->text) < n->size); p++) {
+        for (p = n->text; ((size_t) (p - n->text) < n->size); p++) {
             switch (*p) {
               case '"':
               case '\\':
                   *(e++) = '\\';
                   /* FALL THROUGH */
               default:
-                  if (*p < 32) {
+                  if (*p < 32 || *p >= 127) {
                       *(e++) = '\\';
-                      *(e++) = '0' + (char)  ((uint8_t) *p / 100);
-                      *(e++) = '0' + (char) (((uint8_t) *p / 10) % 10);
-                      *(e++) = '0' + (char)  ((uint8_t) *p % 10);
+                      *(e++) = '0' + (*p / 100);
+                      *(e++) = '0' + ((*p / 10) % 10);
+                      *(e++) = '0' + (*p % 10);
                   } else {
                       *(e++) = *p;
                   }


avahi-common/Makefile.am
```diff
@@ -85,7 +87,10 @@ strlst_test_SOURCES = \
 	strlst.c strlst.h \
 	malloc.c malloc.h \
 	strlst-test.c
-strlst_test_CFLAGS = $(AM_CFLAGS)
+strlst_test_signed_char_SOURCES = $(strlst_test_SOURCES)
+strlst_test_signed_char_CFLAGS = $(AM_CFLAGS) -fsigned-char
+strlst_test_unsigned_char_SOURCES = $(strlst_test_SOURCES)
+strlst_test_unsigned_char_CFLAGS = $(AM_CFLAGS) -funsigned-char
```
evverx: Can you keep `strlst_test` only? It's already run on various arches including `aarch64` (and I also merged https://github.com/avahi/avahi/commit/07a3f45cb4ec960d3ab2a5b876b2492d51d424c9 recently) so it should be enough for that test to fail when it's run on `aarch64`.

There is nothing inherently wrong with `-fsigned-char/-funsigned-char` but it's kind of weird that it's applied to one particular test instead of the whole test suite and even that wouldn't be enough to emulate various arches. It's more reliable to just run the test suite on actual machines to see how it fares there (as long as the codepaths are exercised).

friendlyhugo: Done. I have also rebased on master and it looks like your new tests have passed.

evverx: Thank you! The patch looks good to me on the whole. It got pulled by my downstream CI where the fuzz targets are run on aarch64 among other things (because aarch64 isn't well supported on OSS-Fuzz and was broken (at least aarch64 images are no longer built there: https://github.com/google/oss-fuzz/issues/14177)). Once it's green there I'll double-check it (I don't expect it to find anything but it's the process thing).

avahi-common/strlst.c
```diff
@@ -162,20 +162,20 @@ AvahiStringList* avahi_string_list_reverse(AvahiStringList *l) {
...
-        for (p = (char*) n->text; ((size_t) (p - (char*) n->text) < n->size); p++) {
+        for (p = n->text; ((size_t) (p - n->text) < n->size); p++) {
             switch (*p) {
               case '"':
               case '\\':
                   s += 2;
                   break;
               default:
-                  if (*p < 32) {
+                  if (*p < 32 || *p > 127) {
```
evverx: I think 127 should be escaped as well. It's `DEL` and shouldn't appear unescaped probably in printable strings.

I'm not sure what to do about semicolons. `avahi-browse` embeds those strings into its output separated by semicolons and things like
```sh
=;enp0s8;IPv4;he\059y;_qotd._tcp;local;freebsd.local;192.168.56.9;1;"a=\"hey;ho\""
```
can't be split easily. It can be discussed separately probably.

friendlyhugo: I agree with you on `DEL`. I updated the commits. I also added some more regression tests.

I don't have a strong opinion on `;`. I added it along with a regression test.

evverx: Thank you for moving it into separate commits. I'm still not sure what to do about `;` but I can drop the two commits when I'm sure it isn't needed (or should be escaped elsewhere).

One last thing. Is there any reason `t` and `e` were switched to `uint8_t *` in
```diff
-    if (!(t = e = avahi_new(char, s+1))) /* plus one for the trailing NUL */
+    if (!(t = e = avahi_new(uint8_t, s+1))) /* plus one for the trailing NUL */
```
? I'd keep them `char *` and return the pointer to `t` without the cast.

evverx: I think semicolons should be dealt with separately after all so the commits can be removed from this PR.

I've been looking at the patch and with the `*p >= 127` check technically it isn't even necessary to change anything else so
```diff
index f09ec38..64a7a37 100644
--- a/avahi-common/strlst.c
+++ b/avahi-common/strlst.c
@@ -175,7 +175,7 @@ char* avahi_string_list_to_string(AvahiStringList *l) {
                   s += 2;
                   break;
               default:
-                  if (*p < 32) {
+                  if (*p < 32 || *p >= 127) {
                       s += 4;
                   } else {
                       s ++;
@@ -203,7 +203,7 @@ char* avahi_string_list_to_string(AvahiStringList *l) {
                   *(e++) = '\\';
                   /* FALL THROUGH */
               default:
-                  if (*p < 32) {
+                  if (*p < 32 || *p >= 127) {
                       *(e++) = '\\';
                       *(e++) = '0' + (char)  ((uint8_t) *p / 100);
                       *(e++) = '0' + (char) (((uint8_t) *p / 10) % 10);
```
should suffice. I'd go with that for now. (Ideally that thing would have to be refactored eventually but in terms of fixing the bug it should be fine)

avahi-common/strlst-test.c
```diff
@@ -49,6 +51,12 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     a = avahi_string_list_add_pair(a, "uxknurz", NULL);
     a = avahi_string_list_add_pair_arbitrary(a, "uxknurz2", (const uint8_t*) "blafasel\0oerks", 14);
     a = avahi_string_list_add(a, "i am a \"string\" with embedded double-quotes (\\\")\nand newlines (\\n).");
+    a = avahi_string_list_add(a, "\x1f");
+    a = avahi_string_list_add(a, "\x20");
+    a = avahi_string_list_add(a, "\x7f");
+    a = avahi_string_list_add(a, "\x80");
+    a = avahi_string_list_add(a, "\xff");
```
evverx: I think it would be better to put them all in one string like `gh_issue_169=...\xff...` and then look for `gh_issue_169=...`. This way it should be more targeted and for example the other tests that can be added in the future are unlikely to interfere with this particular test.

friendlyhugo: I'll change this shortly.

avahi-common/strlst.c
friendlyhugo: That technically works, but it is implicit. I think the code is better and easier to understand when we switch to `uint8_t`. I left it as it was, but fixed your comment about `t` and `e`.

evverx:
I'd still go with the two-line patch mentioned in https://github.com/avahi/avahi/pull/766#discussion_r2554825432. I agree `uint8_t` conveys the intent better but in terms of backporting this patch the shorter version would be better. Could you expand on why you'd prefer the uint8_t version?

Can you also adjust the comment in https://github.com/avahi/avahi/blob/8b5a7ca366bfa78366698553be6751d643463fec/avahi-common/strlst.c#L158-L161
to clarify what it does now?

friendlyhugo:
> I'd still go with the two-line patch mentioned in [#766 (comment)](https://github.com/avahi/avahi/pull/766#discussion_r2554825432). I agree `uint8_t` conveys the intent better but in terms of backporting this patch the shorter version would be better. Could you expand on why you'd prefer the uint8_t version?

The two line patch only solves this because we check for `>= 127`, which is by coincidence (or not?) `SCHAR_MAX`. It's a fluke, since the logic still differs between architectures. I think that for future maintenance we should solve it explicitly, at the cost of slightly more cumbersome backports. Moreover, using `uint8_t` reduces casts in the function.

> Can you also adjust the comment in
> 
> https://github.com/avahi/avahi/blob/8b5a7ca366bfa78366698553be6751d643463fec/avahi-common/strlst.c#L158-L161
> 
> 
> to clarify what it does now?

I'll update it.

evverx:
> The two line patch only solves this because we check for >= 127, which is by coincidence (or not?) SCHAR_MAX. It's a fluke, since the logic still differs between architectures.

I'm not sure why it would differ. The printable ASCII 32..126 range is standardized and what the patch does is to escape everything outside of that range. The first version of the patch didn't escape the DEL control character so it had to switch from `char` to `uint8_t` to avoid the compiler warnings with `> 127`. With `>= 127` that switch wouldn't have been necessary.

Either way let's keep the uint8_t version.

I'll merge it once it's pulled my downstream CI and it's green there. (I don't expect it to find anything but as mentioned it's just the process thing).

avahi-common/strlst.c
```diff
@@ -157,25 +157,26 @@ AvahiStringList* avahi_string_list_reverse(AvahiStringList *l) {
  *
  *   - escapes >>> " <<< to >>> \" <<<
  *   - escapes >>> \ <<< to >>> \\ <<<
- *   - escapes bytes less than 32 to backslash-prefixed 3-digit DECIMAL form
+ *   - escapes bytes less than 32 or more then 126 to backslash-prefixed 3-digit DECIMAL form
```
evverx: Looks like there is a typo. It should be "more than" probably.

friendlyhugo: Fixed.

PR #769, opened 2025-11-22
FreeBSD CI: Use radamsa from official packages

arrowd:


.github/workflows/build.sh
```diff
@@ -81,10 +81,9 @@ case "$1" in
             gobject-introspection pkgconf expat libdaemon dbus-glib dbus gdbm \
             libevent glib automake libtool libinotify qt5-core qt5-buildtools \
             gtk3 py311-pygobject py311-dbus py311-gdbm mono git socat \
-            valgrind dfuzzer check
+            valgrind dfuzzer check radamsa
```
evverx, line 84: Last time I checked it wasn't there. Was it brought back somehow?

evverx: I've just seen https://github.com/FreeBSD/freebsd-ports/commit/c4c689ff952921d66fff1eaa9d478200e40c01f9. Nice!

Let me see if some FreeBSD-specific kludges can be removed from the action here as well.

arrowd: I ressurected the port, yes.

evverx: There are FreeBSD-specific kludges but I think it should be fine to keep them to still make it possible to build it from source should the need arise. I'll merge it.

(I'm not sure what happened to OpenScanHub. Look like it got stuck: https://openscanhub.fedoraproject.org/task/83565/)

PR #773, opened 2025-11-26
ci: add ClusterFuzzLite

evverx:
to make it possible to fuzz PRs opened against forks among other things.
CIFuzz isn't suitable for that because it can only fetch PRs opened
against the main repository. CIFuzz is better integrated with OSS-Fuzz
though in that it can download coverage reports, pick the right fuzz targets and so
on but the "MERGE_WITH_OSS_FUZZ_CORPORA" part should help CFLite to make
use of the corpora OSS-Fuzz has been accumulating at least.

https://google.github.io/clusterfuzzlite/

`fuzz-seconds` is bumped to 600 to run the fuzz targets longer.

The second commit starts analyzing actions with CodeQL.


evverx:
The s390x jobs seem to have gotten stuck again but they have nothing to do with this PR.

Issue #495, opened 2023-10-23
avahi fails to process packets where escaped labels can't fit into 256 bytes


evverx:
While I was reviewing https://github.com/lathiat/avahi/pull/494 I set the hostname to `[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[` and it totally confused another instance of avahi where the probes/announcements were rejected with:
```sh
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
server.c: Packet too short or invalid while reading question key. (Maybe a UTF-8 problem?)
```
`[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[.local` doesn't fit because `[` is replaced with `\091` 63 times (63 * 4 = 252) and `local` takes up 5 bytes.
```
    1   0.000000 fe80::3430:b6ff:fe30:94d7 → ff02::fb     MDNS 269 Standard query 0x0000 ANY 7.d.4.9.0.3.e.f.f.f.6.b.0.3.4.3.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa, "QM" question ANY [[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[.local, "QM" question AAAA fe80::3430:b6ff:fe30:94d7 PTR [[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[.local
[...]

PR #774, opened 2025-11-26
core: process QNAMEs correctly

evverx:
It fixes a bug where packets with valid QNAMEs were rejected
because they didn't fit into 256 bytes after they were escaped. Names in
resource records were handled correctly though so avahi could advertise
RRs like that and fail to answer queries at the same time for example.
The reflector didn't reflect packets with QNAMEs like that either. It
affected UTF-8 hostnames/service names mostly.

(nss-mdns still can't handle names like that because 256 is hard-coded
there but it's a separate issue that should be addressed there)

https://github.com/avahi/avahi/issues/495

The tests are added to make sure avahi works with long labels.

That part is already covered by the fuzz targets
https://introspector.oss-fuzz.com/project-profile?project=avahi so the
tests are added to the smoke/unit tests only.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -234,15 +234,32 @@ EOL
             s/^\(publish-hinfo=\).*/\1yes/;
         ' avahi-daemon/avahi-daemon.conf
 
-        cat <<'EOL' >>avahi-daemon/hosts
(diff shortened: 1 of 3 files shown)


evverx:
I'll go ahead and merge it to let OSS-Fuzz start fuzzing it.

PR #780, opened 2025-11-29
check-nss.c: Properly detect the mdns NSS module on FreeBSD

### LINKED ISSUE #775: avahi can't find nss-mdns on FreeBSD
The `avahi_nss_support` function looks for `libnss_*` but on FreeBSD the modules are named `nss_*`. Because of that avahi says that
```sh
WARNING: No NSS support for mDNS detected, consider installing nss-mdns!                                                                                      
```
when it starts and the `IsNSSSupportAvailable` D-Bus method returns `False` even when `nss-mdns` in installed and works:
```sh
$ gdbus call --system --dest org.freedesktop.Avahi --object-path / --method org.freedesktop.Avahi.Server.IsNSSSupportAvailable 
(false,)
```
Other that than it doesn't affect anything.

arrowd:
FreeBSD NSS modules do not have the "lib" prefix and their version suffix is hardcoded to .1

Fixes #775

This is inspired by the patch we already have in Ports: https://github.com/freebsd/freebsd-ports/blob/5a57344b49bb79c26ea74a42aefeeb3e51a71305/net/avahi-app/files/patch-avahi-client_check-nss.c#L1

### DIFF
--- a/avahi-client/check-nss.c
+++ b/avahi-client/check-nss.c
@@ -32,11 +32,19 @@ int avahi_nss_support(void) {
     int b = 0;
 
 #ifdef HAVE_DLOPEN
+#ifndef __FreeBSD__
     static const char * const libs[] = {
         "libnss_mdns.so.2",
         "libnss_mdns4.so.2",
         "libnss_mdns6.so.2",
         NULL };
+#else
+    static const char * const libs[] = {
+        "nss_mdns.so.1",
+        "nss_mdns4.so.1",
+        "nss_mdns6.so.1",
+        NULL };
+#endif
 
     const char * const *l;
 


avahi-client/check-nss.c
```diff
@@ -32,11 +32,19 @@ int avahi_nss_support(void) {
...
         "libnss_mdns4.so.2",
         "libnss_mdns6.so.2",
         NULL };
+#else
+    static const char * const libs[] = {
+        "nss_mdns.so.1",
+        "nss_mdns4.so.1",
+        "nss_mdns6.so.1",
+        NULL };
+#endif
```
evverx, line 47: Can you turn it into
```c
    static const char * const libs[] = {
#ifndef __FreeBSD__
        "libnss_mdns.so.2",
        ...
#else
        "nss_mdns.so.1",
        ...
#endif
        NULL };
```
? Also the minimal modules are missing. Can you add them too? (Generally it should be enough to look for `*_mdns` but in theory the modules can be shipped separately in separate packages and that's probably why it also looks for the other modules).

There are no tests currently. It should be possible to cover it by getting `check-nss-test` to return 1 when `nss-mdns` isn't detected and then add `should_fail ./avahi-client/check-nss-test` before `nss-mdns` is installed and move `run ./avahi-client/check-nss-test` to the place where `nss-mdns` is installed and can be detected. It would be great to also make sure that the `IsNSSSupportAvailable` D-Bus method returns the right result.

arrowd: >  Also the minimal modules are missing. Can you add them too?

For both Linux and FreeBSD?

evverx: They are installed by default on both Linux and FreeBSD so it should look for them on both Linux and FreeBSD I think.

evverx:
The commit is included in https://github.com/avahi/avahi/pull/781 along with the tests. I didn't add the minimal modules though (since it's technically a separate issue affecting both Linux and FreeBSD and can be addressed separately). I'll go ahead and close this.

PR #779, opened 2025-11-29
Properly destroy the pthread_mutexattr_t object after mutex initialization

arrowd:
Fixes #757

PR #783, opened 2025-11-30
libdns_sd: Avoid the race reported by ThreadSanitizer

arrowd:
The thread_running field is set to 1 twice - from the main thread after calling pthread_create() and from the spawned thread. This isn't really a race, because we're setting it to the same value, but these concurrent writes confuse ThreadSanitizer.

### DIFF
--- a/avahi-compat-libdns_sd/compat.c
+++ b/avahi-compat-libdns_sd/compat.c
@@ -310,7 +310,6 @@ static void * thread_func(void *data) {
     pthread_sigmask(SIG_BLOCK, &mask, NULL);
 
     sdref->thread = pthread_self();
-    sdref->thread_running = 1;
 
     for (;;) {
         char command;


evverx:
Let's discuss the fate of those libraries in https://github.com/avahi/avahi/pull/779 first. It should be decided whether they are properly maintained or they are abandoned.

Other than that it still fails under TSan
```sh
==================                                                                                                                                            
WARNING: ThreadSanitizer: data race (pid=436253)                                                                                                              
  Read of size 8 at 0x724000000018 by main thread:                                                                                                            
    #0 sdref_free /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:426:9 (null-test+0x4bdc33) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)          
    #1 sdref_unref /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:467:9 (null-test+0x4b97b9) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)         
    #2 DNSServiceRefDeallocate /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:526:9 (null-test+0x4b9801) (BuildId: 38bda71bab59576f4721970f13e4bed715cf97
9a)                                                                                                                                                           
    #3 DNSServiceRegister /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:1204:9 (null-test+0x4bc552) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a) 
    #4 main /home/vagrant/avahi/avahi-compat-libdns_sd/null-test.c:48:5 (null-test+0x4be766) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)              
                                                                                                                                                              
  Previous write of size 8 at 0x724000000018 by thread T1:                                                                                                    
    #0 thread_func /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:312:19 (null-test+0x4be0db) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)        
                                                                                                                                                              
  Location is heap block of size 248 at 0x724000000000 allocated by main thread:                                                                              
    #0 calloc <null> (null-test+0x4336fb) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)                                                                 
    #1 xcalloc /home/vagrant/avahi/avahi-common/malloc.c:95:15 (libavahi-common.so.3+0xcfd) (BuildId: 2fb2b0799bbba3c5309c6e930f45cb957e3590dc)               
    #2 avahi_malloc0 /home/vagrant/avahi/avahi-common/malloc.c:120:16 (libavahi-common.so.3+0xb94) (BuildId: 2fb2b0799bbba3c5309c6e930f45cb957e3590dc)        
    #3 avahi_new0_internal /home/vagrant/avahi/avahi-compat-libdns_sd/../avahi-common/malloc.h:59:12 (null-test+0x4bdf2e) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #4 sdref_new /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:368:19 (null-test+0x4ba0ee) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #5 DNSServiceRegister /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:1127:19 (null-test+0x4bbe7b) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #6 main /home/vagrant/avahi/avahi-compat-libdns_sd/null-test.c:48:5 (null-test+0x4be766) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)

  Thread T1 (tid=436268, finished) created by main thread at:
    #0 pthread_create <null> (null-test+0x43531e) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #1 sdref_new /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:405:9 (null-test+0x4ba47c) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #2 DNSServiceRegister /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:1127:19 (null-test+0x4bbe7b) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)
    #3 main /home/vagrant/avahi/avahi-compat-libdns_sd/null-test.c:48:5 (null-test+0x4be766) (BuildId: 38bda71bab59576f4721970f13e4bed715cf979a)

SUMMARY: ThreadSanitizer: data race /home/vagrant/avahi/avahi-compat-libdns_sd/compat.c:426:9 in sdref_free
==================
```

arrowd:
That's a different one, so this PR actually fixes a problem. But since you're reluctant to get these in, I'll stop there.

evverx:
> you're reluctant to get these in

@pemensik can have a different opinion so let's keep these PRs open anyway. In my opinion in its current form those things should be left alone and declared unmaintained.

pemensik:
Hmm, this is a strange change. It removes assigning running mark from thread, but does not move it to any better place. Does it make indication of running non-functional completely? In that case it should remove the variable completely.

arrowd:
It is already in a better place: https://github.com/arrowd/avahi/blob/7c7ecca8066f0ac3e9de118cb0217dcc96ec2761/avahi-compat-libdns_sd/compat.c#L408

Before this change this field was being assigned twice.

rgetz:
FWIW - I think the "fix" is just to completely remove `thread_running` (if you want to touch things).

The `thread_running` flag was introduced early in this codebase (circa 2004) as a defensive approximation of thread lifecycle state, reflecting historical caution around `pthread_join()` and thread startup ordering.

As mentioned above, modern ThreadSanitizer now shows that `thread_running` is accessed without synchronization, resulting in a data race during creation and teardown (creation is fixed in this patch, but teardown was not). The flag is also redundant: the worker thread is always created joinable, and teardown already has a reliable, explicit shutdown mechanism via `COMMAND_QUIT` and `pthread_join()`.

Modern POSIX semantics guarantee that `pthread_join()` is safe and well-defined regardless of timing, making the flag unnecessary, incorrect and the cause of confusion.

As @pemensik suggested - completely removing `thread_running` entirely and unconditionally signal and join the worker thread during `sdref_free()`, eliminates the race and simplifies the thread lifecycle logic. 

POSIX already gives us all the lifecycle primitive we need:
- `pthread_create()` : thread exists
- `pthread_join()` : thread is done

Anything else is redundant (unless you need : non-blocking teardown, multiple threads need to observe or react to the worker, threads are detached, thread creation can fail and you need a join guard, need logging / metrics / diagnostics - I don't see any of those things).

That eliminates all the TSAN issues.

rgetz:
FWIW - I posted all the TSan (traces) and fixes at #855 (there is actually an additional one that wasn't mentioned above).

That doesn't mean I disagree with:
>In my opinion in its current form those things should be left alone and declared unmaintained.

Only that if you want to look at the necessary fixes to resolve the issues - you can.

Issue #788, opened 2025-12-03
`valgrind --track-fds=yes` complains about "File descriptor 9 was never created" a lot on FreeBSD 15


evverx:
FreeBSD 15 has a native inotify implementation and it gets picked up instead of `kqueue` there. Looks like it hasn't been well integrated with Valgrind yet probably. (Switching back to `kqueue` make those errors go away)

```sh
$ freebsd-version 
15.0-RELEASE

$ valgrind --version
valgrind-3.26.0

# valgrind --track-fds=yes --error-exitcode=1 --leak-check=full avahi-daemon --debug --no-drop-root --no-proc-title
...
==78270== File descriptor 9 was never created
==78270==    at 0x4D0CECA: _poll (in /lib/libsys.so.7)
==78270==    by 0x498E2F0: ??? (in /lib/libthr.so.3)
==78270==    by 0x486B094: system_poll (simple-watch.c:627)
==78270==    by 0x486AC82: avahi_simple_poll_run (simple-watch.c:527)
==78270==    by 0x486AFA3: avahi_simple_poll_iterate (simple-watch.c:602)
==78270==    by 0x2107F8: run_server (main.c:1279)
==78270==    by 0x20E6C7: main (main.c:1708)
...
==78270== More than 1000 different errors detected.  I'm not reporting any more.
==78270== Final error counts will be inaccurate.  Go fix your program!
==78270== Rerun with --error-limit=no to disable this cutoff.  Note
==78270== that errors may occur in your program without prior warning from
==78270== Valgrind, because errors are no longer being displayed.

...
```

PR #793, opened 2025-12-04
Do not try to use both inotify and kqueue at the same time

arrowd:
It might be possible to get into HAVE_KQUEUE && HAVE_INOTIFY condition on BSD and MacOS systems that have libinotify-kqueue installed.

### DIFF
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -1238,9 +1238,7 @@ static int run_server(DaemonConfig *c) {
             goto finish;
         }
     }
-#endif
-
-#ifdef HAVE_KQUEUE
+#elif defined(HAVE_KQUEUE)
     if ((kq = kqueue()) < 0)
         avahi_log_warn( "Failed to initialize kqueue: %s", strerror(errno));
     else {


evverx:
>  BSD and MacOS systems that have libinotify-kqueue installed.

Given that `libinotify-kqueue` is a dynamic library and d8f34478830282e8f69ffeebf210c07d3d9fe755 was supposed to make sure that `inotify` isn't picked up in cases like that it doesn't seem `libinotify-kqueue` can be problematic here. In fact the FreeBSD CI installs it for some reason and looking at https://github.com/avahi/avahi/actions/runs/19917080563/job/57098231214 it seems it isn't detected:
```sh
2025-12-04T03:56:02.7646432Z     Enable Linux inotify:                      no
...
2025-12-04T03:56:00.3077822Z checking for kqueue... yes
```

FreeBSD 15 on the other hand can be problematic in that regard but I haven't looked at what's going on there closely (apart from https://github.com/avahi/avahi/issues/788). Can you expand on what this commit should address?

evverx:
On FreeBSD 15 they are both detected:
```
checking for kqueue... yes
....
    Enable Linux inotify:                      yes
```
It has nothing to do with `libinotify-kqueue` though. FreeBSD 15 has a native inotify implementation now: https://www.freebsd.org/releases/15.0R/announce/. I'll add the "postponed" label because the CI should be extended first and it needs https://github.com/cross-platform-actions/action/issues/114.

I'll add the "needs-discussion" label because I think it should be decided what should happen in cases like this. I'd go with `kqueue` for now because it's better integrated with Valgrind among other things.

I'll also add the "reviewed/needs-rework" label because in its current form the patch introduces warnings like
```sh
main.c:1071:13: warning: unused function 'kqueue_callback' [-Wunused-function]                                                                                
 1071 | static void kqueue_callback(AvahiWatch *watch, int fd, AVAHI_GCC_UNUSED AvahiWatchEvent event, AVAHI_GCC_UNUSED void *userdata) {                     
      |             ^~~~~~~~~~~~~~~                                                                                                                           
1 warning generated.         
```
so those things aren't separated cleanly.

arrowd:
> It has nothing to do with libinotify-kqueue though.

As well as FreeBSD, even. Both kqueue and inotify are API sets. They are not mutually exclusive and those two blocks of code are doing the same thing depending on the detected API set. There is no sense in repeating these code blocks twice if we somehow detect both API sets.

evverx:
> As well as FreeBSD, even. Both kqueue and inotify are API sets. They are not mutually exclusive and those two blocks of code are doing the same thing depending on the detected API set

`HAVE_KQUEUE` was added in 9225e01a984f730e6b2755a37e83c7ca4ee601cb
> Add FreeBSD kqueue support for watching for configuration file changes

so it was a FreeBSD-specific patch and it was assumed back then that they were mutually exclusive. This assumption is no longer correct but I think in practice in the context of the avahi build system it affects FreeBSD 15 only. My point was that `libinotify-kqueue` isn't relevant here at all because it isn't detected.

evverx:
All in all I think it should default to `kqueue` for now if both `inotify` and `kqueue` are detected but it should probably be possible to pass something like `--with-inotify`/`--with-kqueue` to `configure` to change the default. I'm not 100% yet though.

evverx:
I'll go ahead and close this. I'll merge https://github.com/avahi/avahi/pull/794 instead. It just picks either inotify or kqueue and defaults to kqueue if both are available. I don't think `--with-inotify/--with-kqueue` should be implemented after all because it would complicate things for no reason. When inotify on FreeBSD 15 is suitable to be used by default it should be possible to switch the default.

PR #799, opened 2025-12-11
FreeBSD Netlink support

arrowd:


evverx:
I'll go ahead and add the "needs-discussion" label because I'm not sure why it would be useful. The commit messages should say what it is, why it's needed and so on.

arrowd:
> I'll go ahead and add the "needs-discussion" label because I'm not sure why it would be useful. The commit messages should say what it is, why it's needed and so on.

There is a reason the `avahi-core/netlink.c` module was added, right? Why enabling it for FreeBSD require additional justification?

evverx:
> There is a reason the avahi-core/netlink.c module was added, right?

It was added because it makes it possible to keep track of interfaces.

> Why enabling it for FreeBSD require additional justification?

It's already possible to keep track of interfaces on FreeBSD without that patch.

nigels-com:
Looks fine to me, looks like the maintainer would like more description in the git commits or PR description.  I can sympathise with that.  :smile:

evverx:
> Looks fine to me, looks like the maintainer would like more description in the git commits or PR description

In the context of this PR it would be great to figure out why the netlink messages avahi receives on FreeBSD are different from Linux (which prevents it from picking up IPv6 link-local addresses when interfaces go down and up again and introduces a regression compared to PF_ROUTE).

PR #808, opened 2025-12-18
Simple client limits

pemensik:
Attempt to limit simple clients simple denial of service ([CVE-2025-59529](https://github.com/avahi/avahi/security/advisories/GHSA-73wf-3xmj-x82q))

### DIFF
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -1183,7 +1183,8 @@ static int run_server(DaemonConfig *c) {
         goto finish;
     }
 
-    if (simple_protocol_setup(poll_api) < 0)
+    // We accept clients only of half socket descriptors allowed
+    if (simple_protocol_setup(poll_api, config.rlimit_nofile/2) < 0)
         goto finish;
 
 #ifdef HAVE_DBUS
@@ -1462,6 +1463,20 @@ fail:
     return r;
 }
 
+static int get_one_rlimit(int resource, rlim_t *limit, const char *name) {
+    struct rlimit rl = {0,};
+    int r;
+
+    r = getrlimit(resource, &rl);
+    if (r < 0) {
+        avahi_log_warn("getrlimit(%s) failed: %s", name, strerror(errno));
+    } else {
+        *limit = rl.rlim_cur;
+    }
+
+    return r;
+}
+
 static void set_one_rlimit(int resource, rlim_t limit, const char *name) {
     struct rlimit rl;
     rl.rlim_cur = rl.rlim_max = limit;
@@ -1515,6 +1530,11 @@ static void init_rand_seed(void) {
     seed ^= (unsigned) time(NULL);
 
     srand(seed);
+}
+
+static void read_rlimits(void) {
+    config.rlimit_nofile_set = 0;
+    (void)get_one_rlimit(RLIMIT_NOFILE, &config.rlimit_nofile, "RLIMIT_NOFILE");
 }
 
 int main(int argc, char *argv[]) {
@@ -1615,6 +1635,7 @@ int main(int argc, char *argv[]) {
             goto finish;
         }
 
+        read_rlimits();
         if (load_config_file(&config) < 0)
             goto finish;
 
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -1183,8 +1183,9 @@ static int run_server(DaemonConfig *c) {
         goto finish;
     }
 
-    // We accept clients only of half socket descriptors allowed
-    if (simple_protocol_setup(poll_api, config.rlimit_nofile/2) < 0)
+    /* We accept clients only of 3/4 file descriptors allowed
+     * in this process. Keep extra for UDP sockets. */
+    if (simple_protocol_setup(poll_api, config.rlimit_nofile*3/4) < 0)
         goto finish;
 
 #ifdef HAVE_DBUS
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -90,6 +90,9 @@
 #include "dbus-protocol.h"
 #endif
 
+/* Maximal number of concurrent simple protocol clients. */
+#define MAX_NOFILE_LIMIT 4096
+
 AvahiServer *avahi_server = NULL;
 AvahiSimplePoll *simple_poll_api = NULL;
 static char *argv0 = NULL;
@@ -1151,6 +1154,7 @@ static int run_server(DaemonConfig *c) {
     const AvahiPoll *poll_api = NULL;
     AvahiWatch *sig_watch = NULL;
     int retval_is_sent = 0;
+    unsigned nofiles;
 #ifdef HAVE_INOTIFY
     AvahiWatch *inotify_watch = NULL;
 #endif
@@ -1185,7 +1189,10 @@ static int run_server(DaemonConfig *c) {
 
     /* We accept clients only of 3/4 file descriptors allowed
      * in this process. Keep extra for UDP sockets. */
-    if (simple_protocol_setup(poll_api, config.rlimit_nofile*3/4) < 0)
+    nofiles = config.rlimit_nofile * 3/4;
+    if (nofiles > MAX_NOFILE_LIMIT)
+        nofiles = MAX_NOFILE_LIMIT;
+    if (simple_protocol_setup(poll_api, nofiles) < 0)
         goto finish;
 
 #ifdef HAVE_DBUS
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -276,6 +276,15 @@ static void dns_server_browser_callback(
     }
 }
 
+static void log_request(const Client *c, const char *cmd, const char *arg) {
+    if (arg != NULL)
+        avahi_log_debug(__FILE__": Got %s request for '%s'. " CRED_FMT,
+                        cmd, arg, CRED_VALS(c->credentials));
+    else
+        avahi_log_debug(__FILE__": Got %s request. " CRED_FMT,
+                        cmd, CRED_VALS(c->credentials));
+}
+
 static void handle_line(Client *c, const char *s) {
     char cmd[64], arg[AVAHI_DOMAIN_NAME_MAX];
     int n_args;
@@ -311,19 +320,19 @@ static void handle_line(Client *c, const char *s) {
         if (!(c->host_name_resolver = avahi_s_host_name_resolver_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, arg, c->afquery = AVAHI_PROTO_INET, AVAHI_LOOKUP_USE_MULTICAST, host_name_resolver_callback, c)))
             goto fail;
 
-        avahi_log_debug(__FILE__": Got %s request for '%s'.", cmd, arg);
+        log_request(c, cmd, arg);
     } else if (strcmp(cmd, "RESOLVE-HOSTNAME-IPV6") == 0 && n_args == 2) {
         c->state = CLIENT_RESOLVE_HOSTNAME;
         if (!(c->host_name_resolver = avahi_s_host_name_resolver_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, arg, c->afquery = AVAHI_PROTO_INET6, AVAHI_LOOKUP_USE_MULTICAST, host_name_resolver_callback, c)))
             goto fail;
 
-        avahi_log_debug(__FILE__": Got %s request for '%s'.", cmd, arg);
+        log_request(c, cmd, arg);
     } else if (strcmp(cmd, "RESOLVE-HOSTNAME") == 0 && n_args == 2) {
         c->state = CLIENT_RESOLVE_HOSTNAME;
         if (!(c->host_name_resolver = avahi_s_host_name_resolver_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, arg, c->afquery = AVAHI_PROTO_UNSPEC, AVAHI_LOOKUP_USE_MULTICAST, host_name_resolver_callback, c)))
             goto fail;
 
-        avahi_log_debug(__FILE__": Got %s request for '%s'.", cmd, arg);
+        log_request(c, cmd, arg);
     } else if (strcmp(cmd, "RESOLVE-ADDRESS") == 0 && n_args == 2) {
         AvahiAddress addr;
 
@@ -336,32 +345,28 @@ static void handle_line(Client *c, const char *s) {
                 goto fail;
         }
 
-        avahi_log_debug(__FILE__": Got %s request for '%s'.", cmd, arg);
-
+        log_request(c, cmd, arg);
     } else if (strcmp(cmd, "BROWSE-DNS-SERVERS-IPV4") == 0 && n_args == 1) {
         c->state = CLIENT_BROWSE_DNS_SERVERS;
         if (!(c->dns_server_browser = avahi_s_dns_server_browser_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, NULL, AVAHI_DNS_SERVER_RESOLVE, c->afquery = AVAHI_PROTO_INET, AVAHI_LOOKUP_USE_MULTICAST, dns_server_browser_callback, c)))
             goto fail;
+
         client_output_printf(c, "+ Browsing ...\n");
-
-        avahi_log_debug(__FILE__": Got %s request.", cmd);
-
+        log_request(c, cmd, NULL);
     } else if (strcmp(cmd, "BROWSE-DNS-SERVERS-IPV6") == 0 && n_args == 1) {
         c->state = CLIENT_BROWSE_DNS_SERVERS;
         if (!(c->dns_server_browser = avahi_s_dns_server_browser_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, NULL, AVAHI_DNS_SERVER_RESOLVE, c->afquery = AVAHI_PROTO_INET6, AVAHI_LOOKUP_USE_MULTICAST, dns_server_browser_callback, c)))
             goto fail;
+
         client_output_printf(c, "+ Browsing ...\n");
-
-        avahi_log_debug(__FILE__": Got %s request.", cmd);
-
+        log_request(c, cmd, NULL);
     } else if (strcmp(cmd, "BROWSE-DNS-SERVERS") == 0 && n_args == 1) {
         c->state = CLIENT_BROWSE_DNS_SERVERS;
         if (!(c->dns_server_browser = avahi_s_dns_server_browser_new(avahi_server, AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC, NULL, AVAHI_DNS_SERVER_RESOLVE, c->afquery = AVAHI_PROTO_UNSPEC, AVAHI_LOOKUP_USE_MULTICAST, dns_server_browser_callback, c)))
             goto fail;
+
         client_output_printf(c, "+ Browsing ...\n");
-
-        avahi_log_debug(__FILE__": Got %s request.", cmd);
-
+        log_request(c, cmd, NULL);
     } else {
         client_output_printf(c, "%+i Invalid command \"%s\", try \"HELP\".\n", AVAHI_ERR_INVALID_OPERATION, cmd);
         c->state = CLIENT_DEAD;
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -35,6 +35,9 @@
 #ifdef HAVE_LIBSYSTEMD
 #include <systemd/sd-daemon.h>
 #endif
+#ifdef HAVE_STRUCT_XUCRED
+#include <sys/ucred.h>
+#endif
 
 #include <avahi-common/domain.h>
 #include <avahi-common/llist.h>
@@ -61,8 +64,87 @@
 
 #define BUFFER_SIZE (20*1024)
 
-#define CRED_FMT     "uid=%lu gid=%lu pid=%lu"
-#define CRED_VALS(ucred) (long)(ucred).uid, (long)(ucred).gid, (long)(ucred).pid
+#if __linux__
+/* Linux specific support, man 7 unix */
+typedef struct ucred AvahiCred;
+
+
+static uid_t credentials_getuid(const AvahiCred *cred) {
+	return cred->uid;
+}
+
+static gid_t credentials_getgid(const AvahiCred *cred) {
+	return cred->gid;
+}
+
+static pid_t credentials_getpid(const AvahiCred *cred) {
+	return cred->pid;
+}
+
+static int credentials_getsockopt(int fd, AvahiCred *cred, socklen_t *len) {
+    *len = sizeof(*cred);
+    memset(cred, 0, sizeof(*cred));
+    return getsockopt(fd, SOL_SOCKET, SO_PEERCRED, cred, len);
+}
+
+#  define CRED_FMT     "uid=%lu gid=%lu pid=%lu"
+#  define CRED_VALS(ucred) (long)credentials_getuid(ucred), (long)credentials_getgid(ucred), (long)credentials_getpid(ucred)
+
+#elif defined(HAVE_STRUCT_XUCRED) && defined(XUCRED_VERSION)
+/* FreeBSD support, man 4 unix */
+typedef struct xucred AvahiCred;
+
+static int credentials_version_check(const AvahiCred *cred) {
+	if (cred->cr_version != XUCRED_VERSION) {
+		avahi_log_debug("credentials_version_check: unsupported version %d",
+				cred->cr_version);
+		return 0;
+	}
+	return 1;
+}
+
+static uid_t credentials_getuid(const AvahiCred *cred) {
+	if (!credentials_version_check(cred))
+		return 0;
+	return cred->cr_uid;
+}
+
+static gid_t credentials_getgid(const AvahiCred *cred) {
+	if (!credentials_version_check(cred) || cred->cr_ngroups <= 0)
+		return 0;
+	return cred->cr_groups[0];
+}
+
+static pid_t credentials_getpid(const AvahiCred *cred) {
+	if (!credentials_version_check(cred))
+		return 0;
+	return cred->cr_pid;
+}
+
+static int credentials_getsockopt(int fd, AvahiCred *cred, socklen_t *len) {
+    *len = sizeof(*cred);
+    memset(cred, 0, sizeof(*cred));
+    return getsockopt(fd, SOL_LOCAL, LOCAL_PEERCRED, cred, len);
+}
+
+#  define CRED_FMT     "uid=%lu gid=%lu pid=%lu"
+#  define CRED_VALS(ucred) (long)credentials_getuid(ucred), (long)credentials_getgid(ucred), (long)credentials_getpid(ucred)
+
+#else
+/* No support */
+#  define CRED_FMT     "%s"
+#  define CRED_VALS(ucred) ucred
+typedef unsigned char AvahiCred; /* need known type size. */
+
+static uid_t credentials_getuid(const AvahiCred *cred) {
+	(void)cred; return 0; /* no support for getting remote user. */
+}
+
+static int credentials_getsockopt(int fd, AvahiCred *cred, socklen_t *len) {
+	(void)fd; *len = sizeof(*cred); *cred = 0;
+	return 0;
+}
+#endif
 
 typedef struct Client Client;
 typedef struct Server Server;
@@ -74,8 +156,6 @@ typedef enum {
     CLIENT_BROWSE_DNS_SERVERS,
     CLIENT_DEAD
 } ClientState;
-
-typedef struct ucred AvahiCred;
 
 struct Client {
     Server *server;
@@ -132,7 +212,7 @@ static void client_free(Client *c) {
     c->server->poll_api->watch_free(c->watch);
     close(c->fd);
     avahi_log_debug("simple client %d finished: " CRED_FMT,
-                    c->fd, CRED_VALS(c->credentials));
+                    c->fd, CRED_VALS(&c->credentials));
 
     AVAHI_LLIST_REMOVE(Client, clients, c->server->clients, c);
     avahi_free(c);
@@ -279,10 +359,10 @@ static void dns_server_browser_callback(
 static void log_request(const Client *c, const char *cmd, const char *arg) {
     if (arg != NULL)
         avahi_log_debug(__FILE__": Got %s request for '%s'. " CRED_FMT,
-                        cmd, arg, CRED_VALS(c->credentials));
+                        cmd, arg, CRED_VALS(&c->credentials));
     else
         avahi_log_debug(__FILE__": Got %s request. " CRED_FMT,
-                        cmd, CRED_VALS(c->credentials));
+                        cmd, CRED_VALS(&c->credentials));
 }
 
 static void handle_line(Client *c, const char *s) {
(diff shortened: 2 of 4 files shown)


evverx:
It completely breaks FreeBSD. SO_PEERCRED isn't supported there.

pemensik:
Ooh. freebsd does not know ucred.

https://github.com/avahi/avahi/actions/runs/20348402561/job/58466475450?pr=808#step:3:3691

And I do not know obvious replacement for it. Some hint: https://man.freebsd.org/cgi/man.cgi?query=ucred

pemensik:
This should help: https://forums.freebsd.org/threads/obtaining-user-credentials-through-a-unix-domain-socket.87923/

Potentially, getting user could be avoided if needed, on FreeBSD or other BSD as well.

evverx:
https://docs.fedoraproject.org/en-US/defensive-coding/features/Features-Authentication/#sect-Defensive_Coding-Authentication-UNIX_Domain says
> Nowadays, most systems support the SO_PEERCRED (Linux) or LOCAL_PEERCRED (FreeBSD) socket options, or the getpeereid (other BSDs, OS X). These interfaces provide direct access to the (effective) user ID on the other end of a domain socket connect, without cooperation from the other end.

> Potentially, getting user could be avoided if needed, on FreeBSD or other BSD as well.

As far as I understand it would mean that this PR doesn't fully address the issue in all the places where avahi is currently run then.

@arrowd is there any chance FreeBSD people can take a look at this?

evverx:
Also can you add tests or something?

evverx:
In the meantime I'll go ahead and remove it from the 0.9 milestone. There are a bunch of much more serious issues addressed and queued already and there is absolutely no need to delay them even more given that this PR is unlikely to be merged anytime soon.

avahi-daemon/simple-protocol.h
```diff
@@ -22,7 +22,13 @@
 
 #include <avahi-common/watch.h>
 
+/* setup variant without maximal clients is obsolete, use setup2.
+ * Provided only to not break API compatibility. */
 int simple_protocol_setup(const AvahiPoll *poll_api);
+/* max_clients is number of concurrent simple mdns clients allowed.
+ * When limit is reached, new clients are immediately closed.
+ * Per UID limit is derived from this number. */
+int simple_protocol_setup2(const AvahiPoll *poll_api, unsigned max_clients);
```
evverx: I'm not sure what https://github.com/avahi/avahi/pull/808/commits/77aa2e26d86ddd6b8a48c85506478b1686da67de is trying to accomplish. It's not the public API so it isn't necessary to keep `simple_protocol_setup` and introduce `simple_protocol_setup2`. There is no need to keep `CLIENT_MAX` for "legacy clients" either. avahi-daemon is the only client.

`CLIENT_DEBUG` shouldn't be kept either in the final version. Useful stuff should be logged unconditionally with `avahi_log_debug`. Everything else can be thrown out.

avahi-daemon/main.c
```diff
@@ -1462,6 +1464,25 @@ static int make_runtime_dir(void) {
...
+    } else {
+        *limit = rl.rlim_cur;
+    }
+
+    return r;
+}
+
+static void read_rlimits(void) {
+    config.rlimit_nofile_set = 0;
+    (void)get_one_rlimit(RLIMIT_NOFILE, &config.rlimit_nofile, "RLIMIT_NOFILE");
```
evverx: `config.rlimit_nofile` is historically configurable via `avahi-daemon.conf` (for better or for worse) so it shouldn't be changed like this. It should be taken into account instead if it's set in the config explicitly.

avahi-daemon/main.c
```diff
@@ -1462,6 +1464,25 @@ static int make_runtime_dir(void) {
     return r;
 }
 
+static int get_one_rlimit(int resource, rlim_t *limit, const char *name) {
+    struct rlimit rl = {0,};
+    int r;
+
+    r = getrlimit(resource, &rl);
+    if (r < 0) {
+        avahi_log_warn("getrlimit(%s) failed: %s", name, strerror(errno));
```
evverx, line 1480: please make sure that your PRs follow the coding style: https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md

avahi-daemon/simple-protocol.c
```diff
@@ -101,7 +108,10 @@ struct Server {
     AVAHI_LLIST_HEAD(Client, clients);
 
     unsigned n_clients;
+    unsigned max_clients;     /*< Maximal number of all simple clients. */
+    unsigned max_uid_clients; /*< Maximal number of clients of one UID. */
     int remove_socket;
+    unsigned refused_clients;
```
evverx: why is this stuff needed? It's not used anywhere

arrowd:
> As far as I understand it would mean that this PR doesn't fully address the issue in all the places where avahi is currently run then.
> 
> @arrowd is there any chance FreeBSD people can take a look at this?

Yes, this is basically the same as what I did for SSSD: https://github.com/SSSD/sssd/pull/7869/changes#diff-d005555c4e4227019f9acefa772a70f8f56440e2ae0d17d476032cc3bb49ea23

I can extended this PR or create a new one.

evverx:
> Yes, this is basically the same as what I did for SSSD: https://github.com/SSSD/sssd/pull/7869/changes#diff-d005555c4e4227019f9acefa772a70f8f56440e2ae0d17d476032cc3bb49ea23

Got it. I saw your dbus-daemon commit too: https://gitlab.freedesktop.org/dbus/dbus/-/commit/667d78ff0b1acad7254f2ba1ee8d4b74bf6dbce7.

Generally `_dbus_read_credentials_socket` (https://gitlab.freedesktop.org/dbus/dbus/-/blob/main/dbus/dbus-sysdeps-unix.c?ref_type=heads#L2212) seems to contain a lot of things relevant here (but I don't think they all can/should be covered in this PR). The question is what should happen when it's not Linux/FreeBSD.

> I can extended this PR or create a new one.

I think the approach should be ironed out first. I guess once that part is abstracted away it should be possible to cover FreeBSD.

evverx:
All in all I think it should be fine to bump `max_uid_clients` to `max_clients / 4 * 3` and on systems where avahi can't get uids all the uids can be mapped to one "unknown/unset" uid. It would be possible for one unprivileged user to reach that limit of course but it would still prevent the DoS where one unprivileged user can consume all the avahi file descriptors. Another option would be to skip the uid logic entirely there and stick to `RLIMIT_NOFILE / 4 * 3`. Some warnings should be emitted at startup probably. It isn't ideal but it would be better than completely unbuildable avahi (which is a more serious issue than the local DoS being fixed here).

By analogy with dbus-daemon the build system can also loudly fail just in case when it can't find anything useful on at least Linux and FreeBSD.

avahi-daemon/main.c
```diff
@@ -1183,7 +1183,9 @@ static int run_server(DaemonConfig *c) {
         goto finish;
     }
 
-    if (simple_protocol_setup(poll_api) < 0)
+    /* We accept clients only of 3/4 file descriptors allowed
+     * in this process. Keep extra for UDP sockets. */
```
evverx, line 1191: It isn't going to help as long as avahi leaks those sockets: https://github.com/avahi/avahi/issues/810 but it's somewhat tangential.

avahi-daemon/main.c
```diff
@@ -1183,7 +1183,9 @@ static int run_server(DaemonConfig *c) {
         goto finish;
     }
 
-    if (simple_protocol_setup(poll_api) < 0)
+    /* We accept clients only of 3/4 file descriptors allowed
```
evverx, line 1190: I think it should actually stick to some reasonable default like 4096 if RLIMIT_NOFILE is huge. IOW it should probably pick `min(4096, RLIMIT_NOFILE/4*3)`

avahi-daemon/main.c
evverx: It would make sense to take a look at what `getdtablesize` does (or maybe use it if it's available) to handle corner cases properly. In its current form it doesn't seem to be particularly good at handling corner cases.

evverx:
Looks like now Valgrind is unhappy on FreeBSD
```sh

==32947== Conditional jump or move depends on uninitialised value(s)
==32947==    at 0x212B46: is_client_allowed (simple-protocol.c:546)
==32947==    by 0x21247C: server_work (simple-protocol.c:586)
==32947==    by 0x486AE77: avahi_simple_poll_dispatch (simple-watch.c:585)
==32947==    by 0x486AFBC: avahi_simple_poll_iterate (simple-watch.c:605)
==32947==    by 0x210964: run_server (main.c:1281)
==32947==    by 0x20E85C: main (main.c:1730)
==32947==  Uninitialised value was created by a stack allocation
==32947==    at 0x212390: server_work (simple-protocol.c:570)
==32947== 
==32947== Conditional jump or move depends on uninitialised value(s)
==32947==    at 0x212E6C: credentials_getgid (simple-protocol.c:96)
==32947==    by 0x213D1B: log_request (simple-protocol.c:356)
==32947==    by 0x213981: handle_line (simple-protocol.c:440)
==32947==    by 0x213212: handle_input (simple-protocol.c:468)
==32947==    by 0x212FFA: client_work (simple-protocol.c:492)
==32947==    by 0x486AE77: avahi_simple_poll_dispatch (simple-watch.c:585)
==32947==    by 0x486AFBC: avahi_simple_poll_iterate (simple-watch.c:605)
==32947==    by 0x210964: run_server (main.c:1281)
==32947==    by 0x20E85C: main (main.c:1730)
==32947==  Uninitialised value was created by a stack allocation
==32947==    at 0x212390: server_work (simple-protocol.c:570)
==32947== 
...
```

avahi-daemon/main.c
pemensik: I made this auto-obtained before value from config file would be used. It should allow to configure it by configuration file just the same way as before. The only addition should be new smart default obtained from the system.

avahi-daemon/simple-protocol.c
pemensik: I had a plan to use it in smart logging, but have not implemented it in the end. Okay, will remove refused_clients counter.

avahi-daemon/main.c
pemensik: Uhm, that code was made to avoid using some fixed constant. Using file number limit should make it relative to whatever system it runs on. It should allow behaving well on low memory device as well as main router with plenty of memory, like Turris Omnia. I did not try any serious performance testing. But if the file limit is high for some reason, default limit should be also somehow high. It would be more useful to implement proper separate configuration knobs for per-user limit and for global limit.

evverx: > If the file limit is high for some reason, default limit should be also somehow high.

I think it makes sense as long as it's set to some reasonable value but I don't think that for example `RLIMIT_INFINITY` is reasonable in this context. I don't think it would be desirable for one UID to be able to consume those many file descriptors.

> It would be more useful to implement proper separate configuration knobs for per-user limit and for global limit

I think it would complicate things. For example the varlink stuff in systemd picks 4096 and calls it day and I don't think anyone has ever complained. Then again in the context of systems where avahi is run with systemd it's set to 1024 by default anyway and it's unclear whether it has ever been bumped anywhere to allow more clients to connect to the avahi socket.

pemensik:
Ah, reworked version compiles on FreeBSD also. I am not sure it works properly on that platform. I tried to install my own VM for very basic tests. Have it green on FreeBSD too. I hope it should work on even different platform, where not any of those PEERCRED variants is supported.

@arrowd Interesting, thank you for pointers in other projects. I have [submitted talk](https://fosdem.org/2026/schedule/event/GND79C-lwresd_revived/), which have been accepted on Fosdem.org. I hope to reuse some parts of processing of unix domain sockets platform specifics. Would it make sense to create a tiny shared library for getting remote party? I hope I will reuse this code also on bind9 coding attempts with _lwresd_. If there is more projects solving very similar problem, creating few kilobytes shared library might be helpful.

avahi-daemon/main.c
pemensik: Oh, okay. I did not consider completely unlimited number of handles. Okay, I guess 4096 is likely to be sufficient for any expected usage. We can always add better limit configuration in later versions if it proves insufficient.

avahi-daemon/main.c
evverx: > It should allow to configure it by configuration file just the same way as before

Did you push that change? As far as I can see it stills resets `rlimit_nofile_set` and overrides `config.rlimit_nofile` before calling `enforce_rlimits` and that prevents it from using `rlimit-nofile` from `avahi-daemon.conf` and setting it with
```
set_one_rlimit(RLIMIT_NOFILE, config.rlimit_nofile, "RLIMIT_NOFILE")
```

avahi-daemon/simple-protocol.c
```diff
@@ -438,6 +524,50 @@ static void client_work(AvahiWatch *watch, AVAHI_GCC_UNUSED int fd, AvahiWatchEv
...
+    }
+
+    if (n_clients > s->max_clients) {
+        avahi_log_debug("simple client %d refused: "CRED_FMT" too many clients",
+                        cfd, CRED_VALS(cred));
+        return 0;
+    }
+
+    uid = credentials_getuid(cred);
+    if (uid != 0 && n_clients > s->max_uid_clients) {
```
evverx, line 555: I'm not sure `0` should be special-cased here. For example the varlink stuff kicks it out when it reaches its limit like everyone else and there hasn't been any complaints as far as I know.

avahi-daemon/simple-protocol.c
```diff
@@ -61,7 +64,77 @@
...
+#  define CRED_FMT     "uid=%lu gid=%lu pid=%lu"
+#  define CRED_VALS(ucred) (long)credentials_getuid(ucred), (long)credentials_getgid(ucred), (long)credentials_getpid(ucred)
+
+#elif defined(HAVE_STRUCT_XUCRED)
+/* FreeBSD support, man 4 unix */
+#  define CRED_FMT     "uid=%lu gid=%lu pid=%lu"
+typedef struct xucred avahi_cred_t;
+static uid_t credentials_getuid(const avahi_cred_t *cred) {
+	if (cred->cr_version != XUCRED_VERSION)
+		return 0;
```
evverx, line 108: I don't think it should return `0` when it isn't completely clear that it's actually `root`. This stuff can be reused elsewhere at some point and it would be really bad if those functions treated unclear uids as root.

evverx: I ran that on FreeBSD 15 and to judge from
```sh
simple client 774 refused: uid=0 gid=0 pid=0 too many clients
```
it needs looking into. In that particular case `uid` was 1001, `gid` was 1001 and `pid` was 29714.

avahi-daemon/simple-protocol.c
```diff
@@ -61,7 +64,77 @@
...
+}
+static pid_t credentials_getpid(const avahi_cred_t *cred) {
+	if (cred->cr_version != XUCRED_VERSION)
+		return 0;
+	return cred->cr_pid;
+}
+static int credentials_getsockopt(int fd, avahi_cred_t *cred, socklen_t *len) {
+    *len = sizeof(*cred);
+    memset(cred, 0, sizeof(*cred));
+    return getsockopt(fd, SOL_SOCKET, LOCAL_PEERCRED, cred, len);
```
arrowd: According to FreeBSD `man unix`, the second parameter should be `SOL_LOCAL`. This might be the reason for getting zeroes ATM.

arrowd:
> Would it make sense to create a tiny shared library for getting remote party?

Unless I'm missing something, the library would end up being extremely tiny, so I'm not sure it is worth it.

avahi-daemon/simple-protocol.c
evverx: I rebuilt it with `SOL_LOCAL` and got
```sh
simple client 400 refused: uid=1001 gid=1001 pid=52276 too many uid clients: 385
```
so it seems it was because of that.

avahi-daemon/main.c
evverx: > auto-obtained before value from config file would be used

I think it should be the other way around because avahi can be run with the config with, say, `rlimit-nofile=1024` in the environment where the maximum number of file descriptors is, say, 13752. If 13752 is used before avahi enforces the limits from the config it ends up with the same DoS (where max_clients is 4096 and the actual rlimit_nofile is 1024).

Either way the part where avahi enforces the limits from the config is broken currently and it's a regression.

avahi-daemon/simple-protocol.c
pemensik: Ah, okay, fixed that.

avahi-daemon/simple-protocol.c
pemensik: Well, if it gets some different version, then we have unsupported version and cannot obtain use from the socket. I have used user 0 as a special value. getuid() does not return any special value. I would have to use pointer to uid_t and int return from the function. I used uid 0 to indicate also unsupported user obtaining call. I used it to simplify the fix a bit, because I think the result would be the same.

How differently should we handle case, when we failed to obtain remote UID? We might want to log at least debug log in this case perhaps. At the same time, zero value is used to skip iterating clients list. That is to not slow avahi on platforms we do not have working user getting anyway. I think it makes sense to avoid limiting root the same way, considering it extra privileged account.

I think it is acceptable for now, could be improved later. More combinations would make the code even less readable. But I will add at least debug log with version logging, if that ever changes.

avahi-daemon/simple-protocol.c
pemensik: Ok, I have only checked it compiles. Thanks, fixed.

avahi-daemon/simple-protocol.c
pemensik: I tried to keep the change relatively simple, because it is a CVE fix, not only feature addition. Someone might need to backport it to previous versions. Yes, using value 0 is a simplification, which might bite us sometime later. But I think works reasonably well for now to indicate some problem or missing support. Keeps the code shorter for now.

avahi-daemon/simple-protocol.c
evverx: > How differently should we handle case, when we failed to obtain remote UID?

Have you looked at the D-Bus codebase? It uses a separate type where unknown uids are mapped to -1 (it's called `DBUS_UID_UNSET` there).

> I think it is acceptable for now, could be improved later

Can you add it to the commit message and the comment for audit trail?

avahi-daemon/simple-protocol.c
evverx: As mentioned in https://github.com/avahi/avahi/pull/808#discussion_r2701897998 if you think it's OK to cut corners please mention it explicitly in commit messages/comments in the code to make it clear where the corners are cut and why it's done this way with TODOs or whatever.

pemensik:
@evverx I do not want all those changes to be squashed on merge. I would squash them already if I wanted to do that. I have reduced number of separate commits, but I think they should be merged as separate commits as they are. It is always simpler to merge them than separate them again.

evverx:
> I think they should be merged as separate commits as they are

Can you expand on why for example the commit where the counter is removed should be kept given that it shouldn't have been added in the first place?

Just to clarify there is at least one regression where avahi no longer enforces `rlimit-nofile` from the config and I still have no idea what https://github.com/avahi/avahi/pull/808/commits/825d1cca81388a886ca074a4613d8cf539b63344 does so it hasn't gotten to the point where it's close to being merged. It doesn't follow the coding style either but I think it can be ignored for now.

evverx:
For the record OpenScanHub isn't working at the moment: https://github.com/openscanhub/openscanhub/issues/333. It's a bummer because it's the first thing I look at usually.

pemensik:
I wanted to keep refused clients counter for later reuse, but I guess storing that tiny feature in extra branch is more sensible. Changed to AvahiCred, reduced few unnecessary commits. Not sure what do you mean by broken rlimit_nofile not setting. In any case it is quite late here, enough it for now. I tried it to do getting system value only before custom value would be set. It should not override config file. I am heading to bed now, will check that later. As usual we have no automated tests for that part.

evverx:
> Not sure what do you mean by broken rlimit_nofile not setting

It's all mentioned in https://github.com/avahi/avahi/pull/808#discussion_r2636704776

> As usual we have no automated tests for that part.

I'm not sure if you noticed but there was a backtrace from FreeBSD I copy-pasted from the CI and it was triggered by the nss-mdns tests talking to avahi-daemon running under Valgrind via the avahi socket so I wouldn't say there are no tests for that part. But generally tests don't appear out of thin air and this PR should come with tests (and that's why the "needs-test" label is added).

Either please rebase your PR on top of https://github.com/avahi/avahi/commit/d7fb2103cd9051239641e7320d3e2fa2f0a15761 and it should fail.

evverx:
It fails to build on NetBSD
```sh
gcc -DHAVE_CONFIG_H -I. -I..   -I/usr/pkg/include -I.. '-DDEBUG_TRAP=__asm__("int $3")' -DAVAHI_DAEMON_RUNTIME_DIR=\"/var/run/avahi-daemon/\" -DAVAHI_SOCKET=\"/var/run/avahi-daemon/socket\" -DAVAHI_SERVICE_DIR=\"/usr/pkg/etc/avahi/services\" -DAVAHI_CONFIG_FILE=\"/usr/pkg/etc/avahi/avahi-daemon.conf\" -DAVAHI_HOSTS_FILE=\"/usr/pkg/etc/avahi/hosts\" -DAVAHI_DBUS_INTROSPECTION_DIR=\"/usr/pkg/share/dbus-1/interfaces\" -DAVAHI_CONFIG_DIR=\"/usr/pkg/etc/avahi\" -D_REENTRANT -I/usr/pkg/include -DUSE_EXPAT_H   -I/usr/pkg/include -fstack-protector -std=c99 -Wall -W -Wextra -pedantic -pipe -Wformat -Wold-style-definition -Wdeclaration-after-statement -Wfloat-equal -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wshadow -Wendif-labels -Wpointer-arith -Wbad-function-cast -Wcast-qual -Wcast-align -Wwrite-strings -fdiagnostics-show-option -Wno-cast-qual -fno-strict-aliasing -MT avahi_daemon-simple-protocol.o -MD -MP -MF .deps/avahi_daemon-simple-protocol.Tpo -c -o avahi_daemon-simple-protocol.o `test -f 'simple-protocol.c' || echo './'`simple-protocol.c
simple-protocol.c:175:15: error: variable or field 'credentials' declared void
     AvahiCred credentials;
               ^~~~~~~~~~~
simple-protocol.c: In function 'client_free':
simple-protocol.c:223:54: error: expected expression before ')' token
                     c->fd, CRED_VALS(&c->credentials));
                                                      ^
simple-protocol.c: In function 'client_new':
simple-protocol.c:238:22: warning: dereferencing 'void *' pointer
     c->credentials = *cred;
                      ^~~~~
simple-protocol.c:238:20: error: void value not ignored as it ought to be
     c->credentials = *cred;
                    ^
simple-protocol.c: In function 'log_request':
simple-protocol.c:370:61: error: expected expression before ')' token
                         cmd, arg, CRED_VALS(&c->credentials));
                                                             ^
simple-protocol.c:367:39: warning: unused parameter 'c' [-Wunused-parameter]
 static void log_request(const Client *c, const char *cmd, const char *arg) {
                                       ^
simple-protocol.c: In function 'is_client_allowed':
simple-protocol.c:544:27: warning: invalid application of 'sizeof' to a void type [-Wpointer-arith]
     socklen_t len = sizeof(*cred);
                           ^
simple-protocol.c:558:45: error: expected expression before ')' token
                         cfd, CRED_VALS(cred));
                                             ^
simple-protocol.c:574:57: error: expected expression before ',' token
                                     cfd, CRED_VALS(cred), present);
                                                         ^
simple-protocol.c:581:52: error: expected expression before ')' token
                     cfd, n_clients, CRED_VALS(cred));
                                                    ^
simple-protocol.c: In function 'server_work':
simple-protocol.c:599:23: error: variable or field 'cred' declared void
             AvahiCred cred;
                       ^~~~
At top level:
simple-protocol.c:147:14: warning: 'credentials_getpid' defined but not used [-Wunused-function]
 static pid_t credentials_getpid(const AvahiCred *cred) {
              ^~~~~~~~~~~~~~~~~~
simple-protocol.c:143:14: warning: 'credentials_getgid' defined but not used [-Wunused-function]
 static gid_t credentials_getgid(const AvahiCred *cred) {
              ^~~~~~~~~~~~~~~~~~
gmake[2]: *** [Makefile:898: avahi_daemon-simple-protocol.o] Error 1
gmake[2]: Leaving directory '/home/vagrant/avahi/avahi-daemon'
gmake[1]: *** [Makefile:835: all-recursive] Error 1
gmake[1]: Leaving directory '/home/vagrant/avahi'
gmake: *** [Makefile:745: all] Error 2
```

pemensik:
Ok, correct. I have put reading rlimit after config file parsing, but before setting the limit. It makes it reset the value read from loaded config file. Fixed that and hopefully also unsupported peercred platform, like NetBSD.

evverx:
It still fails on NetBSD
```sh
gcc -DHAVE_CONFIG_H -I. -I..   -I/usr/pkg/include -I.. '-DDEBUG_TRAP=__asm__("int $3")' -DAVAHI_DAEMON_RUNTIME_DIR=\"/var/run/avahi-daemon/\" -DAVAHI_SOCKET=\"/var/run/avahi-daemon/socket\" -DAVAHI_SERVICE_DIR=\"/usr/pkg/etc/avahi/services\" -DAVAHI_CONFIG_FILE=\"/usr/pkg/etc/avahi/avahi-daemon.conf\" -DAVAHI_HOSTS_FILE=\"/usr/pkg/etc/avahi/hosts\" -DAVAHI_DBUS_INTROSPECTION_DIR=\"/usr/pkg/share/dbus-1/interfaces\" -DAVAHI_CONFIG_DIR=\"/usr/pkg/etc/avahi\" -D_REENTRANT -I/usr/pkg/include -DUSE_EXPAT_H   -I/usr/pkg/include -fstack-protector -std=c99 -Wall -W -Wextra -pedantic -pipe -Wformat -Wold-style-definition -Wdeclaration-after-statement -Wfloat-equal -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wshadow -Wendif-labels -Wpointer-arith -Wbad-function-cast -Wcast-qual -Wcast-align -Wwrite-strings -fdiagnostics-show-option -Wno-cast-qual -fno-strict-aliasing -MT avahi_daemon-simple-protocol.o -MD -MP -MF .deps/avahi_daemon-simple-protocol.Tpo -c -o avahi_daemon-simple-protocol.o `test -f 'simple-protocol.c' || echo './'`simple-protocol.c
simple-protocol.c: In function 'credentials_getsockopt':
simple-protocol.c:152:16: warning: assignment to '__socklen_t *' {aka 'unsigned int *'} from 'long unsigned int' makes pointer from integer without a cast [-Wint-conversion]
  152 |  (void)fd; len = sizeof(*cred); *cred = 0;
      |                ^
simple-protocol.c:151:71: warning: parameter 'len' set but not used [-Wunused-but-set-parameter]
  151 | static int credentials_getsockopt(int fd, AvahiCred *cred, socklen_t *len) {
      |                                                                       ^
simple-protocol.c: In function 'client_free':
simple-protocol.c:223:54: error: expected expression before ')' token
  223 |                     c->fd, CRED_VALS(&c->credentials));
      |                                                      ^
simple-protocol.c: In function 'log_request':
simple-protocol.c:370:61: error: expected expression before ')' token
  370 |                         cmd, arg, CRED_VALS(&c->credentials));
      |                                                             ^
simple-protocol.c:367:39: warning: unused parameter 'c' [-Wunused-parameter]
  367 | static void log_request(const Client *c, const char *cmd, const char *arg) {
      |                         ~~~~~~~~~~~~~~^
simple-protocol.c: In function 'is_client_allowed':
simple-protocol.c:558:45: error: expected expression before ')' token
  558 |                         cfd, CRED_VALS(cred));
      |                                             ^
simple-protocol.c:574:57: error: expected expression before ',' token
  574 |                                     cfd, CRED_VALS(cred), present);
      |                                                         ^
simple-protocol.c:581:52: error: expected expression before ')' token
  581 |                     cfd, n_clients, CRED_VALS(cred));
      |                                                    ^
At top level:
simple-protocol.c:147:14: warning: 'credentials_getpid' defined but not used [-Wunused-function]
  147 | static pid_t credentials_getpid(const AvahiCred *cred) {
      |              ^~~~~~~~~~~~~~~~~~
simple-protocol.c:143:14: warning: 'credentials_getgid' defined but not used [-Wunused-function]
  143 | static gid_t credentials_getgid(const AvahiCred *cred) {
      |              ^~~~~~~~~~~~~~~~~~
gmake[2]: *** [Makefile:898: avahi_daemon-simple-protocol.o] Error 1
gmake[2]: Leaving directory '/home/vagrant/avahi/avahi-daemon'
gmake[1]: *** [Makefile:835: all-recursive] Error 1
gmake[1]: Leaving directory '/home/vagrant/avahi'
gmake: *** [Makefile:745: all] Error 2
```
by the way the cross-platform action used upstream to test avahi on FreeBSD supports NetBSD as well (https://github.com/cross-platform-actions/action?tab=readme-ov-file#netbsd-netbsd) so you can turn it on in at least build mode.

I'll take a look at the PR next week. In the meantime it would be great to figure out how to let downstreams like NetBSD know that this stuff doesn't fully work there (one option would be to emit annoying runtime warnings by analogy with the nss-mdns support warnings. They are more likely be noticed than build time warnings nobody looks at). Either way it should certainly be mentioned in the commit messages that only Linux and FreeBSD are fully covered at this point.

pemensik:
Oh, okay. Realized I can test it by modification of ifdef __linux__ line. Made few repetitions to not emit even warnings. This should be finally working now on all platforms.

evverx:
> Made few repetitions to not emit even warnings

cppcheck is unhappy anyway (https://openscanhub.fedoraproject.org/task/93906/log/added.html) but assuming the `(void)*` thing is related to those warnings please use `AVAHI_GCC_UNUSED` for that. In the meantime until I have gotten round to it there should be enough time to start following https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md (especially the curly braces part).

avahi-daemon/simple-protocol.c
```diff
@@ -438,6 +532,48 @@ static void client_work(AvahiWatch *watch, AVAHI_GCC_UNUSED int fd, AvahiWatchEv
...
 
+static int is_client_allowed(Server *s, int cfd, AvahiCred *cred) {
+    socklen_t len = sizeof(*cred);
+    unsigned n_clients;
+    uid_t uid;
+
+    assert(s != NULL);
+
+    n_clients = s->n_clients + 1;
+    if (credentials_getsockopt(cfd, cred, &len) != 0) {
```
evverx, line 543: As far as I can see `len` is passed around and then just ignored. It should make sure that `len == sizeof(*cred)`. With that the redundant `memset`s can be removed from `credentials_getsockopt`. The redundant `len = sizeof(*cred)` can be removed from `credentials_getsockopt` as well.

evverx: Having taken a closer look those `memset`s popped up to silence Valgrind when it complained about "Conditional jump or move depends on uninitialised value". It complained because `SOL_SOCKET` was passed instead of `SOL_LOCAL` on FreeBSD and `getsockopt` actually read 4 bytes instead of 88 (which would have been caught if `len` was checked). Please add the check and remove the kludges.

avahi-daemon/main.c
```diff
@@ -1615,6 +1643,7 @@ int main(int argc, char *argv[]) {
             goto finish;
         }
 
+        read_rlimits();
```
evverx, line 1646: I have to say it's hard to follow this. Is there a reason it's spread all over the place along with unnecessary changes to `config.*` instead of getting the limit in `simple_protocol_setup` and using it there? At that point the config is taken into accout, the soft/hard limits are set and the code setting up the simple protocol can rely on that.

evverx:
Also I rolled out NetBSD in https://github.com/avahi/avahi/pull/839 so please rebase this PR on top of the master branch to make sure it actually compiles and work there.

nikhilgv:
Hello @evverx , when will this be merged?

evverx:
> when will this be merged?

All the points should be addressed and ideally tests should be added.

nikhilgv:
Hello @evverx , any update on when this will be merged?

evverx:
There hasn't been any activity so it hasn't gotten any closer to being merged yet.

sudhireic:
Hi @evverx  , just like to follow up to see if there are any updates on the merge timeline. Would appreciate if share any update.

evverx:
There haven't been any updates.

@pemensik I'll go ahead and close this. Please reopen it when it's rebased and all the points are addressed (especially the kludges silencing Valgrind).

pemensik:
What do you mean close it?? Is there any other fix for already published CVE?

evverx:
> What do you mean close it??

It isn't suitable to be merged in its current form so there is no point in keeping it open.

> Is there any other fix for already published CVE?

No there isn't. Since this PR comes with no tests it's unclear whether it fixes the issue either (and it doesn't fully fix it everywhere of course).

Either way there are questions regarding timelines so please answer them then.

pemensik:
> It isn't suitable to be merged in its current form so there is no point in keeping it open.

I am sorry, this is non-sense. It means there remains some work to do, unless this was solved by alternative approach. That is not the case here, or is it?
> 
> No there isn't. Since this PR comes with no tests it's unclear whether it fixes the issue either (and it doesn't fully fix it everywhere of course).

Tests are good, but if we should fix CVE by manually tested change like this or keep it unfixed, I say fix it somehow. Yes, test coverage is needed and good infrastructure.

> Either way there are questions regarding timelines so please answer them then.

Timelines are independent. Please don't close my PR without offering alternatives. It is okay to add additional commits with more fixes. But closing it is not the way.

evverx:
> It means there remains some work to do

Convert it to draft then.

>  if we should fix CVE by manually tested change like this or keep it unfixed, I say fix it somehow

Go and test it then before opening PRs. I haven't seen a lot of testing and the latest push is red.

evverx:
Either way I'll unsubscribe from this PR. Let me know when you think it's ready and it at least follows https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md.

pemensik:
> Either way I'll unsubscribe from this PR. Let me know when you think it's ready and it at least follows https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md.

Why is that file hidden in .github directory? It seems it should reside at docs/ instead, where people would search for it.

evverx:
Did you read the reply to the same question you asked in January in https://github.com/avahi/avahi/commit/88ab674765e3ac0ac6ba7cefef0e1a6f9d626cec#commitcomment-174357177?

pemensik:
Damn, I have overwritten the upstream branch by incomplete version just to retry CI. I will have to push again from the computer at home, which has correct latest version.

evverx:
Closing this in favor of https://github.com/avahi/avahi/pull/975. That one comes with the test at least.

PR #813, opened 2025-12-25
Fix incorrect memmove in avahi-daemon (simple protocol server)

### LINKED ISSUE #715: incorrect memmove in simple protocol server
client_work() [here](https://github.com/avahi/avahi/blob/2bb03084505fb9e2041ee8083a0ca08f1665f0f5/avahi-daemon/simple-protocol.c#L389) reduces c->outbuf_length by the number of bytes written r, then (if anything remains) calls memmove(c->outbuf, c->outbuf + r, c->outbuf_length - r). Because c->outbuf_length was already decremented, the third argument becomes negative when r exceeds the remaining length - promoting to a huge size_t and causing an over-read/overwrite. The fixed output buffer is 20 KiB, meaning this bug is probably not actually triggerable, as it requires a partial write which the output buffer of 20kib doesn't allow for.

damiank89:
This PR fixes #715.

### DIFF
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -419,8 +419,7 @@ static void client_work(AvahiWatch *watch, AVAHI_GCC_UNUSED int fd, AvahiWatchEv
         assert((size_t) r <= c->outbuf_length);
         c->outbuf_length -= r;
 
-        if (c->outbuf_length)
-            memmove(c->outbuf, c->outbuf + r, c->outbuf_length - r);
+        memmove(c->outbuf, c->outbuf + r, c->outbuf_length);
 
         if (c->outbuf_length == 0 && c->state == CLIENT_DEAD) {
             client_free(c);


evverx:
Thank you for the PR.

Can you add "Reported-by: Joshua Rogers <joshua.hu> using the ZeroPath tool <zeropath.com>" to the commit message? Also the commit message could expand on how unlikely it is for this bug to be triggered. The reason it's reported publicly instead of going the CVE route is that it can't be easily triggered in practice (generally the simple protocol is regularly fuzzed with radamsa on the CI when avahi-daemon itself is run under ASan/UBSan/Valgrind and it has never been observed in practice).

damiank89:
Added.

evverx:
Thank you. This branch is out-of-date with the master branch and "allow edits from maintainers" is off so I can't rebase it myself here. Can you also rebase the commit on top of the master branch?

damiank89:
Rebased.

PR #819, opened 2025-12-29
Define code style using clang-format with guidelines from PulseAudio

damiank89:
I took the time to implement the guidelines from [PulseAudio](https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/CodingStyle/) in `clang-format`.

It's not perfect, but I think it's close enough. I added a workflow in `.github/workflows/style.yml` to enforce this style.

The following style guidelines proved hard to implement using clang-format, so were not implemented:
- Avoid unnecessary curly braces on one-line code blocks.
- No C++ comments please!
- If `#include <>` is used the full path to the file relative to `src/` should be used.

There are some other guidelines, like "Every .c file should have a matching .h file", but these are not things that clang-format is designed to do.

The last commit on this branch just runs `clang-format` on the entire source tree. It might be better to remove it and run `clang-format` directly on `master` after this is merged.


evverx:
I'm afraid the codebase can't be just reformatted like this. Various patches are applied on top of avahi downstream and changes like this invalidate those patches and lead to unnecessary code churn everywhere.

Generally the goal is to enforce this for new changes and apply various tools differentially (for example OpenScanHub reports newly introduced issues, the differential shellcheck looks at diffs and so on). Ideally all those things have to be fixed and there would be no need to look at diffs only but I don't think it can be achieved anytime soon.

damiank89:
Good point. Do you have any idea of how this could be implemented? We could run clang-format only on modified files, but that means that all of the files would need to be reformatted eventually.

Do you have a grasp of the amount of patches that would need to be "rebased" because of this change? If it's not that many maybe it's a good idea to just adjust them.

evverx:
> Do you have any idea of how this could be implemented?

I think it can be implemented by analogy with what OpenScanHub and the differential shellcheck do. Generally they look at whether findings are included in diffs only. Then again in PRs I've seen it's either tabs or curly braces and curly braces aren't covered so I'm not sure whether it would be worth it in the end.

> Do you have a grasp of the amount of patches that would need to be "rebased" because of this change?

Given that avahi is really old I think there are quite a few places where patches are applied on top. For example there is a PR covering a very specific use case where a drone comes with network interfaces with aliases or something like that. It isn't going to be merged upstream but it's useful in some environments anyway. There are of course patches applied downstream without PRs upstream.

> If it's not that many maybe it's a good idea to just adjust them.

I don't think that people who would have to do that would agree with that. I don't agree with that either.

damiank89:
> I don't think that people who would have to do that would agree with that. I don't agree with that either.

Got it.

Do you think it's worth a shot to try and implement this check as a differential check in the CI?

evverx:
No I don't.

PR #820, opened 2025-12-31
Mitigate all compilation warnings to allow avahi to compile with `-Werror`

### LINKED ISSUE #595: avahi fails to compile with `-Werror`
Both `gcc` and `clang` produce a lot of warnings when avahi is compiled. They should be fixed to for example be able to pass `-Werror` on the CI to catch stuff automatically.

Prompted by https://github.com/avahi/avahi/pull/592#issuecomment-2080473304 and some other PRs where new warnings weren't caught automatically.

damiank89:
This PR fixes #595.

It doesn't add `-Werror` as a CFLAG, because I couldn't figure out how it should be added without interfering with regular AC checks.

I ran `autogen.sh` on my Ubuntu machine as so:
```bash
./autogen.sh --disable-mono --disable-gdbm --disable-python-dbus
```

Then ran `make` manually and went through all compilation warnings and fixed them.

There were some caveats:
1. `glib-genmarshal` generates bad ISO C code, which does not compile without `-Wno-pedantic`.
2. In `avahi-gobject`, I think there was a bug which could have printed some warning due to a missing `break`.
3. In `avahi-ui`, there is much usage of deprecated APIs. I passed some specific defines to disable the deprecation warnings.

There aren't many logic changes in this PR, as I made sure to keep the logic as is where possible. With that being said, it's probably a good idea to error-out in some of these cases, instead of just logging a warning / error.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -471,6 +471,7 @@ AC_ARG_ENABLE(gobject,
         [HAVE_GOBJECT=yes])
 
 if test "x$HAVE_GOBJECT" = "xyes" ; then
+        CFLAGS="$CFLAGS -Wno-pedantic"
         PKG_CHECK_MODULES(GOBJECT, [ glib-2.0 >= 2.4.0 gobject-2.0 ])
         AC_SUBST(GOBJECT_CFLAGS)
         AC_SUBST(GOBJECT_LIBS)


evverx:
@damiank89 I'm planning to update the contributing guidelines to mention that if LLMs are involved in producing PRs, bug reports, security reports and so on it should be mentioned explicitly. Until then can you expand in the commit messages on whether LLMs are involved in generating parts of this PR and how exactly they are involved?

(Just to be clear this PR isn't going to be merged because it touches a bunch of unrelated parts of avahi, doesn't follow the coding style (which is weird given that in the other PR there were plans to enforce the coding style so I'd assume it was read)) and so on)

damiank89:
> @damiank89 I'm planning to update the contributing guidelines to mention that if LLMs are involved in producing PRs, bug reports, security reports and so on it should be mentioned explicitly. Until then can you expand in the commit messages on whether LLMs are involved in generating parts of this PR and how exactly they are involved?

No LLMs were used to create this PR, or any other PR / issue I've worked on, and I don't appreciate the assumption.

> (Just to be clear this PR isn't going to be merged because it touches a bunch of unrelated parts of avahi, doesn't follow the coding style (which is weird given that in the other PR there were plans to enforce the coding style so I'd assume it was read)) and so on)

Some `if` statements don't follow the guidelines because I initially added goto / return / exit in them and didn't remove the curly-braces after removing them (lines were removed because I didn't want to change the logic without a discussion). The goal of this PR is to discuss each fixed warning and figure out what's the optimal solution (as written in the PR message), not to be merged as is. I don't have a problem with opening a seperate PR for each of them, as long as we discuss it first.

evverx:
> The goal of this PR is to discuss each fixed warning and figure out what's the optimal solution (as written in the PR message)

Why do you think this PR is useful given that it's necessary (for me or any other reviewer) to go over every change anyway and figure out whether it makes sense or not? The commit messages aren't exactly useful in terms of explaining why they do what they do.

damiank89:
> Why do you think this PR is useful given that it's necessary (for me or any other reviewer) to go over every change anyway and figure out whether it makes sense or not? The commit messages aren't exactly useful in terms of explaining why they do what they do.

To begin with, it maps all of the warnings, but I get your point. I'll update the commits with possible fixes and detailed messages (instead of just logging warnings). I'll let you know when I'm done. It'd be nice to get a proper code-review afterwards.

evverx:
>  it maps all of the warnings

It doesn't map all the warnings. For example the places clang complains about aren't touched (because I assume it was done with gcc only). But other than there are a lot of places in the CI where all those warnings can be viewed.

Either way I think it would make sense to pick the important warnings in the important components first and fix them one by one with the rationale behind every change.

damiank89:
I've decided to open a separate PR for each change. We can close this PR and discuss each change separately.

PR #825, opened 2026-01-01
Add missing break in ga_client_get_property to prevent incorrect warnings in avahi-gobject

damiank89:
GCC complained about an implicit fallthrough, which looks like a missing break statement. This is a bug that just caused incorrect warning messages, without affecting behavior.

This is part of #820.

### DIFF
--- a/avahi-gobject/ga-client.c
+++ b/avahi-gobject/ga-client.c
@@ -103,6 +103,7 @@ static void ga_client_get_property(GObject * object,
             break;
         case PROP_FLAGS:
             g_value_set_enum(value, priv->flags);
+            break;
         default:
             G_OBJECT_WARN_INVALID_PROPERTY_ID(object, property_id, pspec);
             break;


damiank89:
@evverx, Do you think my commit message is detailed enough or does it require any changes?

evverx:
The high-level description is fine (though can you add the warning ga-client emits at runtime to make it easier to find it should anyone actually run into it?). As mentioned in another PR can you also include exact warnings your PRs fix in addition to high-level descriptions. Can you also add "Replaces https://github.com/avahi/avahi/pull/690"?

(Ideally there would be a test calling `ga_client_new` and getting the flags to make sure it actually works but it can be done later. I'll add the "needs-tests" label to indicate that that stuff isn't tested anyway)

damiank89:
Sure, I'll add the warning and the "Replaces" line to the commit message.

evverx:
Please add the following part before "Replaces ..." and it should be good to go.

It fixes the following warning:
```
ga-client.c: In function 'ga_client_get_property':
ga-client.c:105:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
  105 |             g_value_set_enum(value, priv->flags);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ga-client.c:106:9: note: here
  106 |         default:
      |         ^~~~~~~
```
In the meantime I'll close https://github.com/avahi/avahi/pull/690.

damiank89:
Added.

evverx:
The s390x Packit jobs appear to have gotten stuck again but it doesn't matter in this particular case. OpenScanHub confirmed that the warning is gone: https://openscanhub.fedoraproject.org/task/90070/log/fixed.html. Merging.

PR #826, opened 2026-01-05
Fix GCC compilation warnings in `init_rand_seed()`

damiank89:
GCC complains about unused return values from `read()` calls in `init_rand_seed()` in both avahi-autoipd and avahi-daemon. These failures are accounted for by XORing `time()` to the seed, so it does not really matter if it's initialized by `read()`.

This is part of #820.

### DIFF
--- a/avahi-daemon/main.c
+++ b/avahi-daemon/main.c
@@ -1507,7 +1507,8 @@ static void init_rand_seed(void) {
      * less predictable, and to make sure that multiple machines
      * booted at the same time choose different random seeds.  */
     if ((fd = open(RANDOM_DEVICE, O_RDONLY)) >= 0) {
-        read(fd, &seed, sizeof(seed));
+        if (read(fd, &seed, sizeof(seed)) <= 0)
+            avahi_log_warn("read() from \"%s\" failed: %s", RANDOM_DEVICE, strerror(errno));
         close(fd);
     }
 


evverx:
`avahi-autoipd` isn't maintained anymore (it was discussed in https://github.com/avahi/avahi/issues/467 for example) so its warnings aren't relevant. Other than that the part where avahi deals with random stuff needs revisiting (instead of silencing the warnings) and it's tracked in https://github.com/avahi/avahi/issues/347.

damiank89:
> the part where avahi deals with random stuff needs revisiting (instead of silencing the warnings) and it's tracked in #347.

#347 is open since 2021 and nothing has changed. If you'd like, I can add an assertion in `init_rand_seed()`, or change it to return an int and check that in `main()`.

evverx:
> nothing has changed

That part got more disorganized with f8710bdc8b29ee1176fe3bfaeabebbda1b7a79f7 merged actually.

> If you'd like, I can add an assertion in init_rand_seed(), or change it to return an int and check that in main().

I don't think one-off things like that are useful. If you are interested in improving that part it would probably make sense to take a look at places where those things are relevant and figure out how to improve it. Before opening PRs it would make sense to discuss that in that issue first.

damiank89:
These are not one-off things, these are PRs aimed to fix smaller issues, in order to make avahi compile with `-Werror`, and from what I can tell, following the leads the GCC warns about actually results in some weird quirks that need to be ironed out.

More specifically, I didn't know about #347 until you mentioned it now, so it didn't cross my mind to discuss it there. Now that you've mentioned that `getrandom()` mess, I agree that it makes things a bit more clunky. With that being said, this could be a good opportunity to actually fix both issues. What do you think would be the optimal solution to the random source problem? (We can discuss this in #347 if you prefer)

evverx:
> make avahi compile with -Werror

The goal in general is to make avahi work correctly. Silencing warnings without looking at them closely doesn't necessarily help to reach that goal.

> following the leads the GCC warns about actually results in some weird quirks that need to be ironed out

That's why I opened https://github.com/avahi/avahi/issues/595.

> With that being said, this could be a good opportunity to actually fix both issues

I don't think it's productive to try to address a lot of issues at once. I think it would make sense to for example focus on https://github.com/avahi/avahi/pull/827 first and start adding tests, less vague commit messages and so on. Those things take a lot of time (and that's why they aren't addressed).

damiank89:
Got it. I'll focus on #827 for now, but I'd appreciate if you'd take a minute to look at the other PRs, they are really simple.

evverx:
I kind of did. For example https://github.com/avahi/avahi/pull/825 is a duplicate of https://github.com/avahi/avahi/pull/690. The reason https://github.com/avahi/avahi/pull/690 hasn't been merged is because the commit message there is vague and needs adjusting but since it was part of a bunch of LLM-generated PRs the author sent to a lot of different projects (unlocking the "vibe-coded" tag in an LLM-related project of all places along the way) there was no point in any sort of review.

damiank89:
Oh, didn't notice that PR. Let's discuss this in #825.

PR #827, opened 2026-01-05
Fix potential buffer exhaustion and undefined behavior in avahi-core

damiank89:
This issue was pointed out by a GCC warning. The record names generated in `avahi_server_add_dns_server_address()` and `server_add_dns_server_name()` might exhaust the output buffer supplied to `snprintf()`. Since the return value of `snprintf()` isn't checked, this can potentially lead to undefined behavior, as the wrong record names are used in `avahi_record_new_full()`.

This can be triggered by setting the following configuration values for avahi-daemon:
```
[server]
domain-name=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb.cccccccccccccccccccccccccccccccccccccccccccccccc.dddddddddddddddddddddddddddddddddddddddddddddddd.eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee.ffffffffffffffffffffffffffffffffffffffffffffffff.gggggggggggggggggggggggggggggggggggggggggggggggg.hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh.iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii.jjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj.kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk.llllllllllllllllllllllllllllllllllllllllllllllll.mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm.nnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnn.oooooooooooooooooooooooooooooooooooooooooooooooo.pppppppppppppppppppppppppppppppppppppppppppppppp.qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq.rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr.ssssssssssssssssssssssssssssssssssssssssssssssss.tttttttttttttttttttttttttttttttttttttttttttttttt.uuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu

[publish]
publish-resolv-conf-dns-servers=yes
```

This is part of #820.

### DIFF
--- a/avahi-core/avahi-test.c
+++ b/avahi-core/avahi-test.c
@@ -24,11 +24,14 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
+#include <string.h>
 
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+
+#include <avahi-core/internal.h>
 
 #include <avahi-common/malloc.h>
 #include <avahi-common/simple-watch.h>
@@ -324,7 +327,7 @@ static void dsb_callback(
     avahi_log_debug("DSB: (%i.%i): %s/%s:%i [%s]", iface, protocol, hostname ? hostname : "NULL", t, port, browser_event_to_string(event));
 }
 
-int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
+static int test_resolvers(void) {
     AvahiSRecordBrowser *r;
     AvahiSHostNameResolver *hnr;
     AvahiSAddressResolver *ar;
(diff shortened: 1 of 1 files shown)


evverx:
> This issue was pointed out by a GCC warning

Can you add the warnings to the commit messages so that it wouldn't be necessary to guess what exactly it's trying to fix? In this particular case they can be copy-pasted from https://openscanhub.fedoraproject.org/task/89650/log/fixed.html.
```
warning[-Wformat-truncation=]: '%s' directive output may be truncated writing up to 1013 bytes into a region of size between 997 and 1001
```
makes it clear what is being fixed. The specific conditions where it can be triggered in practice along with the actual behavior observed like advertising stuff like
```
ip-7f000035.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.bb.
```
should go to the commit messages as well.

Either way that part isn't covered by tests so it should be covered to make sure it actually works sensibly.

nigels-com:
Looks good to me.  The `snprintf` man page is reasonably clear about it:

> The functions snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was truncated due to this limit, then the return value is the num‐
       ber of characters (excluding the terminating null byte) which would have been written to the final string if enough space had been available.  Thus, a return value of size or more means that the  output
       was truncated.
  
  It does not seem appropriate to continue with truncated output, TBH.

nigels-com:
With `-Wall` on Ubuntu 24.04 LTS

```
In file included from /usr/include/stdio.h:980,
                 from entry.c:27:
In function 'snprintf',
    inlined from 'server_add_dns_server_name' at entry.c:931:5:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:10: note: '__builtin___snprintf_chk' output between 14 and 1031 bytes into a destination of size 1014
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~
entry.c: In function 'avahi_server_add_dns_server_address':
entry.c:987:37: warning: '%s' directive output may be truncated writing up to 63 bytes into a region of size 60 [-Wformat-truncation=]
  987 |         snprintf(n, sizeof(n), "ip6-%s.%s", h, domain);
      |                                     ^~      ~
In function 'snprintf',
    inlined from 'avahi_server_add_dns_server_address' at entry.c:987:9:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:10: note: '__builtin___snprintf_chk' output 6 or more bytes (assuming 69) into a destination of size 64
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~
entry.c: In function 'avahi_server_add_dns_server_address':
entry.c:982:36: warning: '%s' directive output may be truncated writing up to 63 bytes into a region of size 61 [-Wformat-truncation=]
  982 |         snprintf(n, sizeof(n), "ip-%s.%s", h, domain);
      |                                    ^~      ~
In function 'snprintf',
    inlined from 'avahi_server_add_dns_server_address' at entry.c:982:9:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:10: note: '__builtin___snprintf_chk' output 5 or more bytes (assuming 68) into a destination of size 64
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~

```

evverx:
> It does not seem appropriate to continue with truncated output, TBH.

I don't think it has been said anywhere that it's OK for avahi to do what it currently does. The "bug" label was added because it's a bug. It isn't described well enough though and doesn't come with tests. If it did it would probably be noticed that `snprintf` isn't the only issue in `avahi_server_add_dns_server_address`. Either way I'll add the "minor" label to make it clear that it's actually a minor thing unlikely to be seen in practice (which would probably be obvious if the commit messages were less vague) and it isn't a priority in terms of reviewing/merging it.

> With -Wall on Ubuntu 24.04 LTS

There are a gazillion CI jobs where avahi is built with various compilers and so on so it's all transparent and there is no need to confirm that by posting those warnings in comments.

nigels-com:
Agree about the categorisation being minor.  But I do think these sort of things just ought be resolved without so much fuss.  The automated test coverage ought to be giving us confidence that there no unintended consequences or side effects?
Even if only to reduce the "bad smell" of compiler warnings seems reason enough.

evverx:
>  I do think these sort of things just ought be resolved without so much fuss

I don't think that tests covering the code are considered fuss.

> The automated test coverage ought to be giving us confidence that there no unintended consequences or side effects?

How is it going to do that if that code isn't covered?

nigels-com:
We both agree that coverage a good thing to have.
But what's less clear to me is if this particular part of avahi is high priority in terms of testing effort/reward.
I don't know avahi well enough to say.
And this particular change does look low risk indeed, to me.

evverx:
avahi-core is the most important part of the codebase so patches should come with tests. There are parts that aren't important but tests are always welcome (and in the long term they are actually more useful than all these one-off fixes themselves combined).

nigels-com:
Perhaps the contribution docs can be expanded to reflect the priorities and expectations.  Yocto for example is carrying a large set of patches.  I'd personally prefer not to have more patches on top of that, but I'm not sure that I personally expand the test coverage as a precondition.  There are always varied opinions about scoping change requests.

evverx:
> Perhaps the contribution docs can be expanded to reflect the priorities and expectations

Makes sense. I recently added the following rule from LLVM
> Our golden rule is that a contribution should be worth more to the project
than the time it takes to review it.

and it's somewhat indirectly implies that patches with tests are more likely to be accepted because it usually takes less time to review them to make sure they actually do what they are supposed to do but I agree that various things should be spelled out explicitly.

> Yocto for example is carrying a large set of patches. I'd personally prefer not to have more patches on top of that

Given that tests come in separate commits I'm not sure why it would be an issue. It's possible to backport fixes without the corresponding tests.

damiank89:
I updated the original commits with better messages and added some tests in `avahi-test.c`.

damiank89:
@evverx Do you think this is good enough?

PR #828, opened 2026-01-09
Extend avahi-browse to support -4 and -6 CLI options for filtering IPv4 and IPv6 protocols specifically

nigels-com:
I'm debugging some avahi behavior for an in-development embedded device.
I found it helpful to filter out the IPv6 information from avahi-browse, to better focus on IPv4.

Example results:

```
$ ./avahi-utils/avahi-browse -6 _foo._tcp
+   eno1 IPv6 Service on pi5                                _foo._tcp            local
```

```
$ ./avahi-utils/avahi-browse -4 -r _foo._tcp
+   eno1 IPv4 Service on pi5                                _foo._tcp            local
=   eno1 IPv4 Service on pi5                                _foo._tcp            local
   hostname = [pi5.local]
   address = [192.168.8.227]
   port = [0]
   txt = []
```

### DIFF
--- a/avahi-utils/avahi-browse.c
+++ b/avahi-utils/avahi-browse.c
@@ -65,8 +65,7 @@ typedef struct Config {
     char *domain;
     char *stype;
     int ignore_local;
-    int ipv4;
-    int ipv6;
+    AvahiProtocol protocol;
     Command command;
     int resolve;
     int no_fail;
@@ -154,7 +153,7 @@ static char *make_printable(const char *from, char *to) {
 }
 
 static void print_service_line(Config *config, char c, AvahiIfIndex interface, AvahiProtocol protocol, const char *name, const char *type, const char *domain, int nl) {
-    if ((config->ipv4 && protocol==AVAHI_PROTO_INET) || (config->ipv6 && protocol==AVAHI_PROTO_INET6) || (!config->ipv4 && !config->ipv6)) {
+    if (config->protocol == AVAHI_PROTO_UNSPEC || config->protocol == protocol) {
         char ifname[IF_NAMESIZE];
 
 #if defined(HAVE_GDBM)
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR.

Looking at the IPv6 address in the output of `avahi-browse -4` provided in "Example results" can you expand on what your use case is? Can you also expand on why it was decided to add it to `print_service_line`? For example with `-4` avahi-daemon still sends IPv6-related stuff.

Other than that the commit message still says "IPv5" instead of "IPv6", `help()` hasn't been updated and there are no tests exercising those code paths.

nigels-com:
Hello @evverx, apologies for the rushed Friday afternoon PR, I was a bit keen to get my weekend started, I guess.
I'll circle back on this to look at unit testing, but the resolved output, commit message and documentation are fixed now.

```
$ ./avahi-utils/avahi-browse -h
avahi-browse [options] <service type>
avahi-browse [options] -a
avahi-browse [options] -D

    -h --help            Show this help
    -V --version         Show version
    -D --browse-domains  Browse for browsing domains instead of services
    -a --all             Show all services, regardless of the type
    -d --domain=DOMAIN   The domain to browse in
    -v --verbose         Enable verbose mode
    -t --terminate       Terminate after dumping a more or less complete list
    -c --cache           Terminate after dumping all entries from the cache
    -l --ignore-local    Ignore local services
    -4 --ipv4            Show only IPv4 services
    -6 --ipv6            Show only IPv6 services
    -r --resolve         Resolve services found
    -f --no-fail         Don't fail if the daemon is not available
    -p --parsable        Output in parsable format
```

evverx:
`avahi-browse` isn't covered by unit tests. It's run in `smoke-tests.sh` with various options to make sure it doesn't crash. The man page needs updating too: https://github.com/avahi/avahi/blob/master/man/avahi-browse.1.xml.in.

Before that I'm still interested in figuring out what the use case is and how it was tested. Can you add it to the commit message too? In its current form it still sends queries over IPv6 when `-4` is passed to it and I'm not sure whether it's intentional or not.

nigels-com:
My use case is simply "just show me IPv4, ignore IPv6".  I don't have an opinion beyond that, happy to have some guidance there.

Edit:  Oh, `.github/workflows/smoke-tests.sh`, ah!

evverx:
> "just show me IPv4, ignore IPv6"

I'm not sure it's as straightforward as it seems. If IPv6 stuff is thrown out anyway it shouldn't probably send unnecessary queries. Should it show IPv6 addresses advertised over IPv4 when `-4` is passed for example when services are resolved? Currently it does that but I don't think it's what would be expected in general. I think it all depends on the use case in the end.

nigels-com:
I made the change to browse only IPv4 or IPv6 according to the CLI options.
Seems reasonable to me.

Issue #527, opened 2023-11-15
Move away from xmltoman for manpages?


rossburton:
xmltoman is basically unmaintained and uses a bespoke XML format.

As an alternative I suggest either sphinx or docutils: both have restructuredtext input and can generate manpages and more, so the pieces of documentation could be consolidated into a single unit which contains the design overview, manpages, etc.

evverx:
> xmltoman is basically unmaintained and uses a bespoke XML format.

While I agree it isn't ideal I'm not sure why it would be an issue at this point. It kind of works, distros don't plan to throw it away as far as I know. The bespoke format probably makes it harder to change the documentation but it is not that there are a lot of contributors who are particularly interested in that anyway.

Realistically it isn't a priority and I don't think `xmltoman` is going to be replaced with anything else here in the foreseeable future (unless patches are sent probably).

rossburton:
Well, the background is that OpenEmbedded hasn't got a recipe for xmltoman yet (because we've used tarballs of avahi, and nothing else has needed it) and I don't want to add a new recipe to core for a tool that is abandoned.

So I might send a patch :)

evverx:
I get why OpenEmbedded wouldn't want to introduce build dependencies that haven't changed for a while into their build pipelines. I'm not sure the other distros would be happy about that though. They all would have to change their packages to point to a new thing.

Personally I'm fine with switching to something else but I'd wait for the others to chime in so that it would be possible to reach a consensus.

mbiebl:
+1 from the Debian side for moving away from xmltoman

pemensik:
From what I have seen, [BIND9](https://gitlab.isc.org/isc-projects/bind9/) could serve as a nice example of converting from docboox to rst files for manual pages. Could be taken as an inspiration. xml files seems too unconvenient to use for documentation generation today. But it seems to me we have priority in cleaning up the code for now.

evverx:
@pemensik if you are in Europe I think it should be possible to apply for https://www.sovereigntechfund.de/programs/bug-resilience. I think the "direct contribution" part (https://www.sovereigntechfund.de/programs/bug-resilience#direct-contributions) can help here to switch to Sphinx. The program focuses on improving maintainability and security (including I assume supply chain security where abandoned dependencies are frowned upon) and getting rid of xmltoman can help with that. I think they can also help with tests, compiler warnings and stuff like that.

[...]

PR #830, opened 2026-01-10
man: Convert tabs to spaces, administrative only

nigels-com:
There is a mixture of spaces of tabs in the manual pages, convert to spaces only.

### DIFF
--- a/man/avahi-autoipd.8.xml.in
+++ b/man/avahi-autoipd.8.xml.in
@@ -23,12 +23,12 @@
 
   <manpage name="avahi-autoipd" section="8" desc="IPv4LL network address configuration daemon">
 
-	<synopsis>
+    <synopsis>
       <cmd>avahi-autoipd [<arg>options</arg>] <arg>interface</arg></cmd>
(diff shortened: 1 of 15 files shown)


evverx:
Generally the man pages should be converted from xmltoman (https://github.com/avahi/avahi/issues/527) and in that case that churn would be warranted I think. I don't think that replacing stuff that isn't touched very often and should be replace anyway is useful. Sorry.

The same thing regarding the coding style was discussed recently in https://github.com/avahi/avahi/pull/819 and it was closed too.

nigels-com:
It's a nuisance to hand edit files with mixed spaces and tabs.
Potentially makes PR more messy too.
That's the reason for the PR.

evverx:
I saw the man page in https://github.com/avahi/avahi/pull/828 so I more or less understand where it comes from but given that it's just one PR I don't think all the man pages should be reformatted because of that. PRs addressing https://github.com/avahi/avahi/issues/527 would be welcome though.

nigels-com:
As I see it, it's more a matter of leaving things better than how I found them.  And not making perfection the enemy of good enough (or at least better).

Issue #536, opened 2023-11-26
An empty `allow-interfaces=` makes the system not reachable


votdev:
> If set to the empty list all local interfaces except loopback and point-to-point will be used. 

Even if the documentation tells that it should work, it simply does not. I need to uncomment the `allow-interfaces=` line in `/etc/avahi/avahi-daemon.conf` and restart the service with `systemctl restart avahi-daemon.service` to make the system reachable again.

The problem occurs on Debian 12 with the following package versions.

```
# dpkg -l | grep avahi
ii  avahi-daemon                    0.8-10                               amd64        Avahi mDNS/DNS-SD daemon
ii  libavahi-client3:amd64          0.8-10                               amd64        Avahi client library
ii  libavahi-common-data:amd64      0.8-10                               amd64        Avahi common data files
ii  libavahi-common3:amd64          0.8-10                               amd64        Avahi common library
ii  libavahi-core7:amd64            0.8-10                               amd64        Avahi's embeddable mDNS/DNS-SD library
```

votdev:
Sorry, i missed the important word `local` in the docs. So this issue can be closed.

votdev:
I'm sorry, i missed the `all local interfaces` in the docs.

votdev:
After some confusion about what `local` means I have to reopen the issue.

evverx:
Looks like it's a documentation bug. It should probably say that it should be unset to pick up the default interfaces. It's unset in the default config coming with avahi.

tomeq82:
Exactly, this is documentation/functionality bug - empty string "allow-interfaces" makes NO interfaces catched/enumerated thus virtually avahi not working at all. No matter which distro, Avahi 0.8

damiank89:
In my opinion, this isn't a documentation bug. It's expected for an empty value in both `allow-interfaces` and `deny-interfaces` to set the corresponding `server_config` value to `NULL`, but this isn't the case because `avahi_split_csv()` does not return `NULL` for an empty string input. The result is a list of one item (an empty string), instead of an empty list (`NULL`).

This is what causes the bug. `allow-interfaces` is essentially set to one interface, with an empty name.

Not a documentation bug for sure, and can be solved in three ways:
1. Error out when empty values are provided (#833)
2. Warn and skip the line when empty value is provided
3. Change `avahi_split_csv()` to return `NULL` when given an empty string

[...]

PR #833, opened 2026-01-11
Error out when `allow-interfaces` or `deny-interfaces` are empty

damiank89:
This PR fixes https://github.com/avahi/avahi/issues/536.

In my opinion, this "documentation" bug is actually the configuration not being explicit enough. I've changed the parsing logic to deny empty values for these options as it makes no sense to set them to empty values.

Documentation was also updated to reflect these changes.

### DIFF
--- a/man/avahi-daemon.conf.5.xml.in
+++ b/man/avahi-daemon.conf.5.xml.in
@@ -81,9 +81,9 @@
     <option>
       <p><opt>allow-interfaces=</opt> Set a comma separated list of
       allowed network interfaces that should be used by the
-      avahi-daemon. Traffic on other interfaces will be ignored. If
-      set to an empty list all local interfaces except loopback and
-      point-to-point will be used.</p>
+      avahi-daemon. Traffic on other interfaces will be ignored.
+      If unset, all local interfaces except loopback and
+      point-to-point will be used. Empty value is not accepted.</p>
     </option>
 
     <option>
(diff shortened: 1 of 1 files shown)


evverx:
I'll add the "regression" label and close it. Though I wonder why you think it's acceptable to break stuff like
```
allow-interfaces=host0                                                                                                                                        
deny-interfaces=
```
?

damiank89:
Why even bother setting `deny-interfaces=` if it's empty? What actually happens is that `server_config.deny_interfaces` is initialized to a "list" of one item, which is an empty string. The same happens with `allow-interfaces`, which causes this bug. Why did you close my PR? This is a bug, and it's okay if you didn't like my solution, but we can at least discuss it...

The bug originates from `avahi_split_csv()` which returns an empty string when given an empty string, and not `NULL`. We can try and fix it from there, or replace the error I added with a `continue` statement and log a warning (this way it won't cause "regressions").

evverx:
Can you answer the question why you think it's acceptable to break that particular use case?

> Why even bother setting deny-interfaces= if it's empty? 

Did you look at the issue this PR is trying to fix to see the context where `allow-interfaces=` was set?

> This is a bug

That's why https://github.com/avahi/avahi/issues/536 is still open. If you disagree that it's a documentation bug feel free to comment there.

damiank89:
> Can you answer the question why you think it's acceptable to break that particular use case?

Honestly I don't think it's a real use-case. People comment lines they don't want in their configuration. The use-case you provided is buggy for a fact, as it doesn't actually do what you expect it to do in the code, and works by chance. The opposite case, where `allow-interfaces` is empty and `deny-interfaces` isn't doesn't work because of the same bug.

> That's why #536 is still open. If you disagree that it's a documentation bug feel free to comment there.

Okay, I'll comment there.

evverx:
> People comment lines they don't want in their configuration

It isn't necessarily managed by people (as can be seen in https://github.com/openmediavault/openmediavault/commit/0f0e13de419c560d26d25284550fe64369bc5873)

> works by chance

It works though and there can be scripts/configs/whatever setting that and there is absolutely no need to break that stuff for no reason.

Either way the discussion was moved to https://github.com/avahi/avahi/issues/536.

PR #834, opened 2026-01-13
Update service-types

JHentges:
Add aioenet service used by ACCES I/O Products' eNET-AIO16-16F Family of DAQ devices.

### DIFF
--- a/service-type-database/service-types
+++ b/service-type-database/service-types
@@ -222,6 +222,8 @@ _amzn-wplay._tcp:Amazon Fire TV
 
 _qdiscover._tcp:QNAP NAS
 
+_aioenet._tcp:ACCES I/O Products eNET- Server
+
 
 # Other
 


evverx:
Thank you for the PR. I can't find it in https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt. Do you have links to the official documentation where `_aioenet._tcp` is mentioned or something like that that can be added to the commit message?

JHentges:
Thank you for getting back to me, evverx. 
We're just poking our toes into this domain; I didn't think the IANA list applied to us, but freely admit I am an utter novice.  FYI we're using port 18767 (0x494F, "IO", our PCI VendorID).
We are not currently "documenting" the raw service name, other than by providing the source code to our Windows DLL (which provides a high-level discovery function) and Linux image source (and its xml service config file).
The Windows DLL source gets installed by the standard product package installer (https://accesio.com/files/packages/eNET-AIO16-16F%20Install.exe) to this folder (by default): C:\Users\Public\Documents\ACCES\eNET-AIO16-16F\Win32\Driver.SRC\AIOeNET\DLL\ with the discovery function starting in CoreExports.pas on line 413: 
```function ScanForDevices(MaxSize: TUInt32; var NameList: AnsiString): TWinErrCode; cdecl;
const
	QueryStr: WideString = '_aioenet._tcp.local';
	Duration = 1200; //ms

var
	SBR: TDNSServiceBrowseRequest;
	CancelRec: TDNSServiceCancel;
	QueryManager: TDNSQueryManager;

begin
	try
		if not bDNSSDLoaded then begin
			LoadDNSSD;
			if not bDNSSDLoaded then begin
				Result := ERROR_PROC_NOT_FOUND;
				Exit;
			end;
		end;

		QueryManager := TDNSQueryManager.Create;
``` 
If you let me know what kind of documentation regarding our use of _aioenet._tcp would be suitable I can probably craft something.  Perhaps a statement on the product's web page (https://accesio.com/product/enet-aio16-16f/)?

evverx:
Got it. If there are no links it's fine. I asked just to make sure that the service name is associated with those devices and doesn't conflict with the other vendors. The previous reply is enough. I'll merge it once the CI is green.

PR #836, opened 2026-01-16
resolve enum/flag handling 

### LINKED ISSUE #831: Static analyzer warns about enums out of range when setting default/no flags (0)
I have been using avahi in a few different projects (thanks for all the hard work). While scanning our code base with various static analyzers, they all mostly warn when passing `0` to various Avahi functions (as flags). The warning reads:

```
warning: The value '0' provided to the cast expression is not in the valid range of values for the enum [optin.core.EnumCastOutOfRange]
  222 |                         (AvahiPublishFlags) 0, avahi.name,
      |                         ^~~~~~~~~~~~~~~~~~~~~
/usr/include/avahi-common/defs.h:243:9: note: enum declared here
  243 | typedef enum {
      |         ^~~~~~
  244 |     AVAHI_PUBLISH_UNIQUE = 1,           /**< For raw records: The RRset is intended to be unique */
```
enum is defined [here](https://github.com/avahi/avahi/blob/master/avahi-common/defs.h#L243-L257)

code is something like:
```
        ret = avahi_entry_group_add_service(avahi.group,
                        AVAHI_IF_UNSPEC, AVAHI_PROTO_UNSPEC,
                        (AvahiPublishFlags) 0, avahi.name,
                        "_iio._tcp", NULL, NULL, avahi.port, NULL);
```

which is the way it is done in the [example](https://avahi.org/doxygen/html/client-publish-service_8c-example.html). From the static analyzer’s perspective, `0` is not listed in the enum, so it reports a potential “out of range” cast. However, in practice, passing `0` is meaningful, it represents no flags set, and is safely handled by Avahi functions.

Is it possible to go though the enums and have things like `AVAHI_PUBLISH_ZERO = 0`? So that `0` will be a valid enum, and reduce these false errors for everyone?

Happy to send a patch, if someone can provide a suggestion on the suffix that will be acceptable? `_NONE`? `_DEFAULT`? `_ZERO`? `_CLEAR`? `_EMPTY`? (I think my preference is either `_ZERO` or `_EMPTY`)

Thanks in advance.

rgetz:
This PR clarifies and hardens Avahi’s flag and state enums by making bitmask semantics explicit, introducing well-defined ZERO/NONE values, and updating call sites to use those values consistently. Together, these changes improve API clarity, make enum usage unambiguous, and reduce compiler and static-analysis warnings (clang, clang-tidy, clang-analyzer).

Resolves and closed #831, making it easier for downstream projects to compile cleanly.

Bonus commit of avoiding a null pointer in early error paths that no one has seen yet that static analysis pointed out at the same time.

### DIFF
--- a/avahi-gobject/ga-client.c
+++ b/avahi-gobject/ga-client.c
@@ -113,6 +113,7 @@ static void ga_client_get_property(GObject * object,
 static void ga_client_class_init(GaClientClass * ga_client_class) {
     GObjectClass *object_class = G_OBJECT_CLASS(ga_client_class);
     GParamSpec *param_spec;
+    GParamFlags flags;
 
     g_type_class_add_private(ga_client_class, sizeof (GaClientPrivate));
 
@@ -123,29 +124,34 @@ static void ga_client_class_init(GaClientClass * ga_client_class) {
     object_class->set_property = ga_client_set_property;
     object_class->get_property = ga_client_get_property;
 
+    flags = (GParamFlags)(
+                (GParamFlags)G_PARAM_READWRITE |
+                (GParamFlags)G_PARAM_STATIC_NAME |
+                (GParamFlags)G_PARAM_STATIC_BLURB);
+
     param_spec = g_param_spec_enum("state", "Client state",
                                    "The state of the Avahi client",
                                    GA_TYPE_CLIENT_STATE,
                                    GA_CLIENT_STATE_NOT_STARTED,
-                                   G_PARAM_READABLE |
-                                   G_PARAM_STATIC_NAME |
-                                   G_PARAM_STATIC_BLURB);
+                                   flags);
     g_object_class_install_property(object_class, PROP_STATE, param_spec);
+
+    flags = (GParamFlags)(
+                (GParamFlags)flags |
+                (GParamFlags)G_PARAM_CONSTRUCT_ONLY);
 
     param_spec = g_param_spec_enum("flags", "Client flags",
                                    "The flags the Avahi client is started with",
                                    GA_TYPE_CLIENT_FLAGS,
                                    GA_CLIENT_FLAG_NO_FLAGS,
-                                   G_PARAM_READWRITE |
-                                   G_PARAM_CONSTRUCT_ONLY |
-                                   G_PARAM_STATIC_NAME |
-                                   G_PARAM_STATIC_BLURB);
+                                   flags);
     g_object_class_install_property(object_class, PROP_FLAGS, param_spec);
 
     signals[STATE_CHANGED] =
             g_signal_new("state-changed",
                          G_OBJECT_CLASS_TYPE(ga_client_class),
-                         G_SIGNAL_RUN_LAST | G_SIGNAL_DETAILED,
+                         (GSignalFlags)((GSignalFlags)G_SIGNAL_RUN_LAST |
+                                      (GSignalFlags)G_SIGNAL_DETAILED),
                          0,
                          NULL, NULL,
                          g_cclosure_marshal_VOID__ENUM,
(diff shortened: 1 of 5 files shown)


rgetz:
fixed missing header for fedora :(

evverx:
Thank you for the PR.

avahi-autoipd isn't actively maintained anymore (it was discussed in https://github.com/avahi/avahi/issues/467 for example) so the commit should be excluded. Even if it was maintained it would have to be excluded anyway because it's unrelated to the other changes.

Regarding the flags considering that they are supposed be combined I don't think this PR is enough to silence all the warnings without `__attribute__((flag_enum))` somewhere but either way I haven't looked at https://github.com/avahi/avahi/issues/831 closely yet so I'm not sure what the best way to proceed with this would be yet.

rgetz:
dropped the avahi-autoipd - (did you want a PR to rm the directory? - it will always be in the history if you want to go back)

Thanks for the pointer to `__attribute__((flag_enum))` I added that - but it needs a little uglyness before it - since they are system headers, and autoconf magic won't really help downstream projects...

In my project - it doesn't silence all the clang-tidy issues, but it does remove all the '0' warnings, and resolves many of the clang/gcc ones.

evverx:
> did you want a PR to rm the directory?

No I didn't. It should be announced and deprecated properly and things like that.

> it doesn't silence all the clang-tidy issues, but it does remove all the '0' warnings

I took a quick look at that issue. I'm not sure why it says that `(AvahiPublishFlags)` is used in the examples in the avahi codebase (because it's just `0` in `client-publish-service`, `avahi-publish` and so on) but the issue is more or less understood.

I think it's worth mentioning that issues like that are too high a bar for the avahi codebase at this point in general so they are on the backburner generally in terms of code review and so on and I'd prefer it if it was discussed in the issue first along with the ways the CI should be adjusted and so on.

Then again in terms of priorities it's far from the top considering the infinite stream of crashes the fuzz targets have discovered, Coverity findings and so on. The tooling is being set up slowly and clang-tidy/analyzer is just too high a bar. Given it has always been this way and it's the first time it has been reported I suspect it's not a priority in a lot of places either.

rgetz:
I will obviously defer to your judgement - thanks for the feedback and the quick look. I understand that time is short and not all issues are equal. It was an itch that I wanted to scratch - but that doesn't mean that it's valuable to everyone. Feel free to nack or close this issue.

Are you looking for more test cases (similar to what's in the tree already with the *-test.c files?)

evverx:
I think it would make sense to keep it open for now. If nothing else it shows that there is some interest in tidying up the enums. As far as I can see `-Wswitch-enum` is turned on here and it's a bit different from https://github.com/avahi/avahi/issues/831 (which is about `optin.core.EnumCastOutOfRange` and closer to `-Wassign-enum` probably).

> Are you looking for more test cases

In the context of PRs addressing various warnings the first thing would be to turn on those checks. I brought back clang-analyzer yesterday in https://github.com/avahi/avahi/pull/838 but `optin.core.EnumCastOutOfRange` isn't turned on there by default. As soon as they are on it should be possible to see whether issues are fully addressed (and for example https://openscanhub.fedoraproject.org/task/93856/log/added.html shows that this PR hasn't addressed all the `-Wswitch-enum` warnings). Apart from silencing warnings it would of course be great if the affected code was actually exercised at runtime to for example let ASan/UBSan and Valgrind confirm that it's all fine.

rgetz:
Thanks - I didn't have `avahi-compat-howl` or `avahi-compat-libdns_sd` turned on when I did the analysis/made the patch - I can update and make a new version...

rgetz:
build with `avahi-compat-howl` and `avahi-compat-libdns_sd` turned on - fixed the remaining `-Wswitch` issues pointed out (added the missing headers in fedora) - no new issues on rawhide.

rgetz:
turned on `-Wassign-enum`, and fixed all the issues (not in avahi anymore, but all the similar issues in `gobject` which suffers from the same - so you can see what downstream users of avahi need to do) until this is resolved.

PR #823, opened 2026-01-01
Fix GCC implicit-fallthrough compilation warnings in avahi-ui

damiank89:
Move comments outside curly-braces for GCC to detect them.

This is part of #820.

PR #842, opened 2026-01-24
fix fallthrough warnings

rgetz:
Modern compilers increasingly warn about implicit switch fallthroughs. While Avahi already documents many intentional fallthroughs with comments, these are not recognized by the compiler. :(

This change:
- Preserves existing behavior
- Documents intent in a machine-checkable way
- Reduces noise from compiler warnings on newer toolchains
- Avoids impacting older compilers or ABI

It eliminates:
```
../../avahi-common/domain.c:560:13: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
  560 |             case DOMAIN:
      |             ^
../../avahi-common/domain.c:560:13: note: insert '__attribute__((fallthrough));' to silence this warning
  560 |             case DOMAIN:
      |             ^
      |             __attribute__((fallthrough)); 
../../avahi-common/domain.c:560:13: note: insert 'break;' to avoid fall-through
  560 |             case DOMAIN:
      |             ^
      |             break; 
```

It does this by adding a `AVAHI_GCC_FALLTHROUGH` macro:
- A new macro is added to `avahi-common/gccmacro.h`
- Expands to `__attribute__((fallthrough))` on GCC ≥ 7 and Clang
- Expands to a no-op on older compilers

This provides a consistent, portable way to explicitly mark intentional fallthroughs.

The new macro is added at existing intentional fallthrough sites in:
- `avahi-client`
- `avahi-common`
- `avahi-ui`
- `avahi-utils`
- `example code`

These locations already contained comments such as “`/* Fall through */`”. The macro supplements those comments and allows supported compilers to suppress `-Wimplicit-fallthrough` warnings (which is turned on by `-Wextra`) which is turned on by default for gcc/clang.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -311,6 +311,13 @@ if test "x$GCC" = "xyes" ; then
 
     DESIRED_FLAGS="-Wall -W -Wextra -pedantic -pipe -Wformat -Wold-style-definition -Wdeclaration-after-statement -Wfloat-equal -Wmissing-declarations -Wmissing-prototypes -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wshadow -Wendif-labels -Wpointer-arith -Wbad-function-cast -Wcast-qual -Wcast-align -Wwrite-strings -fdiagnostics-show-option -Wno-cast-qual -fno-strict-aliasing"
 
+    extra_flag="-Wimplicit-fallthrough=5"
+    if $CC --version 2>/dev/null | grep -q clang; then
+        # clang doesn't take the "=5" that gcc does
+        extra_flag="-Wimplicit-fallthrough"
+    fi
+    DESIRED_FLAGS="$DESIRED_FLAGS $extra_flag"
+
     if test "x$HAVE_NETLINK" = "xyes" ; then
         # Test whether rtnetlink.h can be included when compiled with -std=c99
         # some distributions (e.g. archlinux) have broken headers that don't


evverx:
> Modern compilers increasingly warn about implicit switch fallthroughs

They have been complaining for some time and there is a PR addressing the gcc warnings for example: https://github.com/avahi/avahi/pull/823. As mentioned there avahi-ui is probably on its way out so I don't think it makes sense to touch it (and if it was maintained this PR wouldn't be complete without that PR since the attributes and the comments aren't exactly in sync).

> suppress -Wimplicit-fallthrough warnings (which is turned on by -Wextra) which is turned on by default for gcc/clang

I don't think it's turned on by default for clang. There are a bunch of clang CI jobs here and "unannotated fall-through" can't be seen anywhere so it should probably be turned on on the CI to make sure that clang no longer complains. Speaking of clang is there a reason the gcc version is checked and the clang version isn't? I don't think that attribute was there before clang-10 was released (either way it would be great if links to clang/gcc release notes/commits where the attribute was announced were added to the commit message).

rgetz:
>I don't think it's turned on by default for clang. 

My choice of "default" was bad. I mean clang passes the gcc tests in the current autoconf system, and it's turned on [here](https://github.com/avahi/avahi/blob/master/configure.ac#L312).

I thought clang supported it for a long time - but it looks like that was `[[clang::fallthrough]];` not the `gcc __attribute__()` version (which was added in [clang 10](https://releases.llvm.org/10.0.0/tools/clang/docs/AttributeReference.html#fallthrough)).

If it's OK - I will just do this:
```
#if defined(__clang__)
#  if __has_attribute(fallthrough)
#    define AVAHI_GCC_FALLTHROUGH __attribute__((fallthrough))
#  else
#    define AVAHI_GCC_FALLTHROUGH /* fallthrough */
#  endif
#elif defined(__GNUC__) && (__GNUC__ >= 7))
#  define AVAHI_GCC_FALLTHROUGH __attribute__((fallthrough))
#else
#  define AVAHI_GCC_FALLTHROUGH /* fallthrough */
#endif
```

I will remove the -ui portions (sorry - that is the 2nd time you told me that).

as for #823 - I think with the above improved macro - you shouldn't need the comments at all (even for old versions of gcc), and I can remove them. I will test to ensure. Maybe @damiank89  can also comment.

evverx:
> My choice of "default" was bad. I mean clang passes the gcc tests in the current autoconf system, and it's turned on [here](https://github.com/avahi/avahi/blob/master/configure.ac#L312).

`-Wextra` is turned on there but it doesn't automatically turn on `-Wimplicit-fallthrough` for clang and logs emitted by clang jobs like https://github.com/avahi/avahi/actions/runs/21326750921/job/61385343421 can confirm that. The warning mentioned in the description is obviously emitted by clang so I'm still not sure what is meant by "default". That's important in the context of the CI.

Regardless everything else if you plan to change this please add the rationale behind those decisions to the commit messages so that it wouldn't be necessary to guess what's going on.

damiank89:
> Maybe @damiank89 can also comment.

In my opinion these comments make the code easier to understand as they makes the fall-throughs look intentional. With that being said, I don't have a strong preference, as long as the warnings are fixed.

rgetz:
I was wrong - `-Wimplicit-fallthrough=3` turned on by gcc with [`-Wextra`](https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/Warning-Options.html#index-W) (knew I saw it somewhere...)

This removes the `avahi-ui` (as requested) - which were the only ones that #823 fixed. 

This also removes the (now) redundant comments to try to keep things clean.

evverx:
> -Wimplicit-fallthrough=3 turned on by gcc with [-Wextra](https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/Warning-Options.html#index-W) (knew I saw it somewhere...)

gcc does that and it's covered by the CI currently. The point I'm trying to make is that in its current form clang isn't covered by the CI and `-Wimplicit-fallthrough` should be added explicitly to make sure both gcc and clang are covered. Since the comments are removed I guess `-Wimplicit-fallthrough=5` can be used to enforce the attribute.

> This removes the avahi-ui (as requested) 

I don't think it does. It's in the diff.

> This also removes the (now) redundant comments to try to keep things clean.

Please add the rationale behind this decision and all the other decisions to the commit messages. How safe is it to drop those comments in terms of working with older toolchains? Also I'm not sure where `cplusplus` came from. Why was it added?

rgetz:
>I don't think it does. It's in the diff.

you are correct. fixed now. Problem between chair and keyboard on my side. Sorry.

>The point I'm trying to make is that in its current form clang isn't covered by the CI and -Wimplicit-fallthrough should be added explicitly

got it - done as a separate commit in this series. (now). complicated since gcc and clang take different options :( 

>Also I'm not sure where cplusplus came from. Why was it added?

when I saw the cpp code in `./avahi-qt/`, and thought to be proactive - but in looking at other PRs/issues - that's also one of the things that isn't actively maintained, so I removed it.

Now all you get is warnings in the `avahi-ui` if that is turned on.

rgetz:
As mentioned in #851:
>gccmacro.h is installed along with libavahi-common and it shouldn't be polluted with all those macros.

Agree/Ack - I didn't realize those were installed.

I can make a new `avahi-common/internalmacros.h` ?
 - `private_macros.h`
 - `build_macros.h`
 - `private.h`
 
 The alternative - is to put them into `config.h.in` which is included almost everywhere, but not installed - but I don't really like that idea...

PR #843, opened 2026-01-24
tests: increase coverage in avahi-common/utf8.c to 100%

rgetz:
Add additional UTF-8 test cases covering edge conditions that were previously untested. These include:

- Overlong encodings (invalid)
- Bad continuation bytes
- Illegal leading bytes
- UTF-16 surrogate halves (U+D800–U+DFFF)
- Unicode noncharacters (U+FFFE, U+FFFF)
- Valid 4-byte characters (including emoji)
- Codepoints above Unicode max
- Truncated UTF-8 sequences

These tests raise code coverage in avahi-common/utf8.c from 81.8% to 100% without changing any functional behavior.

I know this is mostly trivial - but I thought I would start working through `avahi-common` increasing coverage, and wanted to make sure I was following the acceptable upstream process first.

### DIFF
--- a/avahi-common/utf8-test.c
+++ b/avahi-common/utf8-test.c
@@ -29,6 +29,7 @@
 
 int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
 
+    /* Functional tests */
     assert(avahi_utf8_valid("hallo"));
     assert(avahi_utf8_valid("1234567890."));
     /* same word in iso-8859-1 as utf-8 below. */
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR.

> These tests raise code coverage in avahi-common/utf8.c from 81.8% to 100%

How was it measured? Without this PR the full coverage report says that the coverage is 93.94%. The fuzz targets cover it completely https://storage.googleapis.com/oss-fuzz-coverage/avahi/reports/20260122/linux/src/avahi/avahi-common/report.html but they don't make sure it works correctly of course.

Just to clarify I really think that PRs increasing coverage are helpful but I'm curious how it's measured and whether the tests are generated automatically (with consolic execution or anything like that).

evverx:
Regarding the upstream process it's documented in https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md. It documents the coding style and LLM usage currently (so if for example stuff is LLM-generated it should be mentioned explicitly). Based on recent feedback It should be extended to mention that it's expected that PRs come with tests and so on but I haven't gotten round to it yet.

rgetz:
I was measuring coverage with:
```
# configure project
../configure --disable-python --disable-nss --disable-gtk --disable-qt3 --disable-qt4 --disable-qt5 --disable-mono  --disable-manpages --enable-tests --enable-compat-howl --enable-compat-libdns_sd
# make and run built in tests
make && make check
# check coverage of built in tests
lcov --capture --directory . --output-file avahi.info --ignore-errors inconsistent
# remove actual test programs, and system include files, don't care if those are covered or not
lcov --remove avahi.info '/usr/*' '*-test.c' --output-file avahi.filter.info --ignore-errors inconsistent
# generate human readable report
genhtml avahi.filter.info --output-directory avahi-html --ignore-errors inconsistent
```

and checking the resulting html pages.

Tests were generated based on the above, (manually examining where existing tests were not covering), adding more negatives tests, inspired (but not copied) from [World Wide Web Consortium](https://www.w3.org/2001/06/utf-8-wrong/UTF-8-test.html). And I think I found the utf8 encoding for pile of poo at [Compart GmbH](https://www.compart.com/en/unicode/U+1F4A9) or a google search, that maybe gemini answered.

evverx:
Can you add it to the commit message and clarify that it's talking about the line coverage? The branch coverage isn't 100%.

rgetz:
Because it's so small - I just added a few more tests to increase branch coverage (to 100%). (thanks for poke/tip).

avahi-common - utf8.c (as reported by gcov/lcov)			
```
Lines:	    100.0 %	33	33
Functions:	100.0 %	1	1
Branches:	100.0 %	30	30
```

tried to be more descriptive in commit message - happy to thin it if you want less.

evverx:
In terms of coverage the most important PR at this point is https://github.com/avahi/avahi/pull/808 and it comes with no tests as usual. There are tests exercising happy paths (and combined with Valgrind they already caught a bug on FreeBSD (which was unfortunately addressed by silencing Valgrind at first)) but the corner cases aren't covered. If there is interest in expanding the smoke tests to cover those corner cases and exercising those code paths that would be great.

rgetz:
I started flushing out the parts in avahi-common (though they would be the most straightforward - but as you allude to - they are already likely the most tested, and are unlikely to have undiscovered issues, making them - less valuable? - to work on).

I might (if it's OK) finish up some of the in progress things I have, send them upstream and then look at the subsystems you pointed to in the PR.

evverx:
I didn't mean to say that covering avahi-common is not valuable. It's valuable and PRs are appreciated. It wasn't actually tested anywhere up until recently and that's where relatively recent commits like 93b14365c1c1e04efd1a890e8caa01a2a514bfd8, 905028bdf6510033d57aeb119df15b7915ba8b16, fb981c22af3dc501d250e833dce45e5cde66b5d2 come from. It's just that PRs are on the backburner (unless they uncover and fix new crashes or something like that).

PR #844, opened 2026-01-25
Fix logspam issue for IPv6 addresses in avahi-core

### LINKED ISSUE #668: [logspam] avahi-daemon repeatedly registering the same IPv6 addresses without withdrawing them first.
**Avahi version**
0.8 (debian patch 13+b3)
**Distribution**
Debian trixie
**Describe the bug**
logspam of "registering address" every few seconds, appearing to correlate with RAs being received. Appears to be a different cause to #41 as I don't have any deprecated addresses present or advertised, and see no 'withdrawing' messages.

**To Reproduce**
Suspect that reproduction might be as simple as "it happens every time the IP addresses on an interface are updated, even if they're the same"

**Expected behavior**
avahi to not spam the logs by continually re-registering the same addresses once a minute, or if not possible to detect that, downgrade the relevant logging to debug level.

**Avahi config files if applicable**
(`/etc/avahi/avahi-daemon.conf`, `/etc/avahi/hosts`, `/etc/avahi/services/*.service`)

**Avahi debug logs if applicable**
No additional debug output is printed after startup.

**Additional context**
systemd-networkd is unconditionally reassigning the addresses to the interface whenver an RA is received, confirmed via `ip monitor` (which makes perfect sense, as that would incidentally update any expiration timers; I've also observed this exact behavior in networkd debug logs elsewhere) and avahi's conditionals around when to register an address are apparently missing a case for "address already registered", or that check occurs after the log message is printed and simply makes the registration a no-op.

The issue appears to also exist with IPv4, likely due to the common code path, as I noticed a stray "registering address [static ipv4 that's been there since startup]" without an accompanying withdraw while digging through logs for this.
Given that, and the file where the log message is printed having not been touched in 7 years, this behavior has probably been around for a _very_ long time and was likely not noticed (or simply not considered problematic) due to the much lower volume of DHCP requests in most networks vs potentially-unsolicited RAs (and potentially that other network managers don't unconditionally spray updates at netlink even when nothing's actually *changed*...)

damiank89:
Fixes #668.

### DIFF
--- a/avahi-core/iface-linux.c
+++ b/avahi-core/iface-linux.c
@@ -184,6 +184,7 @@ static void netlink_callback(AvahiNetlink *nl, struct nlmsghdr *n, void* userdat
         AvahiAddress raddr, rlocal, *r;
         int raddr_valid = 0, rlocal_valid = 0;
         uint32_t flags = ifaddrmsg->ifa_flags;  /* may be overridden by IFA_FLAGS */
+        int addr_changed = 0, global_scope, deprecated;
 
         /* We are only interested in IPv4 and IPv6 */
         if (ifaddrmsg->ifa_family != AF_INET && ifaddrmsg->ifa_family != AF_INET6)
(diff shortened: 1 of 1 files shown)


damiank89:
Force pushed to trigger CI.

PR #849, opened 2026-01-29
Fix 32-bit overflow for avoiding year 2038 problem

GermanAizek:
`unsigned int` 32-bit dangeours use for `msec` and `jitter`, since it is multiplied by 1000, judging by code before PR changes.

References:
- https://en.wikipedia.org/wiki/Year_2038_problem
- https://www.gnu.org/software/gnulib/manual/html_node/Avoiding-the-year-2038-problem.html

### DIFF
--- a/avahi-common/timeval.c
+++ b/avahi-common/timeval.c
@@ -83,15 +83,15 @@ AvahiUsec avahi_age(const struct timeval *a) {
     return avahi_timeval_diff(&now, a);
 }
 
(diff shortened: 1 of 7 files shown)


evverx:
If you think there is an issue please open an issue first and describe it precisely. "unsigned 32-bit dangeours use for msec and jitter" is too vague.

Also there is at least one PR you opened where you didn't answer the questions you were asked. Please follow up there first and see https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md.

Issue #845, opened 2026-01-26
intent of avahi-common/strlst.c:avahi_string_list_get_service_cookie()


rgetz:
in [`avahi-common/strlst.c:avahi_string_list_get_service_cookie()`](https://github.com/avahi/avahi/blob/master/avahi-common/strlst.c#L556), there is a call to 
```
ret = (uint32_t) strtoll(value, &end, 0);
```

which immediately takes a  `long long` and jams it into a uint32_t, without checking if there are errors (`errno == ERANGE`). This allows you get alot of MAX_UINT on unbounded inputs.

Is this on purpose or a miss?

normally it would be something like (where you only push it when you know if fits).

```
errno = 0;
value = strtoll(buf, &end, 0);
if (end == buf || value < 0 || errno == ERANGE || value > UINT_MAX) {
        avahi_free(value);
        return AVAHI_SERVICE_COOKIE_INVALID;
    }
ret = (uint32_t) value;
```
right now (at least in theory) you can get alot of values, which will return the same number, and will not survive a round trip.

Thanks
-Robin

evverx:
It's used in the context of cookies added in https://github.com/avahi/avahi/blob/1b0dd2a4956398749ecbdd5b8e1f0e95adc6b900/avahi-core/entry.c#L566
and it's normally `uint32_t`
https://github.com/avahi/avahi/blob/1b0dd2a4956398749ecbdd5b8e1f0e95adc6b900/avahi-core/internal.h#L157-L158

> right now (at least in theory) you can get alot of values, which will return the same number, and will not survive a round trip

Where did those round trips come from? Is that in the context of the tests mentioned in https://github.com/avahi/avahi/pull/843 to increase the coverage of avahi-common?

rgetz:
Yes. Because it's string parsing, and the string can be any collection of text in a test/synthetic example - it's easy to make something > MAX_UINT...

Plus it gets flagged from tools (like MathWorks [Polyspace](https://www.mathworks.com/products/polyspace.html)) that see strtoll with no checks of errno afterwards (which is recommended/required in the [man page](https://man7.org/linux/man-pages/man3/strtol.3.html)).

Since strtoll() can legitimately return 0, LLONG_MAX or LLONG_MIN on both success and failure, the calling program should set errno to 0 before the call, and then determine if an error occurred by checking whether errno == ERANGE after the call.

We aren't doing this today.

evverx:
[...]

PR #851, opened 2026-01-30
Expand  avahi-common/strlst.c test coverage and fix sanitizer-reported defects afterwards.

rgetz:
None of the fixes address known real-world failures or regressions; they eliminate edge-case undefined behavior, leak paths, and error-handling weaknesses that are only observable under stress, fault injection, or malformed inputs, partially described in #845 - all should be considered "minor" - and given the appropriate priority.

### Overview

This PR expands test coverage for the AvahiStringList API and uses that coverage to identify and fix several correctness issues surfaced by sanitizers afterwards.

- Drive AvahiStringList through all functional, error, and cleanup paths
- Run the resulting code under ASan / UBSan (GCC + Clang)
- Fix only the issues proven by tests and tools
- No behavioral changes were made (unless the previous behavior was incorrect, undefined, or unsafe - tightening the API).

### Test coverage

The strlst test suite has been expanded to exercise:
- All public entry points and helper APIs
- Argument combinations, NULL-handling, and cleanup paths
- Varargs and formatting helpers
- Empty strings, binary data, embedded NULs, and truncation
- Deterministic low-memory failure paths

Coverage results (was 69.25% of lines)
 - GCC: 100% line coverage, 100% branch coverage just with this test
- Clang: 100% line coverage, ~100% branch coverage just with this test
(single gap is compiler-specific diagnostics)
 - (100% coverage assumes not including "assert" lines :) )

Tests pass cleanly under:
- AddressSanitizer (ASan)
- UndefinedBehaviorSanitizer (UBSan)

### Issues fixed

The expanded coverage and sanitizer runs exposed several latent issues, which are fixed in this series:
- Memory leaks on partial allocation failure
- Unsafe retry behavior in `vsnprintf()` error cases (previously led to OOM)
- Incorrect handling of numeric conversion errors
- Minor clarity issues that triggered static analysis warnings

All fixes are narrow, test-driven, and limited to correctness and safety.

**Test-only safeguards** (for this and other planned tests)

To keep tests viable as compilers become more aggressive at compile-time
diagnostics, this PR introduces test-only diagnostic suppression macros.
These are:
- Available only when building tests
- Never usable in production or library code
- Explicitly documented as test-only

This allows negative and invalid cases to be tested without weakening normal build guarantees, and still build tests without warnings.

### Impact
 - No public API changes
 - No semantic changes except where behavior was previously undefined or unsafe
 - Production code remains warning-clean and sanitizer-clean
 - Tests are deterministic and self-contained

### DIFF
--- a/avahi-common/strlst-test.c
+++ b/avahi-common/strlst-test.c
@@ -22,46 +22,142 @@
 #endif
 
 #include <stdio.h>
+#include <stdlib.h>
 #include <assert.h>
 #include <string.h>
 
+#include "defs.h"
 #include "strlst.h"
 #include "malloc.h"
+#include "gccmacro.h"
 
 /*
+ **************************************************************
+ * mock allocator that fails after n malloc/realloc calls.
+ * Used to test low-memory paths and error handling
+ ***************************************************************/
+typedef struct FailAfterState {
+    int remaining;
+} FailAfterState;
+
+static FailAfterState fail_after_state;
+
+static void* fail_after_malloc(size_t size) {
+    if (fail_after_state.remaining <= 0)
+        return NULL;
+
+    fail_after_state.remaining--;
+    return malloc(size);
+}
+
+static void* fail_after_realloc(void *ptr, size_t size) {
+    if (fail_after_state.remaining <= 0)
+        return NULL;
+
+    fail_after_state.remaining--;
+    return realloc(ptr, size);
+}
+
+static void fail_after_free(void *ptr) {
+    free(ptr);
+}
+
+static AvahiAllocator fail_after_allocator = {
+    .malloc  = fail_after_malloc,
+    .realloc = fail_after_realloc,
+    .free    = fail_after_free
+};
+
+static const AvahiAllocator* avahi_allocator_fail_after(int n) {
+    fail_after_state.remaining = n;
+    return &fail_after_allocator;
+}
+
+/************* end fake memory allocator *********/
+
+/***************************************************************
  * Build a representative AvahiStringList containing:
  *  - plain strings
  *  - empty strings
  *  - key/value pairs
  *  - arbitrary binary data
  *  - escaped / non-printable characters
- *
- * This list is reused across most tests.
  */
 static AvahiStringList* build_test_string_list(void) {
-    AvahiStringList *a = NULL;
+    AvahiStringList *a = NULL, *p;
+    const char *expected_texts[] = {
+        "end", "gh_issue_169=\x1f\x20\x7e\x7f\xff",
+        "i am a \"string\" with embedded double-quotes (\\\")\nand newlines (\\n).",
+        "uxknurz2=blafasel\0oerks","uxknurz=","blubb=blaa","seven=7 x",
+        "null\0null","","quux","","","bar","foo=99","start","b","a","prefix",
+    };
+    size_t expected_sizes[] = {
+        3, 18, 67, 23, 7, 10, 9, 9, 0, 4, 0, 0, 3, 6, 5, 1, 1, 6,
+    };
+    size_t i = 0;
 
-    a = avahi_string_list_new("prefix", "a", "b", NULL);
+    /* Basic list with a few strings */
+    a = avahi_string_list_new("prefix", "a", "b", (const char*)NULL);
+    p = a;
+    assert(avahi_string_list_length(a) == 3);
+    assert(p && strcmp((const char*)avahi_string_list_get_text(p), "b") == 0);
+    p = p->next;
+    assert(p && strcmp((const char*)avahi_string_list_get_text(p), "a") == 0);
+    p = p->next;
+    assert(p && strcmp((const char*)avahi_string_list_get_text(p), "prefix") == 0);
+    p = p->next;
 
+    /* Append more strings */
     a = avahi_string_list_add(a, "start");
+    assert(strcmp((const char*)avahi_string_list_get_text(a), "start") == 0);
+    assert(avahi_string_list_length(a) == 4);
+
     a = avahi_string_list_add(a, "foo=99");
     a = avahi_string_list_add(a, "bar");
     a = avahi_string_list_add(a, "");
     a = avahi_string_list_add(a, "");
     a = avahi_string_list_add(a, "quux");
     a = avahi_string_list_add(a, "");
+    assert(avahi_string_list_length(a) == 10);
+
+    /* Traverse and check the last added "quux" */
+    p = a;
+    while (p->next) p = p->next;
+    assert(strcmp((const char*)avahi_string_list_get_text(a), "") == 0);
+    assert(strcmp((const char*)avahi_string_list_get_text(p), "prefix") == 0);
+
+    /* Arbitrary binary data */
     a = avahi_string_list_add_arbitrary(a, (const uint8_t*) "null\0null", 9);
+    assert(avahi_string_list_get_size(a) == 9);
+
+    /* Formatted string */
     a = avahi_string_list_add_printf(a, "seven=%i %c", 7, 'x');
+    assert(strcmp((const char*)avahi_string_list_get_text(a), "seven=7 x") == 0);
+
+    /* Key/value pairs */
     a = avahi_string_list_add_pair(a, "blubb", "blaa");
-    a = avahi_string_list_add_pair(a, "uxknurz", NULL);
+    a = avahi_string_list_add_pair(a, "uxknurz", (const char*)NULL);
     a = avahi_string_list_add_pair_arbitrary(
             a, "uxknurz2", (const uint8_t*) "blafasel\0oerks", 14);
-    a = avahi_string_list_add(
-            a,
-            "i am a \"string\" with embedded double-quotes (\\\")\n"
-            "and newlines (\\n).");
+    assert(avahi_string_list_get_size(a) == 14 + strlen("uxknurz2="));
+
+    /* Escaped / special characters */
+    a = avahi_string_list_add(a,
+        "i am a \"string\" with embedded double-quotes (\\\")\n"
+        "and newlines (\\n).");
     a = avahi_string_list_add(a, "gh_issue_169=\x1f\x20\x7e\x7f\xff");
     a = avahi_string_list_add(a, "end");
+
+    assert(avahi_string_list_length(a) == 18);
+
+    for (p = a; p; p = p->next, i++) {
+        assert(p);
+        assert(avahi_string_list_get_size(p) == expected_sizes[i]);
+        /* Only compare text for printable strings, skip comparison for binary */
+        if (expected_sizes[i] == strlen(expected_texts[i]))
+            assert(strcmp((const char*)avahi_string_list_get_text(p), expected_texts[i]) == 0);
+    }
+    assert(i == sizeof(expected_texts)/sizeof(expected_texts[0]));
 
     return a;
 }
@@ -81,6 +177,46 @@ static void test_string_rendering(AvahiStringList *a) {
     assert(strstr(t, "gh_issue_169=\\031 ~\\127\\255"));
     avahi_free(t);
 }
+
+
+/**
+ * Test avahi_string_list_add_arbitrary() for a range of string sizes.
+ *
+ * This test iterates over string sizes from 1 to 256 bytes, creating
+ * a new AvahiStringList node for each size. It verifies:
+ *   - The returned pointer is non-NULL.
+ *   - The size field in the node matches the requested size.
+ *   - The contents of the node match the input data.
+ *
+ * Any allocation failure aborts the test immediately.
+ * This helps ensure that edge cases, including maximum single-byte
+ * length strings, are correctly handled.
+ */
+static void test_avahi_string_list_add_arbitrary_sizes(void) {
+    AvahiStringList *l = NULL, *n;
+    uint8_t buf[256];
+
+    for (size_t i = 1; i <= 256; i++) {
+        // Fill buffer with predictable data
+        for (size_t j = 0; j < i; j++)
+            buf[j] = (uint8_t)(j & 0xFF);
+
+        // Add to list
+        n = avahi_string_list_add_arbitrary(l, buf, i);
+        assert(n);  // allocation must succeed
+
+        // Check contents
+        for (size_t j = 0; j < i; j++)
+            assert(n->text[j] == (uint8_t)(j & 0xFF));
+
+        // Move to next
+        l = n;
+    }
+
+    // Clean up
+    avahi_string_list_free(l);
+}
+
 
 /*
  * Serialize the list to wire format, dump it for debugging,
@@ -171,7 +307,7 @@ static void test_find_and_get_pair(AvahiStringList *a) {
 }
 
 /*
- * Edge case: serializing and parsing a NULL list.
+ * serializing and parsing a NULL list.
  */
 static void test_null_list_serialization(void) {
     uint8_t data[1024];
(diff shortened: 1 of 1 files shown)


rgetz:
CI issues seem to be dns failures?
`>>> Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject`

evverx:
I'm afraid in its current form it's hard to review this so it should be split one way or another. I don't think I'm going to review this anytime soon closely but there are some things that need looking into.

`gccmacro.h` is installed along with `libavahi-common` and it shouldn't be polluted with all those macros (it affects the PR where the fallthrough attribute is being added as well but it hasn't gotten to the point where it can be discussed yet).

Allocation failures aren't intrinsic to strlst so it shouldn't be tied to that one test only. Generally in terms of priorities there are plans to inject allocation failures into the fuzzing infrastructure upstream but it's not a priority (and it's going to be done using a separate thing used by projects like suricata, nDPI and so on on OSS-Fuzz).

In the meantime looking at this PR I pushed https://github.com/avahi/avahi/commit/b10d81d7b0dbeb41cc5afc716074d97ce55d7d57 to run the unit tests under Valgrind as well. The smoke tests run everything under ASan/UBSan and Valgrind and it turned out that the unit tests were run under ASan/UBSan only.

Either way I'd need to think how to split into reviewable chunks.

evverx:
In the meantime can you add the ASan/UBSan backtraces to the commit messages and also take a look at https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md? The curly braces and things like that are inconsistent. If it's LLM-assisted please mention it explicitly in the commit messages as well.

rgetz:
>I'm afraid in its current form it's hard to review this so it should be split one way or another.

Certainly up for suggestions. I can split it into multiple (separate PR for each commit into the core code), and then a larger one for the test adds if that's easier (likely make more sense/easier for discussion). I wasn't sure which was less work on your side. (look at multiple things vs one larger PR).

Didn't realize you wanted the backtraces in the commit messages - that's easy.

>curly braces and things like that are inconsistent.

yeah - I have been playing with vim settings to try to get things set up - most of the projects I work on are more kernel style, and need fix that.

evverx:
> I wasn't sure which was less work on your side

I'm not sure yet either. The parts where failures are injected need discussion and it can take a while in the context of the whole test suite as opposed to this one test. It shouldn't probably delay commits that don't depend on that so they can go to separate PRs (for example a fix along with its test would be easier to review than the same fix followed by a huge change to the test). I'd need to think about this. Then again in terms of injecting failures the fuzzing infrastructure is more productive and has a priority on my side.

> Didn't realize you wanted the backtraces in the commit messages

I do. Compiler warnings, Valgrind/ASan/UBsan backtraces and so on go to commit messages like 2d48e42d44a183f26a4d12d1f5d41abb9b7c6355. The latest example is 91bf04085a2399abbc23693ef386a27142b862fe and it can be compared with the downstream fix (https://github.com/NetBSD/pkgsrc/commit/e6a3c3e4ea4da2ff5e4640c9c435e155ad7c13ad) with just "Fix -Wint-conversion". It the warning would have been included it would have probably been easier to spot the issue during the review there and fix it differently.

rgetz:
> Then again in terms of injecting failures the fuzzing infrastructure is more productive and has a priority on my side.

Ack/Agree - fuzzing is definitely the most efficient way to inject failures at scale and discover unexpected edge cases, so it’s great for broad exploration and I agree it has strong value there.

My intent with the directed tests isn’t to compete with fuzzing, but to complement it. Directed failure injection gives us determinism and precision: we can reliably hit specific error-handling branches, document intended behavior, and lock in regressions once something is found. The fault injection I added is intentionally very deterministic (essentially the opposite of what you want from fuzzing, “fail exactly on the 3rd `malloc()`.”) That level of control is useful for making sure a specific cleanup or rollback path is always exercised and immediately reproducible in CI.

Fuzzing, on the other hand, benefits from randomness and scale, it’s the discovery engine for “unknown unknowns” and emergent interactions. Directed tests act more like executable specifications and long-term guardrails. The benefit of keeping some explicit failure injection in directed tests, rather than leaving it entirely to fuzzing, is that critical failure paths remain consistently covered, understandable to future contributors, and fast to verify. Ideally the two reinforce each other: fuzzing finds new classes of issues, and targeted tests make sure the important ones stay fixed.

I see directed tests as part of the API contract, they describe what the code must do under specific, known working and failure conditions, and look at functionality (correctness, error codes, etc). Fuzzing is more of a journey of discovery, exploring what the code might encounter. Both are valuable, but IMHO - they serve different roles: one defines the guarantees, the other probes the unknowns. You can't just do one (again, IMO).

BTW - I get that this all takes time - and that the "juice might not be worth the squeeze" or whatever terrible analogy you want to use. I'm hoping that I can learn what you are looking for, to make it easier to review, starting at the bottom, and working my way up/across the stack.

> > I wasn't sure which was less work on your side

> I'm not sure yet either. 

How about split this (and future) PR(s) into : 
 - things that don't need any changes in existing avahi code (in this case 'strlst.c') - so you get the benefit of increased coverage, and knowledge where the gaps are - this would also be no fault injection.
 - add tests that expose issue(s) as secondary PRs, with the corresponding changes to core code (in this case 'strlst.c') - with either compiler warnings, or UBSan, MSan, TSan or ASan logs, or backtraces. Each issue as a separate PR (with fixes and tests).
 - Add Fault injection and compiler flags as driven by failures and PRs. These can be then debated individually based on their merits, and potential future re-use.

I guess part of it is how shared or standalone you envision directed tests. I can copy/paste failure injection in each `*-test.c` file (which seems dirty/tech-debt ish), or I can make a common header : `test-malloc.h` that they are also in. Or I can start with it in a 'strlst-test.c', and then move it to a common place when re-use requires (that was my plan, but I didn't want to create the future, without having a specific use case).

evverx:
> My intent with the directed tests isn’t to compete with fuzzing, but to complement it

I get that but I think it's generally better to complement it with symbolic/consolic execution to discover code paths automatically. The problem with that hard-coded allocator is that it's all done manually and future changes have to be done manually as well (which makes it harder to keep it all up to date and problematic in general since in most cases PRs come with no tests so all this manual testing infrastructure is actually less useful than it can appear). I get that it's great to define guarantees but test failures popping up because `malloc` is added/removed elsewhere and have to be manually counted again aren't helpful.

Either way I'll think about it.

rgetz:
obviously, I’m saying this as more of a tourist/long-time user than a core maintainer. Your perspective on the long-term health and maintainability of the project carries more weight than mine. I just wanted to explain the motivation and trade-offs behind the approach.

>I get that it's great to define guarantees but test failures popping up because malloc is added/removed elsewhere and have to be manually counted again aren't helpful.

On the point that this makes things harder to keep up to date or less useful because many PRs arrive without tests, I tend to see that a bit differently. If there were pre-existing deterministic tests covering core infrastructure, then a PR either passes them or it doesn’t. The best time to have that would have been 20 year ago. The second best time is today (completely cliché - but also (IMHO) true).

If you decide to agree, in a future state where there are more manual tests, when a PR comes in and causes a manual test to fail - that doesn’t mean the PR is wrong, sometimes that means the code is right, and the test needs to be updated, but that’s the normal maintenance burden of any codebase with a deterministic/manual test suite. In that case - the fail in the test that needs to be updated is also a signal to everyone: it forces the author of the PR to reconcile behavioral changes with the existing guarantees instead of them drifting silently. It is a forcing function for conversation. (Why did this PR break this test case, and is that change in behavior what we want).

If you decide to not to agree, that's totally fine - I have (based on your comments) been reading and playing more with [KLEE](https://klee-se.org/) - it's an interesting tool that I had not used before. Thanks for that pointer.

>Either way I'll think about it.

Cool - thanks - just let me know either way. I won't re-work/work on new things until your feedback.

evverx:
Just to clarify I'm not saying that the manual tests aren't useful. They are useful and I agree that it makes sense to cover the codebase with more tests like that. It's just that there are basic things that aren't covered and they should probably be covered first. Allocation failures are far from the top on that list. As mentioned in the other PR there was absolutely no testing infrastructure up until recently and the goal was to at least stop and prevent the infinite stream of crashes actually triggerable by incoming packets, D-Bus methods and so on.

> playing more with [KLEE](https://klee-se.org/) - it's an interesting tool that I had not used before

Usually it has to be patched one way or another to do various things but it can be useful by itself of course.

rgetz:
>Allocation failures are far from the top on that list.

understood - but I think that is what you are going to find at the bottom to _start_. I never said I would stop there... My background is always low level things - so that's where I naturally start - in the basement, and work my way up... Solid foundation, and even more clichés).

>prevent the infinite stream of crashes actually triggerable by incoming packets, D-Bus methods and so on.

That makes sense, and I agree the externally-triggerable crashes are absolutely the highest priority. 

I think we are saying similar things - here’s value in some low-level deterministic coverage happening in parallel, it’s just not equally urgent. 

The real constraint, of course, is time/bandwidth. If I can try to keep coverage improvements narrow and well-scoped so the review burden stays low, and don't go crazy (one or two a week, so I'm making progress, but it's not overwhelming to reviewers).

I also have been around open source long enough to understand the risks - Drop-in contributions - where someone adds code and doesn’t follow up - creates debt even if the change itself is well-intentioned. It’s just the reality of stewarding a long-lived project. 

My goal with these tests is to add something self-contained and low-risk, but I understand why you/other maintainers have to balance technical benefit against long-term ownership and reviewer capacity. Later or never is an Ok answer.

evverx:
> I think that is what you are going to find at the bottom to start

As mentioned at the bottom there are plans to integrate allocation failures into the fuzzing infrastructure. That thing actually intercepts malloc, finds bugs in glibc along the way so it's going to be much more productive in terms of covering those code paths. I can see that there is some interest in injecting allocation failures into the test suite as well so it's noted but it doesn't shift the priorities.

> Drop-in contributions - where someone adds code and doesn’t follow up - creates debt even if the change itself is well-intentioned. It’s just the reality of stewarding a long-lived project.

I think the current reality is that it's incredibly easy to generate plausibly looking code "fixing" issues by randomly arriving at solutions and offloading the actual thinking to reviewers. That's why I borrowed the following rule from LLVM

There are actually PRs that crossed that threshold but I keep them open because they highlight issues that should be addressed.

rgetz:
I do think fuzzing and directed tests end up serving very different purposes:
- Fuzzing is inherently statistical and discovery-oriented, it’s excellent at finding things you didn’t know to look for. 
- Directed tests, especially fast deterministic ones, are more about intent and guarantees: they prove that a specific behavior or invariant holds every single run, not just eventually by chance. 
IMHO - Since the code base handles allocation failures, and I know avahi is deployed in memory constrained embedded devices - I think that's an important aspect of that "intent and guarantee".

From a contributor perspective, I mostly just try to align with whatever testing philosophy the project prefers. If the expectation is “PRs should come with tests,” then deterministic tests are often the most practical way for an external contributor to provide something reviewable and repeatable. If the project leans more toward centralized fuzzing and symbolic tooling instead, that’s fine too, it just shifts where that validation effort lives.

I’ll close this (and others) if that's OK - since they don’t seem aligned with current project priorities.

Thanks for the thoughtful feedback, it helped me understand project direction better, and made me investigate fuzzing aspects for other projects that I will use.

evverx:
> I do think fuzzing and directed tests end up serving very different purposes

They do and I'm not saying that they serve the same purpose.

> Since the code base handles allocation failures

It's documented that currently it isn't the case https://github.com/avahi/avahi/blob/8bd8c4e64520944f9f38ddd426fe8957d51d9f21/docs/MALLOC#L1-L8

> If the project leans more toward centralized fuzzing and symbolic tooling instead, that’s fine too, it just shifts where that validation effort lives

I'm not sure how this conclusion is made. avahi uses the unit tests, the integration tests (combined with ASan/UBSan and Valgrind), static analyzers, fuzzing and so on because they serve different purposes. Fuzzing is just cheap and effective in terms of quickly discovering issues that are more important at this point and making sure they are gone for good.

I don't think the PRs should be closed but it's up to you of course.

rgetz:
>I don't think the PRs should be closed but it's up to you of course.

I'm open to refactoring things attempting to make it easier to review - but I don't want them to (a) be a burden (which I completely understand - these first will be more basic/rudimentary) and (b) don't want them to bit rot either. If you don't have time to review - I get it - there are only so many hours in the day - and lots of things to work on, and everything competes for priority.

evverx:
> If you don't have time to review

I have time for reviews. I don't have the time to validate the output of someone else's LLMs and things like that. I'll close all the PRs.

rgetz:
Thanks for the clear and simple answer. 

However - saying you don't have time, saying it's not interesting, saying you have different priorities, seems to be more honest assessment to me.

If the patches were LLVM generated - I would hope they would be more consistent, less duplicative, and in the proper style. (Which I agree - they are not today, due to my style issues, and not running them through a proper clang-format before sharing - that's on me).

If we ever end up in the same city/conference - still open to buy you a beverage of your choice.

evverx:
> saying you don't have time, saying it's not interesting, saying you have different priorities, seems to be more honest assessment to me

As I said I have time. For example in the PR where the fallthrough attribute was introduced I actually launched a bunch of containers with different versions of gcc and clang to see whether it actually worked (and it didn't work as far as I can remember) and I'm pretty sure that it took me more time than the time that was spent to write the patch (regardless of how it was written).

The LLVM rule doesn't necessarily cover LLMs only. It just says

rgetz:
> I actually launched a bunch of containers with different versions of gcc and clang to see whether it actually worked (and it didn't work as far as I can remember) 

it worked on:
```
Debian clang version 19.1.7 (3+b1)
```
and
```
gcc (Debian 14.2.0-19) 14.2.0
```

That was my tests - Based on the defaults for Debian 13.3 - I understand that as a widely deployed package - you need more than that - but I thought that CI would also catch those. When I looked, I didn't see anything - or I would have fixed it.

I added `-Wimplicit-fallthrough=5` (doesn’t recognize any comments as fallthrough comments, only attributes disable the warning) - so you should be seeing warnings - current CI does not.

all (ubuntu-24.04, gcc; netbsd, 10.1, gcc; omnios, r151056, gcc; alpine gcc; ubuntu-24.04-arm, gcc; ) in the CI logs have:
```
checking whether gcc accepts -Wimplicit-fallthrough=5... yes
```
and no warnings.

others (ubuntu-24.04, clang; freebsd, 14.3, clang; freebsd, 15.0, clang; alpine clang) have:
```
checking whether clang accepts -Wimplicit-fallthrough... yes
```
and same - no warnings.

The ones that had warnings on rawhide were the ones in `avahi-ui.c` that (I thought) you told me not to fix, since it was unmaintained.
```
avahi-ui.c:522:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
avahi-ui.c:652:13: warning: this statement may fall through [-Wimplicit-fallthrough=]
```

I'm not trying to argue with you (and I know I am - sorry) - I just want you to know I was actually trying to do the right thing - in my mind. I understand not all minds are aligned - and that's what drives the wonderfulness and diversity of the world. It's OK to have different opinions, priorities, and thoughts on things.

evverx:
> That was my tests

And that's why I launched a bunch of containers with different versions of clang and gcc because it did look like the gcc version check was untested and the clang check was missing and already at that point it took me more time to find the commit in llvm where that attribute was introduced than the time that was spent to write the patch.

>  I thought that CI would also catch those

Unfortunately the CI can't cover everything especially in cases like this.

> I added -Wimplicit-fallthrough=5

Right and I asked for that after spending a weird amount of time on saying that that warning isn't emitted by clang by default (even with `-Wextra`) and should be added to the CI.

> I just want you to know I was actually trying to do the right thing

I get that but I'm pretty sure that the attitude towards various things is different where I am and that's why I mostly focus on launching avahi-daemon and destroying it with actual packets first instead of looking at warnings emitted by static analyzers complaining about things that aren't even used anywhere.

rgetz:
>I get that but I'm pretty sure that the attitude towards various things is different where I am and that's why I mostly focus on launching avahi-daemon and destroying it with actual packets first instead of looking at warnings emitted by static analyzers complaining about things that aren't even used anywhere.

That still sounds like my assessment - admitting that a change/fix is not interesting and you have different priorities is OK.

My thought (at the time of the fallthrough), was similar to your from a few years ago, where in #425 you wrote:
>The idea is to turn on -Werror eventually to let compilers do their thing and catch various mistakes automatically. At this point it isn't possible to fully turn it on 

That makes total sense to me.

> Unfortunately the CI can't cover everything especially in cases like this.

Ack - that why I tried something (I thought) was low effort, (to fix the fall through warnings) to better understand what the bar was before looking at some of the other warnings generated on a modern systems, so you can eventually turn on `-Werror`...

turning on a variety of gcc/clang versions in CI - well, is what CI is for. It all depends on how far back your way back machine is - I can imagine for avahi (where people may want to re-build on older LTS distributions) - it goes pretty far back...

evverx:
> That still sounds like my assessment - admitting that a change/fix is not interesting and you have different priorities is OK.

It's not accurate. It was a low-effort PR where I spent more time on review than the time that was spent on writing the patch. If it was interesting/had a high priority somewhere I can't imagine what things with low priority look like.

Either way I hope that at this point it's clear that PRs are actually reviewed here and it shouldn't be surprising when I close the next low-effort PR.

rgetz:
>It was a low-effort PR where I spent more time on review than the time that was spent on writing the patch. 

IMHO - because you have standards (which is great) that aren't written down or communicated (which is bad).
 - it's clear - and you have stated many times, and it's in the contributing doc : "_A patch must save the maintainer more time than it costs to review._", that doesn't describe who owns verification effort, the contributor or the reviewer? People are happy to show results, but they need to know what you are looking for.
 - you want to see failures before fixes. (if something is going to fix a compiler warning, you want to see the warning; if something is going to fix a Sanitizer issue - you want to see the sanitizer trace, if it fixes a OOM, you want to see the crash). Show the concrete failure before submitting the fix, if possible include that in the commit message.
 - you have a list in your head (which has been discussed in other PRs) about which sections of avahai are unmaintained, and which are not. You don't want to give end users the wrong impression that something is maintained by accepting PRs into those sections.
 - You want to test on older versions of toolchains that aren't in CI. (I still don't know how far back you want to go).

Because these aren't written down - they are learned/communicated socially (potentially frustrating both reviewer and contributor - and I do hope that I haven't crossed the line into the frustrating side) rather than procedurally.

My interpretation of the discussion, is your “good” PR implicitly needs to:
 - Demonstrate a real, observable problem.
 - Prove the fix across more environments than CI covers.
 - Minimize reviewer investigation time (contributor owns validation).
 - Touch only actively maintained areas.
 - Align with runtime/user-visible impact over theoretical correctness.
 - Avoid style or tooling churn unless bundled with substantive benefit.

evverx:
>  "A patch must save the maintainer more time than it costs to review.", that doesn't describe who owns verification effort, the contributor or the reviewer?

I think it's obvious. If contributors offload everything onto reviewers their PR don't follow the rule.

I agree that it would be better if it was all spelled out explicitly but unfortunately it doesn't look like contributors sending low-quality patches read anything. I'm still thinking how to deal with that.

rgetz:
>I think it's obvious. If contributors offload everything onto reviewers their PR don't follow the rule.

Right, but when reviewers have additional expectations (for example, I still don’t know which distributions or how far back in gcc/clang versions you validate against), it can be difficult for contributors to know what level of verification is expected before submission. I agree 100% - that it should be contributors doing these things - but when contributors don't know what reviewers are looking for - it's naturally/unintentionally going to put friction on PRs.

For example - I didn’t initially realize that demonstrating the concrete failure prior to the fix was important for your review workflow (ie enabling `-Wimplicit-fallthrough=5` would be necessary to surface the warning you were looking for.) Had I known that up front, I would have included it in the original PR. 

Now that I understand the expectation, I can align any future/potential submissions accordingly.

My only point is that these kinds of project-specific review preferences aren’t always obvious to external / new contributors (since they are not in CI), and a bit of guidance can help reduce iteration for contributors and reduce reviewer load.

evverx:
> reviewers have additional expectations

I don't think I have additional expectations. It's not unreasonable to expect contributors to understand what their patches do (and why) and when I see commit messages confidently stating the wrong things like "No public API changes" (when a public header gets polluted) or "-Wimplicit-fallthrough is turned on by default in clang" (when it isn't actually turned on) it's safe to say that those patches aren't even verified/understood by their authors.

PR #853, opened 2026-01-30
Fix crash in `avahi-publish -a` by erroring out on name collisions

### LINKED ISSUE #478: `avahi-publish -a` crashes when conflicts occur

damiank89:
Fixes #478 by erroring out when a name collision occurs.

An alternative fix would be to enforce single-label names after parsing the arguments by checking `config->name` (then adding `.local` in `register_stuff()`). Multi-label names aren't forbidden by the RFC, so it wouldn't make much sense to disallow them.

@evverx Which approach do you prefer?

### DIFF
--- a/avahi-utils/avahi-publish.c
+++ b/avahi-utils/avahi-publish.c
@@ -88,7 +88,9 @@ static void entry_group_callback(AvahiEntryGroup *g, AvahiEntryGroupState state,
                 n = avahi_alternative_service_name(config->name);
             else {
                 assert(config->command == COMMAND_PUBLISH_ADDRESS);
-                n = avahi_alternative_host_name(config->name);
+                fprintf(stderr, _("Name collision detected, '%s' was already published by another network member.\n"), config->name);
+                avahi_simple_poll_quit(simple_poll);
+                break;
             }
 
             fprintf(stderr, _("Name collision, picking new name '%s'.\n"), n);


evverx:
Why do you think it doesn't make sense to pick another name when conflicts occur? The commit message doesn't explain the reasoning.

> An alternative fix would be to enforce single-label names

That would break existing scripts.

damiank89:
> Why do you think it doesn't make sense to pick another name when conflicts occur? The commit message doesn't explain the reasoning.
> 

Since `-a` means "publish address", you'd expect the program to publish the name as-is, so that you could rely on it being resolved exactly as asked to. No one would send queries to the new randomly generated name, so it's a bit redundant. On the other hand, when publishing a service it would make sense to pick a new name, since you could resolve it using the service type and not strictly using the name.

> > An alternative fix would be to enforce single-label names
> 
> That would break existing scripts.

Sounds right.

evverx:
> No one would send queries to the new randomly generated name, so it's a bit redundant. On the other hand, when publishing a service it would make sense to pick a new name, since you could resolve it using the service type and not strictly using the name

Given that among other things it's used to publish aliases and avahi-publish can accept hostnames when publishing services it can all be discovered in the end so I'm not sure why this option is ruled out.

damiank89:
> Given that among other things it's used to publish aliases and avahi-publish can accept hostnames when publishing services it can all be discovered in the end

I'm not sure I understand that scenario, can you give an example?

evverx:
I'll go ahead and close this. I added the following part to CONTRIBUTING recently:

and this PR doesn't pass this check. It's not the first PR like this by the way (https://github.com/avahi/avahi/pull/833) so please reconsider the approach to sending PRs here.

damiank89:
> I'll go ahead and close this. I added the following part to CONTRIBUTING recently:
> 
> > Our golden rule is that a contribution should be worth more to the project than the time it takes to review it.
> 
> and this PR doesn't pass this check. It's not the first PR like this by the way (#833) so please reconsider the approach to sending PRs here.

I think we just had a minor miscommunication since I genuinely found your comment [here](https://github.com/avahi/avahi/pull/853#issuecomment-3823885127) difficult to understand. I would really appreciate it if you clarified what you meant (which is why I asked for an example).

I try to only open PRs for existing issues. Some of them were even reported by you, which makes me believe they are meaningful to the project.

PR #863, opened 2026-02-07
retain const qualifier from pointer

heitbaum:
Since glibc-2.43:

For ISO C23, the functions bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr and wmemchr that return pointers into their input arrays now have definitions as macros that return a pointer to a const-qualified type when the in put argument is a pointer to a const-qualified type.

https://lists.gnu.org/archive/html/info-gnu/2026-01/msg00005.html

fixes the follow warnings:
- warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]

### DIFF
--- a/avahi-daemon/static-services.c
+++ b/avahi-daemon/static-services.c
@@ -83,7 +83,8 @@ struct StaticServiceGroup {
 static AVAHI_LLIST_HEAD(StaticServiceGroup, groups) = NULL;
 
 static char *replacestr(const char *pattern, const char *a, const char *b) {
-    char *r = NULL, *e, *n;
+    const char *e;
+    char *r = NULL, *n;
 
     while ((e = strstr(pattern, a))) {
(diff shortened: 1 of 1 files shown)


evverx:
Thank you for the PR. The CI confirms that the warnings are gone: https://openscanhub.fedoraproject.org/task/97674/log/fixed.html. I'll double-check and merge once the CI is green.

evverx:
Can you replace the tab in front of `char *r = NULL, *n` with four spaces? Looks like it was added automatically by the editor/IDE?

heitbaum:
> Can you replace the tab in front of `char *r = NULL, *n` with four spaces? Looks like it was added automatically by the editor/IDE?

👍 Fixed.

PR #90, opened 2016-11-16
Support GTK+ 4.0

murraycu:
These commits first avoid some deprecated API, then add a GTK4 version of avahi-ui, and then make some GTK4-specific changes.

I have not tested this at all and there doesn't seem to be any example/demo code for avahi-ui in the avahi repository itself.

Issue #679, opened 2025-02-05
avahi-discover-standalone: reachable assertion in `service_browser_callback` on receiving PTR resource records


evverx:
When unsolicited announcements with the following PTR resource records are received
```sh
        _services._dns-sd._udp.local: type PTR, class IN, _wat._udp.local                                                                                     
            Name: _services._dns-sd._udp.local                                                                                                                
            Type: PTR (domain name PoinTeR) (12)
            .000 0000 0000 0001 = Class: IN (0x0001)
            0... .... .... .... = Cache flush: False
            Time to live: 4500 (1 hour, 15 minutes)
            Data length: 17
            Domain Name: _wat._udp.local
        _wat._udp.local: type PTR, class IN, Q._woot._tcp.local                                                                                               
            Name: _wat._udp.local                                                                                                                             
            Type: PTR (domain name PoinTeR) (12)                                                                                                              
            .000 0000 0000 0001 = Class: IN (0x0001)                                                                                                          
            0... .... .... .... = Cache flush: False                                                                                                          
            Time to live: 4500 (1 hour, 15 minutes)                                                                                                           
            Data length: 20                                                                                                                                   
            Domain Name: Q._woot._tcp.local                        
```
`avahi-discover-standalone` crashes with:
```sh
**
ERROR:main.c:149:service_browser_callback: assertion failed: (s->service_type)
Bail out! ERROR:main.c:149:service_browser_callback: assertion failed: (s->service_type)

#0  0x00007ffff6a9e114 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff6a44f9e in raise () at /lib64/libc.so.6
#2  0x00007ffff6a2c942 in abort () at /lib64/libc.so.6
#3  0x00007ffff6d3614c in g_assertion_message[cold] () at /lib64/libglib-2.0.so.0
#4  0x00007ffff6da58c7 in g_assertion_message_expr () at /lib64/libglib-2.0.so.0
#5  0x00000000004e58b6 in service_browser_callback
    (b=0x506000191c00, interface=3, protocol=0, event=AVAHI_BROWSER_NEW, service_name=0x7ffff430c020 "Q", service_type=0x7ffff430c080 "_woot._tcp", domain_name=0x7ffff430c500 "local", flags=(AVAHI_LOOKUP_RESULT_CACHED | AVAHI_LOOKUP_RESULT_MULTICAST), userdata=0x0) at main.c:149
[...]

Issue #516, opened 2023-11-03
Document in 0.9 release notes, which parts/bindings are obsolete/going away


mbiebl:
Following the discussion in https://github.com/lathiat/avahi/pull/509#issuecomment-1792528401 I'd like to see documented in the release notes of 0.9, which parts/bindings are declared obsolete and scheduled to be removed by a subsequent release.
Candidates for the chopping list:
- Qt3/Qt4 bindings
- mono bindings
- gtk2.0 bindings 
- HOWL compatibility layer
- ...

mbiebl:
Do we know the maintainers of Avahi in OpenSUSE, Arch, $your_favourite_distro, we could ping regarding this matter?

evverx:
I'd EOL `avahi-compat-libdns_sd` and `avahi-compat-howl`. https://avahi.org/doxygen/html/ already says
> Please note that these compatibility layers are incomplete and generally a waste of resources

Plus they have come with banners saying
> *** WARNING: The application '%s' uses the "COMPAT_LAYER" compatiblity layer of Avahi. Please fix your application to use the native API of Avahi!
since 2005.

but bug reports like https://github.com/lathiat/avahi/issues/479 are still opened in 2023. I'm not sure they should be removed from the repository completely but I think they should be officially deprecated, issues and PRs should be closed and they should just be frozen.

Unfortunately I don't know who maintains avahi on distros other than Debian and Fedora/RH but I believe Debian and Fedora/RH are enough to get the ball rolling.

evverx:
@nkarstens got it. Based on issues I've seen here on GitHub it's indeed used and that's why I don't think it should be just dropped. The idea is to somewhat show that it isn't maintained (upstream at least). My understanding that you carry patches fixing various avahi bugs and deprecating the compat lib but keeping it shouldn't break a lot of things hopefully.

evverx:
I'll add the "needs-discussion" label just to make it clear that nothing has been set in stone yet. It's just my opinion after all. Then again looking at, for example, https://github.com/lathiat/avahi/issues/467 (unrelated to the compat libs) it seems to me that the "help wanted" label is overly optimistic. That part isn't tested at all: https://coveralls.io/builds/63727427 (and personally I'm not planning to change that), small changes like https://github.com/lathiat/avahi/pull/462 turn out to break other set-ups and so on. I think things like that should be just left alone to save contributors/bug reporters/maintainers time.

mbiebl:
[...]

PR #868, opened 2026-02-20
Port Avahi UI to GTK4

mikelolasagasti:
**Description:**
This Pull Request introduces GTK4 support for the Avahi UI components (`libavahi-ui`, `bssh`, and `avahi-discover-standalone`).

Instead of adding `#ifdef` blocks throughout the existing GTK2/GTK3 codebase (as was attempted in [PR90](https://github.com/avahi/avahi/pull/90)), this PR introduces a clean, dedicated GTK4 code path (`avahi-ui-gtk4.c`). This approach avoids legacy technical debt and allows the GTK4 build to use modern API paradigms natively.

**Key changes:**

1. **Build System:** Added `--enable-gtk4` to `configure.ac`. The build system now outputs `libavahi-ui-gtk4.la` and `avahi-ui-gtk4.pc` when enabled, leaving GTK2 and GTK3 builds completely unaffected.
2. **Library Modernization (`avahi-ui-gtk4.c`):** 
* Replaced the deprecated `GtkTreeView` with the modern `GtkListView` backed by a `GListStore`.
* Implemented lightweight `GObject` wrappers (`AuiServiceRow` and `AuiDomainRow`) to satisfy the `GListStore` requirement that all items must be objects.
* Replaced the deprecated synchronous `gtk_dialog_run()` with a fully asynchronous event flow using the `"response"` signal.
* Replaced deprecated layout properties (like `border_width` and `GtkVBox`) with modern `GtkBox` and widget `margin-*` properties. Injected a small CSS provider to fix standard spacing without needing an external stylesheet.

4. **Application Updates (`bssh` and `avahi-discover-standalone`):**
* Updated `bssh.c` to conditionally use a nested `GMainLoop` when built with GTK4 to maintain its synchronous command-line blocking behavior without using `gtk_dialog_run()`.
* Created a dedicated `avahi-discover-gtk4.ui` XML file that conforms to the GTK4 Builder schema (removed `<packing>` tags, fixed scroll policies).
* *Note:* `avahi-discover-standalone` currently retains the use of `GtkTreeView` in the C code to avoid a highly complex rewrite of the hierarchical `GtkTreeStore` data model. A comment is left for future refactoring to `GtkColumnView`.

**Testing:**
Built and tested locally on Fedora 43 using out-of-tree builds for both `--enable-gtk3` and `--enable-gtk4`. Both `bssh` and `avahi-discover-standalone` render correctly and resolve services.

**LLM-assisted:**
This change includes content generated with the help of Cursor (LLM-based assistant). The author reviewed, tested, and takes full responsibility for the final contribution.

### DIFF
--- a/avahi-ui/avahi-ui.c
+++ b/avahi-ui/avahi-ui.c
@@ -755,7 +755,7 @@ static gboolean start_callback(gpointer data) {
     gtk_tree_view_set_headers_visible(GTK_TREE_VIEW(d->priv->service_tree_view), FALSE);
     gtk_widget_show(d->priv->service_progress_bar);
 
-    if (d->priv->service_pulse_timeout <= 0)
+    if (d->priv->service_pulse_timeout == 0)
         d->priv->service_pulse_timeout = g_timeout_add(100, service_pulse_callback, d);
 
     for (i = 0; d->priv->browse_service_types[i]; i++)
@@ -1207,7 +1207,7 @@ static void aui_service_dialog_init(AuiServiceDialog *d) {
 static void restart_browsing(AuiServiceDialog *d) {
     g_return_if_fail(AUI_IS_SERVICE_DIALOG(d));
 
-    if (d->priv->start_idle <= 0)
+    if (d->priv->start_idle == 0)
         d->priv->start_idle = g_idle_add(start_callback, d);
 }
 


mikelolasagasti:
Packit fails as it uses upstream dist-git's specfile and it doesn't include GTK4 support. Not sure how to cleanly workaround this chicken-and-egg problem.

evverx:
Thank you for the PR. That part of avahi isn't actually maintained and for example https://github.com/avahi/avahi/pull/90 links to https://github.com/avahi/avahi/issues/516 where it was discussed that it should be fine to drop it (but Fedora/RedHat has never shown up there for some reason). Is there a reason it should be kept in the avahi repository instead of keeping it elsewhere?

Either way I'll leave it to @pemensik. I don't even look at those things.

mikelolasagasti:
> That part of avahi isn't actually maintained and for example https://github.com/avahi/avahi/pull/90 links to https://github.com/avahi/avahi/issues/516 where it was discussed that it should be fine to drop it (but Fedora/RedHat has never shown up there for some reason). Is there a reason it should be kept in the avahi repository instead of keeping it elsewhere?


I understand from that issue that there are two separate goals being discussed:

1. Dropping old technologies (Qt3, GTK2, etc.)
2. Moving the UI code entirely to a separate repository.

To answer your question: since the UI code hasn't been moved to a new repository yet, there is currently no other place to open this PR.

Also, this PR actually helps with the first goal. By providing a clean GTK4 replacement, it makes it much easier for distributions to finally drop the old GTK2 and GTK3 bindings.

Happy to wait for @pemensik to review!

evverx:
That issue was initially about dropping GTK2 and thing like that but in https://github.com/avahi/avahi/issues/516#issuecomment-1793567370 it was concluded that it should be fine to remove the avahi-ui part completely and I don't think there was any interest in moving/maintaining it anywhere else.

> since the UI code hasn't been moved to a new repository yet, there is currently no other place to open this PR

I think in terms of moving to modern technologies it would be better to keep those things elsewhere so as not being tied to the choices made a long time ago like the build system and so on.

> By providing a clean GTK4 replacement, it makes it much easier for distributions to finally drop the old GTK2 and GTK3 bindings

My understanding is that distros just can just drop that stuff completely because it isn't used anywhere.

As for the clean replacement as far as I can see OpenScanHub found 85 new findings in https://openscanhub.fedoraproject.org/task/100200/log/added.html and that's a lot of deprecation warnings.

.github/workflows/build.sh
```diff
@@ -131,15 +131,18 @@ case "$1" in
         pkg install -y gettext-runtime gettext-tools gmake intltool \
             gobject-introspection pkgconf expat libdaemon dbus-glib dbus gdbm \
             libevent glib automake libtool libinotify qt5-core qt5-buildtools \
-            gtk3 py311-pygobject py311-dbus py311-gdbm mono git socat \
+            gtk3 gtk40 py311-pygobject py311-dbus py311-gdbm mono git socat \
```
arrowd: Package's name is `gtk4`.

mikelolasagasti: I checked the port name in https://cgit.freebsd.org/ports/tree/x11-toolkits/gtk40 and that's why used `gtk40`. Will change.

arrowd: Port is a build recipe, like `.spec` files in RPM. In both packaging systems it is possible to get a package that is named differently than a build recipe, or even get multiple packages from a single recipe.

evverx:
@arrowd does FreeBSD ship that stuff? Is there any interest in fixing crashes like https://github.com/avahi/avahi/issues/679 triggered by incoming packets?

arrowd:
> does FreeBSD ship that stuff?

Yes, we do ship `avahi-discover-standalone` executable. No idea how widely it is used, though.

> Is there any interest in fixing crashes like https://github.com/avahi/avahi/issues/679 triggered by incoming packets?

I wouldn't mind looking at this, but I'm too swamped with other work, unfortunately.

evverx:
> Yes, we do ship avahi-discover-standalone executable. No idea how widely it is used, though.

My guess would be that it was flipped a long time ago and since it's just a pile of technical debt now it can be safely dropped.

evverx:
I saw https://github.com/avahi/avahi/pull/874. I'm not a fan of one-off fixes because it doesn't make it much better and it isn't maintained anyway but if there is interest it would make sense to take a closer look at all that stuff and for example escape strings with TXT resource records before passing them to `gtk_label_set_markup` and things like that.

The CI here still fails FWIW. `make distcheck` fails and things like that.

mikelolasagasti:
> As for the clean replacement as far as I can see OpenScanHub found 85 new findings in https://openscanhub.fedoraproject.org/task/100200/log/added.html and that's a lot of deprecation warnings.

With latest changes many of the issues should be gone.

Deprecation warnings are expected as described in the PR "`avahi-discover-standalone` currently retains the use of `GtkTreeView`". Fixing those would be a follow-up patch, but doesn't make sense the effort if current PR might not be merged.

> I saw https://github.com/avahi/avahi/pull/874. I'm not a fan of one-off fixes because it doesn't make it much better and it isn't maintained anyway but if there is interest it would make sense to take a closer look at all that stuff and for example escape strings with TXT resource records before passing them to gtk_label_set_markup and things like that.

Having 100% perfect and complete PRs on first iteration is not easy. You pointed to an issue and I created a PR to fix it. You have more experience with the codebase and the issues than I do, I'm just trying to help and that's an example. More can be done? sure.

> The CI here still fails FWIW. make distcheck fails and things like that.

`distcheck` should be OK now.

evverx:
> Fixing those would be a follow-up patch, but doesn't make sense the effort if current PR might not be merged.

Agreed. But if it's decided that it makes sense to keep it after all I don't think it makes sense to start with introducing even more technical debt and fix it later. It isn't urgent so it doesn't look like it should be done in a hurry.

> Having 100% perfect and complete PRs on first iteration is not easy.

Sorry I didn't mean to say that it should be fixed here (or anywhere else). My point was that that part of avahi has been abandoned and it's unclear whether it makes sense to keep it if it just keeps piling technical debt.

evverx:
CIFuzz and ClusterFuzzLite fail because it isn't turned off in https://github.com/avahi/avahi/blob/21f8772e6fd80a0b5c223633151b199a060c99ac/fuzz/oss-fuzz.sh#L75-L81

The coverage job fails because of https://status.coveralls.io/incidents/pdbt7vdzlvpj and can be ignored for now.

Alpine, NetBSD and OmniOS appear to miss the dependencies.

PR #875, opened 2026-02-25
ci: validate XML in the linter action

evverx:
to make sure XML files are more or less fine.


evverx:
The coverage job failed because of https://status.coveralls.io/incidents/pdbt7vdzlvpj
> Outage will be extended for at least several more hours

PR #879, opened 2026-03-04
Correct obsolete FSF address in all license notices

pemensik:
```
    Fixed with following sed script:
    s|License along with avahi; if not, see <https://www.gnu.org/licenses/>.|License along with avahi; if not, see <https://www.gnu.org/licenses/>.|
    /^[# ] Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307/ d
    /^[# ] USA./ d
    s|along with avahi; if not, see <https://www.gnu.org/licenses/>.|along with avahi; if not, see <https://www.gnu.org/licenses/>.|
    /^[# ] Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA./ d
    /^[# ] Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA/ d
    /^[# ] 02111-1307 USA./ d
```

### DIFF
--- a/.github/workflows/coverity.sh
+++ b/.github/workflows/coverity.sh
@@ -13,9 +13,7 @@
 # License for more details.
 #
 # You should have received a copy of the GNU Lesser General Public
-# License along with avahi; if not, write to the Free Software
-# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
-# USA.
+# License along with avahi; if not, see <https://www.gnu.org/licenses/>.
 
 set -eux
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,9 +1,8 @@
-
                   GNU LESSER GENERAL PUBLIC LICENSE
                        Version 2.1, February 1999
 
  Copyright (C) 1991, 1999 Free Software Foundation, Inc.
-	51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
@@ -23,8 +22,7 @@ specially designated software packages--typically libraries--of the
 Free Software Foundation and other authors who decide to use it.  You
 can use it too, but we suggest you first think carefully about whether
 this license or the ordinary General Public License is the better
-strategy to use in any particular case, based on the explanations
-below.
+strategy to use in any particular case, based on the explanations below.
 
   When we speak of free software, we are referring to freedom of use,
 not price.  Our General Public Licenses are designed to make sure that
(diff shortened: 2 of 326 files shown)


evverx:
Please add the rationale behind this change to the commit message. For example https://github.com/systemd/systemd/commit/4a8c41f3eb83aa1524501d9dc1c56c9026579f18 came with
> rpmlint complains about the old address, even though it totally
doesn't matter and nobody cares.

evverx:
Also I'm not sure I understand how that sed script turned
```sh
  <signature of Ty Coon>, 1 April 1990
  Ty Coon, President of Vice
```
to
```sh
  <signature of Moe Ghoul>, 1 April 1990
  Moe Ghoul, President of Vice
```
and where a bunch of other changes unrelated to the address came from.

.github/workflows/coverity.sh
```diff
@@ -13,9 +13,7 @@
 # License for more details.
 #
 # You should have received a copy of the GNU Lesser General Public
-# License along with avahi; if not, write to the Free Software
-# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
-# USA.
+# License along with avahi; if not, see <https://www.gnu.org/licenses/>.
```
adriaandegroot, line 16: If you're going to touch all the files like this anyway, you might consider adding an `SPDX-License-Identifier:` line to make it machine-readable as well. (Whether avahi wants SPDX, I don't know, this is a drive-by review).

adriaandegroot:
> Also I'm not sure I understand how that sed script turned
> 
> ```shell
>   <signature of Ty Coon>, 1 April 1990
>   Ty Coon, President of Vice
> ```
> 
> to
> 
> ```shell
>   <signature of Moe Ghoul>, 1 April 1990
>   Moe Ghoul, President of Vice
> ```
> 
> and where a bunch of other changes unrelated to the address came from.

The sample text on the FSF website at https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt is different from whenever LICENSE was first imported. Nowadays the sample text upstream actually says Moe Ghoul. Note that the SPDX license text sample for LGPL version 2.1 is the "old" one, which matches what is in the avahi repository.

evverx:
> The sample text on the FSF website at https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt is different from whenever LICENSE was first imported

The commit message should mention that. Currently it just says that the sed script was run and that script couldn't change "Ty Coon" to "Moe Ghoul" obviously.

> If you're going to touch all the files like this anyway, you might consider adding an SPDX-License-Identifier: line to make it machine-readable as well

I think with `SPDX-License-Identifier` the text can be removed as well.

PR #881, opened 2026-03-09
Move documents from .github directory to better locations

pemensik:


evverx:
Can you expand on why you think it's better than `.github`?

pemensik:
@evverx because ``.github`` is hidden, therefore not suitable for instructions directed to people? Github is not the only interface people should work with the code.

evverx:
> because .github is hidden, therefore not suitable for instructions directed to people?

It's the standard place and a lot of projects keep those files there so I don't think people can't find it. The only place I've heard of where it was an issue involved some sloppy tooling using `ls`. As mentioned elsewhere I don't care much where it's kept as long as it's visible on the GitHub and according to https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#about-contributing-guidelines it's fine to keep CONTRIBUTING.md in the root directory.

Either way please add the rationale to the commit message.

Issue #503, opened 2023-10-27
New release to fix all CVEs since 2021?


Neustradamus:
Dear Avahi team, @lathiat, @evverx, @pemensik,

Current unsecure stable version is 0.8.0 (2020-02-18), 3 years, 8 months, 9 days.

It is possible to create the 0.9.0 release build to fix all CVEs?
- https://www.cvedetails.com/vulnerability-list/vendor_id-4481/Avahi.html
- https://cve.report/software/avahi/avahi
- ...

CVE-2023-38469: https://github.com/lathiat/avahi/issues/455
CVE-2023-38470: https://github.com/lathiat/avahi/issues/454
CVE-2023-38471: https://github.com/lathiat/avahi/issues/453
CVE-2023-38472: https://github.com/lathiat/avahi/issues/452
CVE-2023-38473: https://github.com/lathiat/avahi/issues/451
CVE-2023-1981: https://github.com/lathiat/avahi/issues/375

CVE-2021-36217
CVE-2021-26720
CVE-2021-3502: https://github.com/lathiat/avahi/issues/338
CVE-2021-3468

The original tickets have been closed without the new release build:
- https://github.com/lathiat/avahi/issues/325
- https://github.com/lathiat/avahi/issues/430

Thanks in advance.

cc: @ilkery.

pemensik:
I have been busy working on *dnsconfd* project, but I think it is time for a release already. Not much new features, but a lot of important bugs already fixed. Without known regressions AFAIK. So yes, it would be nice prepare a release finally. There is still few things that needs fixing, but hopefully next release will come in summer or so. We do not have a perfect release, but it is time for a release. Ideally before FOSDEM starts.

pemensik:
There is still open issue #501, which seems should be fixed if possible before we release.

pemensik:
I thought I would do a release today, but I have to finish preparations for my fosdem talk. I have pushed tag ``v0.9-rc1`` to allow more testing and downloading of prepared archive. There might be just few minor changes on top in final v0.9. I would like to make a release in 14 days, but have no time now to prepare decent release notes, which I want in final release. I guess for development branches of distributions rc1 would be enough.

xypiie:
[...]

PR #895, opened 2026-04-15
CVE-2025-59529: simple protocol connection limits with review fixes and tests

abhinavagarwal07:
Picking up where #808 left off. First commit is @pemensik's connection limit fix (squashed from #808), second addresses @evverx's review and adds smoke tests, third fixes CI issues.

Commit 1 adds per-UID and global client limits to the simple protocol, derived from `RLIMIT_NOFILE`. Commit 2 fixes the review feedback from #808: unknown UID returns `(uid_t)-1` not `0`, validates `getsockopt` length, fixes format string UB on no-credential platforms, fixes the off-by-one in the per-UID check, and cleans up style. It also adds smoke tests that set `rlimit-nofile=100` (so max_clients=75, per_uid=37) and flood the socket to verify refusals are logged.

Commit 3 fixes CI: under Valgrind, `setrlimit()` is intercepted, so the test sets `ulimit -n 100` before launching Valgrind. On systemd, a transient `LimitNOFILE=100` drop-in is needed since the unit's limit overrides the daemon config. Also restarts `avahi-dnsconfd` after each daemon kill in the test (it dies when the daemon does), and fixes an off-by-one in the per-UID fast-path guard and a `max_clients=0` edge case.

**Limitations**

- `cr_pid` may need `AC_CHECK_MEMBERS` guard for FreeBSD < 13.2
- Per-UID enforcement requires credential support; without it, only the global limit applies
- Per-UID smoke test needs `runuser` (Linux only)

### DIFF
--- a/avahi-daemon/simple-protocol.c
+++ b/avahi-daemon/simple-protocol.c
@@ -35,6 +35,9 @@
 #ifdef HAVE_LIBSYSTEMD
 #include <systemd/sd-daemon.h>
 #endif
+#ifdef HAVE_STRUCT_XUCRED
+#include <sys/ucred.h>
+#endif
 
 #include <avahi-common/domain.h>
 #include <avahi-common/llist.h>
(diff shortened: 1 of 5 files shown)


abhinavagarwal07:
Hope this helps unblock the v0.9 release — the lack of tests for the connection limit fix was one of the open items in #503 and #808.

avahi-daemon/simple-protocol.c
```diff
@@ -438,6 +546,56 @@ static void client_work(AvahiWatch *watch, AVAHI_GCC_UNUSED int fd, AvahiWatchEv
...
+     * treated as non-root so they cannot bypass the per-UID limit.
+     *
+     * On platforms without credential support, all connections share one
+     * "unknown UID" bucket, which effectively limits them to max_uid_clients
+     * instead of max_clients. This is an acceptable tradeoff: returning 0
+     * (root) would silently skip per-UID enforcement entirely. */
+    if (uid != 0 && n_clients > s->max_uid_clients) {
+        /* There are enough clients to reach the limit for one UID.
+         * Check this UID does not have too many connections already. */
+        Client *c;
```
abhinavagarwal07, line 580: Worth noting: on platforms without credential support (NetBSD, OmniOS), all connections share one "unknown UID" bucket since they all get uid=(uid_t)-1. This means the effective limit is max_uid_clients instead of max_clients — roughly halved. I went with this because the alternative (returning uid=0) silently skipped per-UID enforcement entirely, which felt worse. Open to changing this if you'd prefer unknown UIDs to be exempt from per-UID checks.

abhinavagarwal07:
@evverx fixed CI failures

evverx:
I didn't get past the part where the test is skipped under Valgrind because it's already tested in non-Valgrind jobs. Is it LLM-generated?

abhinavagarwal07:
I did use LLM assistance for this PR.

evverx:
LLM-generated contributions are covered in https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md#llms. It was added to make sure that LLM-generated code is reviewed before PRs are opened. Our golden rule is that a contribution should be worth more to the project than the time it takes to review it.

Before proceeding with this can you expand on why it was decided to launch a bunch of socat processes and why  it says that "systemd's LimitNOFILE overrides the daemon config"?

PR #841, opened 2026-01-21
 Add support for conf.d style dirs (config fragments)

manuelafm:
This ticket addresses avahi#439.

It implements the feature to read config from conf.d style directories (also known as "config fragments"), commonly supported by other tools, as an extension of the main config file.

This is useful and convenient in several situations, for example:
* distributions can set different values from the defaults shipped upstream without having to modify the main .conf files
  * this is specially important in some distributions, where otherwise they would have to have a local fork / vendored version of avahi just for the sake of setting some different default configuration.
* different packages or config-management tools, or simply users manually, can change config just by dropping, removing or modifying files in that directory; it's sometimes more convenient or easy than modfying the main config file.

The main config file, will be processed first, and then files in the conf.d dir, if they exist.

The directory to use is derived automatically, e.g. the same path as `avahi-daemon.conf`, suffixed with `.d`.  So, usually it will be `/etc/avahi/avahi-daemon.conf.d`.

Only files starting with `00-` to `99-` and ending in `.conf` within that directory are considered, and follow a simple naming format (pattern) to make it easy and predictable to sort them.  The candidate files are sorted with strcmp().

The configuration in the files being processed later will override previous config, in particular the main config file, but also the one from previous files processed, if the Group and Key coincide.

PR #898, opened 2026-04-23
ini-file-parser: raise line buffer from 1024 to 2048 bytes

leonard-eero:
The ini parser reads each line of avahi-daemon.conf into a fixed 1024-byte stack buffer via fgets(). For list-valued keys such as allow-interfaces= this limits the number of interface names that can be listed on a single config line. With a maximum interface name length of 15 characters plus a comma separator (16 bytes per entry), a 1024-byte buffer supports only ~64 interfaces.

Raise the buffer to 2048 bytes to support up to 128 interfaces in a single list-valued key.

The buffer lives on the stack of avahi_ini_file_load(), which only runs once at daemon startup from a shallow call site, so the extra bytes of stack usage are inconsequential.

No functional change for configurations that already fit in 1024 bytes.

### DIFF
--- a/avahi-daemon/ini-file-parser.c
+++ b/avahi-daemon/ini-file-parser.c
@@ -50,7 +50,7 @@ AvahiIniFile* avahi_ini_file_load(const char *fname) {
 
     line = 0;
     while (!feof(fo)) {
-        char ln[1024], *s, *e;
+        char ln[2048], *s, *e;
         AvahiIniFilePair *pair;
 
         if (!(fgets(ln, sizeof(ln), fo)))


evverx:
Thank you for the PR. Is there a reason it was decided to bump it instead of switching to allocating it properly as was discussed in another PR where it was bumped?

(FWIW in https://github.com/avahi/avahi/pull/841 avahi should start combining interfaces listed in various `allow-interfaces=` settings)

evverx:
Either way I don't think it makes sense to keep bumping it to cover every use case for eternity. It's a duplicate of https://github.com/avahi/avahi/issues/315 where it was already bumped once without actually addressing it so I'll close it.

https://github.com/avahi/avahi/pull/841 should help eventually by allowing splitting lists (but it doesn't fully address https://github.com/avahi/avahi/issues/315 either).

PR #904, opened 2026-04-29
avahi-daemon: fix -Wmaybe-uninitialized warning in static-services.c

heitbaum:
Initialize value_buf to NULL to silence GCC's -Wmaybe-uninitialized warning. The warning is a false positive since all valid txt_type branches assign value_buf before use, but GCC cannot see through assert(0) in the default case when NDEBUG is set.

(Split of of #900)

```c

In function 'xml_end',
    inlined from 'xml_end' at static-services.c:568:21:
static-services.c:659:17: warning: 'value_buf' may be used uninitialized [-Wmaybe-uninitialized]
  659 |                 memcpy(u->service->txt_records->text + key_len + 1, value_buf, value_buf_len);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
static-services.c: In function 'xml_end':
static-services.c:625:26: note: 'value_buf' was declared here
  625 |                 uint8_t *value_buf;
      |                          ^~~~~~~~~

```

### DIFF
--- a/avahi-daemon/static-services.c
+++ b/avahi-daemon/static-services.c
@@ -622,7 +622,7 @@ static void XMLCALL xml_end(void *data, AVAHI_GCC_UNUSED const char *el) {
             assert(u->service);
             if (u->txt_key != NULL) {
                 size_t key_len = strlen(u->txt_key);
-                uint8_t *value_buf;
+                uint8_t *value_buf = NULL;
                 uint8_t *free_value_buf = NULL;
                 size_t value_buf_len = 0;
 


evverx:
> GCC cannot see through assert(0) in the default case when NDEBUG is set

The idea behind that assertion is to indicate that that part can never be reached so it should either abort or use something like `__builtin_unreachable()`. I'd need to think about it.

evverx:
Can you replace `assert(0)` with `abort()` only and drop the NULL assignment. It should be consistent with the other places in the codebase where the default case isn't supposed to be reachable like https://github.com/avahi/avahi/blob/9b8e798c70f374253326910074e8feb03c57da54/avahi-core/resolve-service.c#L263-L264

Unnecessary initializations can prevent various dynamic analysis tools like Valgrind/ASan from discovering issues. For example https://github.com/avahi/avahi/pull/906 wouldn't have worked out if things were initialized there explicitly.

PR #912, opened 2026-05-08
core: fill out msg_controllen for IPV6_PKTINFO control messages correctly

evverx:
to prevent avahi_send_dns_packet_ipv6 from failing on systems where the
length is strictly validated. For example without this patch on OpenBSD
all the attempts to send packets over IPv6 get rejected with
```
CALL  sendmsg(13,0x7d91b98158c8,0)
STRU  struct msghdr { name=0x7d91b98158f8, namelen=28, iov=0x7d91b98158b8, iovlen=1, control=0x7d91b9815880, controllen=36, flags=0 }
STRU  struct iovec { base=0x8f99574c030, len=228 }
STRU  struct sockaddr { AF_INET6, [ff02::fb]:5353 }
STRU  struct cmsghdr { len=36, level=41<ipv6>, type=46 }
RET   sendmsg -1 errno 22 Invalid argument
```
It has been addressed downstream with a similar patch:
https://github.com/openbsd/ports/blob/92b23c283dfd8173e6035db42b5ec82acad64ae8/net/avahi/patches/patch-avahi-core_socket_c

OpenBSD is rolled out to make sure it's buildable and releasable there.

### DIFF
--- a/avahi-core/socket.c
+++ b/avahi-core/socket.c
@@ -584,7 +584,10 @@ int avahi_send_dns_packet_ipv6(
     struct msghdr msg;
     struct iovec io;
     struct cmsghdr *cmsg;
-    size_t cmsg_data[(CMSG_SPACE(sizeof(struct in6_pktinfo))/sizeof(size_t)) + 1];
+    union {
+        uint8_t cmsg_data[CMSG_SPACE(sizeof(struct in6_pktinfo))];
+        struct cmsghdr hdr;
+    } u;
 
     assert(fd >= 0);
     assert(p);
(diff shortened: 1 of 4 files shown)


avahi-core/socket.c
```diff
@@ -584,7 +584,10 @@ int avahi_send_dns_packet_ipv6(
     struct msghdr msg;
     struct iovec io;
     struct cmsghdr *cmsg;
-    size_t cmsg_data[(CMSG_SPACE(sizeof(struct in6_pktinfo))/sizeof(size_t)) + 1];
+    union {
+        uint8_t cmsg_data[CMSG_SPACE(sizeof(struct in6_pktinfo))];
+        struct cmsghdr hdr;
+    } u;
```
evverx, line 590: OpenScanHub complains https://openscanhub.fedoraproject.org/task/114977/log/added.html
```
Error: CPPCHECK_WARNING ([CWE-398](https://cwe.mitre.org/data/definitions/398.html)): [[#def1]](https://openscanhub.fedoraproject.org/task/114977/log/added.html#def1)
avahi-0.9.rc4/avahi-core/socket.c:590: style[variableScope]: The scope of the variable 'u' can be reduced.
#  588|           uint8_t cmsg_data[CMSG_SPACE(sizeof(struct in6_pktinfo))];
#  589|           struct cmsghdr hdr;
#  590|->     } u;
```
but it's a false positive. The scope can't be reduced because `msg` points to it so if `u` was moved to the inner scope it would lead to
```
==35332==ERROR: AddressSanitizer: stack-use-after-scope on address 0xff6d208172f0 at pc 0xff6d23cce194 bp 0xffffe1975030 sp 0xffffe1974810
READ of size 8 at 0xff6d208172f0 thread T0
     #0 0xff6d23cce190 in read_msghdr_control ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3177
     #1 0xff6d23cce190 in read_msghdr ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3215
     #2 0xff6d23ce35e4 in sendmsg ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3230
     #3 0xff6d2391cb94 in sendmsg_loop /home/runner/work/avahi/avahi/avahi-core/socket.c:450
     #4 0xff6d2391e6d4 in avahi_send_dns_packet_ipv6 /home/runner/work/avahi/avahi/avahi-core/socket.c:637
```

PR #914, opened 2026-05-21
core: free copy's HINFO 'os' field on rdata copy failure

dxbjavid:
### Summary

In `avahi_record_copy()` (avahi-core/rr.c) the HINFO branch allocates
two strings, one for the copy's `data.hinfo.os` and one for
`data.hinfo.cpu`. When the second `avahi_strdup()` call fails, the
existing error path frees the wrong pointer: it calls
`avahi_free(r->data.hinfo.os)` on the *source* record's field instead
of `avahi_free(copy->data.hinfo.os)`.

avahi-core/rr.c:452-455 (before patch):

```c
if (!(copy->data.hinfo.cpu = avahi_strdup(r->data.hinfo.cpu))) {
    avahi_free(r->data.hinfo.os);
    goto fail;
}
```

`r` is the source record, still live and still owned by the caller
(e.g. `make_goodbye_record()` in avahi-core/announce.c:366). Freeing
its `hinfo.os` field has three consequences:

1. The source record is left with a dangling, non-NULL
   `data.hinfo.os` pointer.
2. The copy's `data.hinfo.os`, which was just allocated on the line
   above, is leaked (`fail:` only frees `copy->key` and `copy` itself,
   not its rdata fields).
3. When the source record is later destroyed via
   `avahi_record_unref()` (avahi-core/rr.c:147), it calls
   `avahi_free(r->data.hinfo.os)` on the already-freed pointer,
   producing a double-free.

### How to observe

A standalone reproducer that mirrors the snippet against a tracking
allocator (NULL-on-Nth-allocation, abort-on-wrong-free) deterministically
triggers the fault. With the current code:

```
BUGGY:  calling unref_src() - expect FAULT...
FAULT: xfree(0x100749c30) on a pointer not in the live set
       (double-free or wrong-pointer free)
BUGGY:  after failed copy, src->hinfo.os = 0x100749c30 (dangling)
BUGGY:  after failed copy, copy.hinfo.os = 0x100749c40 (leaked)
```

After the patch the same reproducer completes cleanly with no live
allocations.

In practice the second `avahi_strdup()` returns NULL whenever a
custom allocator installed via `avahi_set_allocator()` reports OOM;
the default allocator aborts on failure so the path is not directly
reachable there, but the freed-wrong-pointer logic is incorrect
regardless of the allocator in use.

### Fix

Free `copy->data.hinfo.os` (the field that was actually allocated
into) rather than `r->data.hinfo.os`. The change is local to the
callee and does not affect any successful path.

### Build / test

`avahi-core/rr.c` compiles cleanly with the patch applied.

### DIFF
--- a/avahi-core/rr.c
+++ b/avahi-core/rr.c
@@ -450,7 +450,7 @@ AvahiRecord *avahi_record_copy(AvahiRecord *r) {
                 goto fail;
 
             if (!(copy->data.hinfo.cpu = avahi_strdup(r->data.hinfo.cpu))) {
-                avahi_free(r->data.hinfo.os);
+                avahi_free(copy->data.hinfo.os);
                 goto fail;
             }
             break;


evverx:
Thank you for the PR. The patch looks good to me and it fixes the double-free (which can be triggered with a modified avahi record fuzz target with a fuzzing engine triggering allocation failures)
```sh
==343346==ERROR: AddressSanitizer: attempting double-free on 0x7b998dbf28b0 in thread T0:                                                                     
    #0 0x0000005462ba in free (/home/vagrant/avahi/out/fuzz-consume-record+0x5462ba) (BuildId: 2344765d33bd38376effd66a60243a4d44c477ad)                      
    #1 0x000000599de9 in avahi_record_unref .../avahi/avahi-core/rr.c
...
0x7b998dbf28b0 is located 0 bytes inside of 12-byte region [0x7b998dbf28b0,0x7b998dbf28bc)
freed by thread T0 here:
    #0 0x0000005462ba in free (/home/vagrant/avahi/out/fuzz-consume-record+0x5462ba) (BuildId: 2344765d33bd38376effd66a60243a4d44c477ad)
    #1 0x00000059fd28 in avahi_record_copy .../avahi/avahi-core/rr.c:453:17
...
previously allocated by thread T0 here: 
    #0 0x000000546558 in malloc (/home/vagrant/avahi/out/fuzz-consume-record+0x546558) (BuildId: 2344765d33bd38376effd66a60243a4d44c477ad)
    #1 0x0000005aeacf in avahi_strdup /home/vagrant/avahi/avahi-common/malloc.c:168:15
    #2 0x000000594ef4 in parse_rdata /home/vagrant/avahi/avahi-core/dns.c:569:38
    #3 0x000000594477 in avahi_dns_packet_consume_record /home/vagrant/avahi/avahi-core/dns.c:654:9
```

That being said avahi is really bad at handling allocation failures and there are quite a few places that should be adjusted to make it robust so it would be interesting to find out how this specific issue was discovered. Could you expand on how it was discovered and add it to the commit message? It would probably make sense to mention how unlikely it is for this to be actually triggered.

dxbjavid:
I found the issue while fuzzing the fuzz consume record target with allocation failure injection enabled. The crash happens when an allocation fails during the HINFO copy path, which can leave the copied record referencing memory still owned by the original record. That later results in a double-free during cleanup.

I’ve updated the commit message with the discovery details and mentioned that this is unlikely to occur in normal runtime conditions since it depends on a specific allocation failure path.

evverx:
The commit message looks good. I'll merge it once the CI is green but generally in its current form it's documented that avahi doesn't handle allocation failures very well and instead of fixing things one-by-one with no tests the plan is to inject allocation failures into the fuzzing infrastructure eventually first to let the CI make sure things actually get fixed and no longer pop up. I have those fuzz targets downstream and I'm aware of the issues they can find and one-off fixes aren't the way to go. Just out of curiosity is that a known fuzzing engine stable enough to be used in the CI or is it some research project?

dxbjavid:
I was mainly experimenting with allocation failure scenarios around the existing fuzz target to see how the error paths behaved. It was not based on a dedicated fuzzing engine or CI ready framework,, just a small local setup to help trigger allocation failures during fuzzing.

PR #915, opened 2026-05-22
compat-libdns_sd: bound trailing-dot strcat in service_resolver_callback

dxbjavid:
avahi_service_name_join() snprintfs into full_name (a stack buffer of
AVAHI_DOMAIN_NAME_MAX bytes) and silently truncates when the joined
name would exceed that limit. When a resolved service's type+domain
push the joined name to the limit, the following strcat(full_name,
".") writes one byte past the buffer end. Reproduced locally against
the real avahi_service_name_join; append the dot only when there is
room in the buffer.


evverx:
https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md#llms

PR #916, opened 2026-05-22
common: fix pointer-before-start UB in drop_incomplete_utf8

dxbjavid:
drop_incomplete_utf8() computes `strchr(c, 0) - 1`, which is undefined
when c is empty. The empty case is reached from
avahi_alternative_service_name(" #N") and avahi_alternative_host_name("-N")
via avahi_strndup(s, 0); fuzz-domain exercises both entry points.

### DIFF
--- a/avahi-common/alternative-test.c
+++ b/avahi-common/alternative-test.c
@@ -91,6 +91,12 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
 
     avahi_alternative_service_name("\xc1\x0a");
 
+    /* These reach drop_incomplete_utf8() with an empty string (via
+       avahi_strndup(s, 0)), which used to form a pointer before the start
+       of the buffer. */
+    avahi_free(avahi_alternative_service_name(" #1"));
+    avahi_free(avahi_alternative_host_name("-2"));
+
     avahi_free(r);
     return 0;
 }
--- a/avahi-common/alternative-test.c
+++ b/avahi-common/alternative-test.c
@@ -22,6 +22,7 @@
 #endif
 
 #include <stdio.h>
+#include <string.h>
 
 #include "alternative.h"
 #include "malloc.h"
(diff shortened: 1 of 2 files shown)


evverx:
As mentioned in another PR one-off fixes with no tests aren't considered especially in cases like this where tests are straightforward. I'll keep it open as a bug report though. Generally it would probably be better to open issues first and mention how issues are discovered along with all the details like the tooling, *san/Valgrind backtraces, the output of static analyzers and their names or anything like that.

I'll also point to https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md again

dxbjavid:
Fair enough on the test. Pushed one to alternative-test.c that runs the two inputs which reach drop_incomplete_utf8() with an empty string (" #1" and "-2", both via avahi_strndup(s, 0)).

On how it turned up: this came from reading the trimming loop rather than a crash. `strchr(c, 0) - 1` forms a pointer before the start of the buffer when c is empty, which is UB per C11 6.5.6/8, but it's never dereferenced and the address doesn't wrap, so *san/valgrind stay quiet and there's no backtrace to attach. The empty case is reachable from the fuzz-domain entry points. I'll open issues first for the next ones.

evverx:
> Pushed one to alternative-test.c that runs the two inputs

The code is already exercised by the fuzz targets so in terms of testing the function the new tests don't add much without assertions making sure that it works as expected.

dxbjavid:
Right, those bare calls were really just leaning on the fuzzers. Added assertions on the actual output now: " #1" -> " #2" on the service path and "-2" -> "-3" on the host path, both of which take the empty-string trim branch.

dxbjavid:
Hi @evverx ,

Just following up on this - now that the tests include assertions for the expected outputs, could you please let me know if there’s anything else needed from my side?

If everything looks good, I would appreciate your feedback on the next steps or merge decision.

Thanks for your time!

dxbjavid:
Closing this pull request due to inactivity

PR #921, opened 2026-06-02
dns: reject oversize rdata in avahi_rdata_parse

dxbjavid:
avahi_rdata_parse() passes its size_t size straight into parse_rdata(), whose rdlength argument is a uint16_t, so an rdata of 65536 bytes or more is truncated mod 65536 and parsed as a different (often empty) record while still returning success. It's reachable from a local client through EntryGroup.AddRecord, whose byte array goes here unchecked. Reject sizes above AVAHI_DNS_RDATA_MAX before the narrowing, matching the assert already in avahi_rdata_serialize().

### DIFF
--- a/avahi-core/dns-test.c
+++ b/avahi-core/dns-test.c
@@ -170,5 +170,22 @@ int main(AVAHI_GCC_UNUSED int argc, AVAHI_GCC_UNUSED char *argv[]) {
     avahi_record_unref(r);
     avahi_record_unref(r2);
 
+    /* rdata larger than the 16 bit rdlength field must be rejected, not
+     * truncated mod 65536 and parsed as a shorter record */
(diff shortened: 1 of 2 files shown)


evverx:
The test should be moved to a separate commit and the commit message should say why the commit is needed. Currently it just says "dns: reject oversize rdata in avahi_rdata_parse" and it makes it look scarier than it actually is. As mentioned in your other PRs it should be mentioned how issues are discovered. In this case I'm guessing it was the `implicit-conversion` UBSan check so the backtrace should be attached as well.

dxbjavid:
Done. The test is in its own commit now, and I reworded the fix commit message to spell out what actually happens: it's local-only via AddRecord, the parse still returns success but the record ends up empty/unrelated rather than crashing, so it's not as scary as the old subject made it sound. And yeah, it was the implicit-conversion UBSan check, backtrace is in the commit message.

PR #922, opened 2026-06-02
pass unsigned char to ctype functions in name parsers

dxbjavid:
Noticed `avahi_unescape_label` and `avahi_domain_hash` hand a plain `char` to isdigit/tolower. On signed-char platforms a high-bit byte from a UTF-8 name goes negative, which is undefined per C11 7.4 and indexes the ctype table out of bounds. Those names come straight off the wire, and domain_hash backs the cache/entry hashmap keys. Cast to unsigned char here and at the matching sites in avahi_alternative_* and avahi_str{up,down}. Added a domain-test case covering a backslash-escaped high byte and a UTF-8 hash.

### DIFF
--- a/avahi-common/alternative.c
+++ b/avahi-common/alternative.c
@@ -69,7 +69,7 @@ char *avahi_alternative_host_name(const char *s) {
         e++;
 
         for (p = e; *p; p++)
-            if (!isdigit(*p)) {
+            if (!isdigit((unsigned char) *p)) {
                 e = NULL;
                 break;
             }
@@ -151,7 +151,7 @@ char *avahi_alternative_service_name(const char *s) {
             e = n + 2;
 
         for (p = e; *p; p++)
-            if (!isdigit(*p)) {
+            if (!isdigit((unsigned char) *p)) {
                 e = NULL;
                 break;
             }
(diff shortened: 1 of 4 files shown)


evverx:
Can you report issues instead of opening PRs (as you already said you were going to do in one of your PRs)? Looks like the context gets reset somewhere along the way so the review is pointless (and I'm not particularly interested in repeating the same things over and over again).

I'll point to https://github.com/avahi/avahi/blob/master/.github/CONTRIBUTING.md once again. It doesn't look like there is a human in the loop.

dxbjavid:
Thanks for the feedback.

You are right that I said I would report issues first, and I haven't been consistent about that. That is on me.

For the issues I have been looking at, I manually reviewed the code paths, reproduced the behavior where possible, and prepared patches based on that analysis. And, I can understand why my recent submissions may have caused some doubt about the amount of review and validation I was doing before opening a PR.

I did read the contributing guidelines, but I haven't followed the workflow the project expects. Going forward, I will report suspected issues first, discuss them, and only prepare patches after there's agreement that the issue is valid and worth addressing.

I appreciate the candid feedback. My goal is to help improve the project, not create unnecessary review work for maintainers, so I will adjust my approach accordingly.

dxbjavid:
Hi, just checking whether this PR should remain open for review, or if you'd prefer that I close it and open an issue first according to the project's contribution workflow. Thanks.

evverx:
It's actually tracked in https://github.com/avahi/avahi/security/advisories/GHSA-8pmv-23p5-92mw. Closing.

PR #924, opened 2026-06-04
avahi-browse: pass unsigned char to isprint in make_printable

dxbjavid:
make_printable in avahi-browse hands a plain char to isprint, so a service name carrying a high-bit UTF-8 byte (any accented name, since valid service names only need to be valid UTF-8) passes isprint a negative value on signed-char builds, which is undefined behaviour per the ctype contract and an out-of-bounds table read on some libcs. I hit it reading back through the ctype call sites looking for ones that take untrusted bytes. Casting to unsigned char before the call keeps the output identical for normal characters while dropping the UB.


evverx:
https://github.com/avahi/avahi/pull/922

PR #936, opened 2026-06-19
Use clock_gettime when POSIX monotonic clock is available

### LINKED ISSUE #359: avahi-autoipd can fail if system time is changed
We have a system that is running both avahi-autoipd and an NTP client. After a reboot, where the system time survives in the hardware RTC and is "almost" correct after the restart, the NTP client would try to avoid a discontinuous time jump but this would cause problems in our application on the system which needs to be synchronized exactly to the NTP server right from the start. Therefore we decide to reset the system time to a defined timepoint well in the past in order to avoid that behavior and force the NTP client to hard set the system time when synchronizing with the NTP server.

We could observe that the system occasionally fails to obtain a link local address. After more investigation we could find out that the reset of the the system time during startup can take place when avahi-autoipd is still in its PROBING state and that the reset of the system time distorts the internal timeout calculation. See the attached [syslog_autoipd.txt](https://github.com/lathiat/avahi/files/7183900/syslog_autoipd.txt) for the resulting debug output in that case.

Looking at the implementation of sleeping in avahi-autoipd the culprit is clearly the usage of `gettimeofday()` for the computation of the sleep intervals. I will probably come up with a pull request that fixes the issue (just in avahi-autoipd) by using `clock_gettime()` (together with the monotonic clock) instead soon.

dobey:
This is a simplified version of #96 which avoides the API breaking.

Fixes #359

### DIFF
--- a/avahi-autoipd/main.c
+++ b/avahi-autoipd/main.c
@@ -785,7 +785,7 @@ int is_ll_address(uint32_t addr) {
 static struct timeval *elapse_time(struct timeval *tv, unsigned msec, unsigned jitter) {
     assert(tv);
 
-    gettimeofday(tv, NULL);
+    avahi_now(tv);
 
     if (msec)
         avahi_timeval_add(tv, (AvahiUsec) msec*1000);
--- a/avahi-common/simple-watch.c
+++ b/avahi-common/simple-watch.c
@@ -486,13 +486,12 @@ int avahi_simple_poll_prepare(AvahiSimplePoll *s, int timeout) {
         if (next_timeout->expiry.tv_sec == 0 &&
             next_timeout->expiry.tv_usec == 0) {
 
-            /* Just a shortcut so that we don't need to call gettimeofday() */
+            /* Just a shortcut so that we don't need to call avahi_now() */
             timeout = 0;
             goto finish;
         }
 
-        gettimeofday(&now, NULL);
-        usec = avahi_timeval_diff(&next_timeout->expiry, &now);
+        usec = avahi_timeval_diff(&next_timeout->expiry, avahi_now(&now));
 
         if (usec <= 0) {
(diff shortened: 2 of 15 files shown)


dobey:
@evverx @pemensik Could we perhaps get this version of the monotonic clock patch merged? And maybe a new release of avahi, since it's been ~6.5 years since the last actual release? :)

evverx:
Thank you for the PR

As far as I can remember the API was changed deliberately so that the consumers of the API could switch properly. With this PR it's possible to silently end up using gettimeofday and clock_gettime at the same time unless I'm missing something. Can you expand on why you think it's safe to do that?

dobey:
> Thank you for the PR
> 
> As far as I can remember the API was changed deliberately so that the consumers of the API could switch properly. With this PR it's possible to silently end up using gettimeofday and clock_gettime at the same time unless I'm missing something. Can you expand on why you think it's safe to do that?

I'm not sure why @JPEWdev thought API should be changed to have an `AvahiTimeVal` struct instead of using `struct timeval`, as it is 1:1 compatible. Perhaps as a means to force devs consuming Avahi API to not use `gettimeofday`, though I don't think that would really work as the structs are 1:1 matches.

That said, with this change, both would not be used internally, unless someone else introduced a new call to `gettimeofday` somewhere else in the code, instead of using `avahi_now` as it provides a compile-time guarantee through the `#if` switching. And we can't really guarantee what consumers of Avahi may do in their own code.

evverx:
> I don't think that would really work as the structs are 1:1 matches

Why not? They would fail to compile against the new headers.

> we can't really guarantee what consumers of Avahi may do in their own code

I think it's safe to assume that they can use gettimeofday because that's what avahi has always worked with.

Has this PR been tested against any existing clients?

PR #934, opened 2026-06-18
core: add NSEC record type support to fix mDNS reflection

dustin-johnson:
Fixes #379.

## Summary

Avahi parses NSEC (type 47) as opaque rdata. But an NSEC record starts with a domain name (the Next Domain Name) that mDNS name-compresses (RFC 6762 §18.14), so storing it as raw bytes freezes the source packet's compression pointer into the blob. On reflection that pointer is re-emitted at a different offset, where it points at unrelated data — the record decodes to garbage and Wireshark flags the reflected packet as malformed.

This makes NSEC a first-class type like PTR/CNAME/SRV/NS: the Next Domain Name is decompressed on parse and re-compressed for the destination packet on serialize, with the type bitmap kept as raw bytes. Parse also rejects any NSEC that couldn't be re-serialized within the 16-bit rdata limit, so anything we accept can always be reflected.

I stumbled upon #379 as I noticed Avahi's reflected mDNS traffic was being flagged as malformed by Wireshark when I was investigating why my Brother printer wasn't being discovered correctly. I can confirm that this PR fixes both the printer discovery issue and Wireshark's malformed check.

## Tests

`avahi-core/dns-test.c` adds: rdata round-trip, empty bitmap, a compression regression (reflect through two differently-laid-out packets), and oversize rejection. Clean under `-Wall -Wextra -Werror`; passes under ASan/UBSan.

No fuzzing changes needed as a coverage build confirms the existing `fuzz-consume-record`/`fuzz-packet` targets reach the new code naturally (~16-18M hits/target in a cold 30-min run, no crashes).

## Disclosure

Per CONTRIBUTING.md: developed with substantial LLM assistance (Claude); I've reviewed and verified all of it, am accountable for it, and can answer questions in review.

PR #937, opened 2026-06-20
core: forward NSEC records without a public struct change

dustin-johnson:
An alternative fix for https://github.com/avahi/avahi/issues/379 without committing to a public interface change as #934 did.

## Summary
Avahi parses NSEC (type 47) as opaque rdata. But an NSEC record starts with a domain name (the Next Domain Name) that mDNS name-compresses (RFC 6762 §18.14), so storing it as raw bytes freezes the source packet's compression pointer into the blob. On reflection that pointer is re-emitted at a different offset, where it points at unrelated data, the record decodes to garbage and Wireshark flags the reflected packet as malformed.

Rather than add a dedicated NSEC member to the public AvahiRecord union (rr.h is an installed header, so that would be an ABI commitment before NSEC is supported properly), store the decompressed Next Domain Name as a string followed by the type bitmap in the existing generic blob. append_rdata() re-compresses the name for its destination packet. The record is therefore always forwardable, with no stale pointers, and no public layout is committed.

I stumbled upon https://github.com/avahi/avahi/issues/379 as I noticed Avahi's reflected mDNS traffic was being flagged as malformed by Wireshark when I was investigating why my Brother printer wasn't being discovered correctly. I can confirm that this PR fixes both the printer discovery issue and Wireshark's malformed check.

## Tests
avahi-core/dns-test.c adds: rdata round-trip, empty bitmap, a compression regression, and oversize rejection. Clean under -Wall -Wextra -Werror; passes under ASan/UBSan.

No fuzzing changes needed as a coverage build confirms the existing fuzz-consume-record/fuzz-packet targets reach the new code naturally (~16-18M hits/target in a cold 30-min run, no crashes).

## Disclosure
Per CONTRIBUTING.md: developed with substantial LLM assistance (Claude); I've reviewed and verified all of it, am accountable for it, and can answer questions in review.

### DIFF
--- a/avahi-common/defs.h
+++ b/avahi-common/defs.h
@@ -335,7 +335,8 @@ enum {
     AVAHI_DNS_TYPE_MX = 0x0F,
     AVAHI_DNS_TYPE_TXT = 0x10,
     AVAHI_DNS_TYPE_AAAA = 0x1C,
-    AVAHI_DNS_TYPE_SRV = 0x21
+    AVAHI_DNS_TYPE_SRV = 0x21,
+    AVAHI_DNS_TYPE_NSEC = 0x2F  /**< NSEC record (RFC 4034); used by mDNS per RFC 6762 */
 };
 
 /** DNS record classes, see RFC 1035 */
--- a/avahi-core/dns-test.c
+++ b/avahi-core/dns-test.c
@@ -22,6 +22,8 @@
 #endif
 
 #include <assert.h>
+#include <stddef.h>
+#include <string.h>
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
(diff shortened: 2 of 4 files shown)


evverx:
Sorry but it's a kludge. It makes sense to come up with a way to support it properly and it's discussed in https://github.com/avahi/avahi/pull/934.

dustin-johnson:
I thought this was a pretty good compromise. Remember that the status quo is that Avahi is emitting broken mDNS packets (the compressed hostname points to a random location in the reflected packet). I get that this approach isn't an ideal long-term state, but it does make forward progress on the actual function of valid mDNS reflection while keeping the state internal until more fully fledged NSEC handling is worked on.

If this approach isn't acceptable, then perhaps you can provide guidance on #934.

evverx:
> I thought this was a pretty good compromise

If it covers your use cases it can be used downstream by analogy with the other patch where NSEC RRs are dropped.

> If this approach isn't acceptable, then perhaps you can provide guidance on https://github.com/avahi/avahi/pull/934.

As mentioned there it needs some thinking and if I start doing that I don't need external PRs to guide Claude in the right direction.

dustin-johnson:
> If it covers your use cases it can be used downstream by analogy with the other patch where NSEC RRs are dropped.

Yes, my usage is covered. I'm trying to contribute back so everyone else who has a subtly broken network due to the existing sloppy reflection behavior will also benefit.

> if I start doing that I don't need external PRs to guide Claude in the right direction.

Oh, I see. I didn't realize that the tone of this project was exclusionary. Sorry for the trouble, I'll drop off now.

evverx:
> I didn't realize that the tone of this project was exclusionary

It's not exclusionary. CONTRIBUTING mentions the following rule though

and I think it's fair when code is cheap.

dustin-johnson:
Yes, I agree that contributions should not be low effort, and I spent many hours (not Claude, me) getting this far. Avahi is a large codebase with poor onramps for folks to get up to speed. I'm just a guy trying to fix his router and contribute back to the community. Even with some powerful tooling, I apparently blundered enough to be dismissed. Nevertheless, I will stop distracting you in the hopes that you'll have the time to fix this 4+ year old bug. :)

evverx:
> blundered enough to be dismissed

The other PR is still open and code suitable to be put in the public header and kept there for eternity will be considered.

> this 4+ year old bug

That bug has always been there so it's much older. In places where it really matters it has already been fixed one way or another. I think the vendor of your router implemented its own reflector FWIW.

PR #947, opened 2026-06-27
FreeBSD CI: Chase the Python 3.12 switch

### LINKED ISSUE #946: CI fails on FreeBSD 15
It failed in https://github.com/avahi/avahi/pull/945 and https://github.com/avahi/avahi/actions/runs/28298448244/job/83842447222 with
```sh
Updating FreeBSD-ports repository catalogue...
FreeBSD-ports repository is up to date.
Updating FreeBSD-ports-kmods repository catalogue...
FreeBSD-ports-kmods repository is up to date.
All repositories are up to date.
pkg: No packages available to install matching 'py311-pygobject' have been found in the repositories
pkg: No packages available to install matching 'py311-dbus' have been found in the repositories
```

arrowd:
Closes #946

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -137,8 +137,11 @@ case "$1" in
         pkg install -y gettext-runtime gettext-tools gmake intltool \
             gobject-introspection pkgconf expat libdaemon dbus-glib dbus gdbm \
             libevent glib automake libtool libinotify qt5-core qt5-buildtools \
-            gtk3 py311-pygobject py311-dbus py311-gdbm mono git socat \
+            gtk3 python3 mono git socat \
             valgrind dfuzzer check radamsa wget
+        python_version=$(python3 -c 'import platform; print("".join(platform.python_version().split(".")[0:2]))')
+        pkg install -y "py${python_version}-pygobject" "py${python_version}-dbus" \
+            "py${python_version}-gdbm"
         # some deps pull in avahi itself, remove it
         pkg remove -fy avahi-app
         ;;
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -134,6 +134,11 @@ case "$1" in
         install_radamsa
         ;;
     install-build-deps-freebsd)
+        # Use latest package set
+        mkdir -p /usr/local/etc/pkg/repos/
+        cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+        sed -i.bak -e 's|/quarterly|/latest|' /usr/local/etc/pkg/repos/FreeBSD.conf
+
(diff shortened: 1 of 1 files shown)


arrowd:
For some reason I can't remove the draft status from this PR.

evverx:
It says 'Author has reached the open pull request limit' and I suspect it has something to do with a feature GitHub rolled out recently. Let me check where the limit comes from.

In the meantime looks like it fails on FreeBSD 14 now.

evverx:
The "ready for review" button should start working hopefully.

evverx:
I pushed https://github.com/avahi/avahi/commit/fc2d3fe503eb112838e33e25a6cd9157829d69f9 to unblock the CI for now. I guess the right fix would be to detect the python version automatically and append it to the package names instead of hardcoding `py311/py312` or something like that. I'm not sure yet.

arrowd:
> The "ready for review" button should start working hopefully.

It is!

arrowd:
Sigh, all right, globs are bad idea - they match multiple Python versions. We'll have to resort to explictly specifying the version. Luckily, Python major bumps only happen once per several years in FreeBSD.

evverx:
> globs are bad idea - they match multiple Python versions

Agreed. To judge from "The process will require 1 GiB more space." it pulls a lot of things.

> We'll have to resort to explictly specifying the version

As far as I understand it wouldn't work with the two latest FreeBSD releases so it should be hardcoded twice and it isn't obvious what should be hardcoded there (it took me a while to find https://github.com/freebsd/freebsd-ports/commit/135741412e9fe53270e9078d14f71066a2aecaa6).

Would maybe something like
```sh
pkg install python3
python_version=$(python3 --version | perl -ne 'print "$1$2" if /(\d+)\.(\d+)/')
pkg install "py${python_version}-pygobject" "py${python_version}-dbus" "py${python_version}-gdbm"
```
work?

arrowd:
> As far as I understand it wouldn't work with the two latest FreeBSD releases

It will, packages are the same for all supported versions.

> python_version=$(python3 --version | perl -ne 'print "$1$2" if /(\d+)\.(\d+)/')

Yes, this seems even better.

evverx:
> It will, packages are the same for all supported versions.

As far as I understand it doesn't work with the latest package set at least to judge from https://github.com/avahi/avahi/actions/runs/28299839269/job/83846131891 and even though I temporarily stopped using it in https://github.com/avahi/avahi/commit/fc2d3fe503eb112838e33e25a6cd9157829d69f9 I'd bring it back to always get the latest version of Valgrind.

Can you also revert https://github.com/avahi/avahi/commit/fc2d3fe503eb112838e33e25a6cd9157829d69f9? I think with the automatic detection it should be good to go.

evverx:
Looks like OpenScanHub got stuck. Merging.

PR #952, opened 2026-07-03
fix one-byte full_name overflow in service_resolver_callback

Samin061:
service_resolver_callback fills full_name[AVAHI_DOMAIN_NAME_MAX] via avahi_service_name_join and then unconditionally strcat's a trailing dot, so a resolved service whose joined name reaches sizeof-1 writes one byte past the buffer. Give full_name room for the extra dot and keep the join bounded to AVAHI_DOMAIN_NAME_MAX so the terminator stays in range.


Samin061:
gentle ping

Samin061:
any update?

PR #969, opened 2026-08-23
avahi-core: don't assert on already-dead group in avahi_s_entry_group_check_probed()

spockfish:
## Summary

`avahi_s_entry_group_free()` marks a group dead (`g->dead = 1`) and defers actually freeing its entries via `schedule_cleanup()`. If the whole server is torn down (`avahi_server_free()`, which frees all entries directly) before that deferred cleanup runs, the teardown path `avahi_entry_free()` -> `avahi_goodbye_entry()` -> `remove_announcer()` reaches `avahi_s_entry_group_check_probed()` for an entry whose group is already marked dead, hitting `assert(!g->dead)` and aborting the whole daemon.

The function already handles `g->dead` turning true *during* its own execution (the second `if (g->dead) return;` after `avahi_s_entry_group_change_state()`); this makes it handle `g->dead` already being true on entry the same way, instead of asserting.

## Correction from the original version of this PR

The original version of this PR proposed a fix in `simple_protocol_restart_queries()` instead. That was wrong, and I want to flag it explicitly rather than quietly swap it: I'd found the assertion string `simple-protocol.c:574: simple_protocol_restart_queries: Assertion `server' failed.` via `strings` on the coredump and mistook its presence for proof it was the assertion that fired -- but that string (like every `assert()` message in the binary) is compiled in unconditionally, so finding it in the coredump's memory doesn't mean it's the one that aborted. Building a real sysroot with the device's actual (aarch64) shared libraries and running `gdb bt full` against them, as requested, gave the real answer below, which is what this PR now fixes.

## Backtrace

```
Program terminated with signal SIGABRT, Aborted.
#0  0x0000007f8cf1817c in ?? () from libc.so.6
#1  0x0000007f8cecda18 in raise () from libc.so.6
#2  0x0000007f8ceb9c48 in abort () from libc.so.6
#3  0x0000007f8cec74ac in ?? () from libc.so.6
#4  0x000000557be36704 in ?? ()
#5  0x0000007f8d3373a0 in avahi_s_entry_group_check_probed () from libavahi-core.so.7
#6  0x0000007f8d337874 in ?? () from libavahi-core.so.7          <- static remove_announcer(), per source
#7  0x0000007f8d3381f4 in avahi_goodbye_entry () from libavahi-core.so.7
#8  0x0000007f8d32e838 in avahi_entry_free () from libavahi-core.so.7
#9  0x0000007f8d32c718 in avahi_server_free () from libavahi-core.so.7
#10 0x000000557be330c8 in ?? ()
#11 0x0000007f8ceba29c in ?? () from libc.so.6
#12 0x0000007f8ceba378 in __libc_start_main () from libc.so.6
```

(frame #6 has no exported symbol since `remove_announcer()` is `static`; matching it to source by the call graph -- it's the only caller of `avahi_s_entry_group_check_probed()` reachable from `avahi_goodbye_entry()`.)

## Test plan

- [x] Compiles (`gcc -fsyntax-only`)
- [x] Matches the existing `if (g->dead) return;` pattern already present later in the same function
- [ ] Not yet reproduced in a minimal standalone test case -- found via a real-world coredump; happy to add a regression test if you can point to the preferred harness (thanks for the `smoke-tests.sh` pointer, will look at whether this is expressible there)


evverx:
Thank you for the PR.

Can you attach the backtrace?

> happy to add a regression test if maintainers can point to the preferred harness for this

In https://github.com/avahi/avahi/blob/master/.github/workflows/smoke-tests.sh avahi-daemon is fully run and then coredumps and Valgrind/ASan/UBSan backtraces are collected so it can be tested there probably.

spockfish:
Updated the PR (title, description, and the commit itself) with a correction -- see the "Correction from the original version of this PR" section above for the full explanation. Short version: the first version of this PR was based on a misread coredump (I mistook a compiled-in assert string for proof of which assert actually fired). Built a proper aarch64 sysroot with the device's real shared libraries and got a real symbolized backtrace this time, which points to a different (and I believe correctly diagnosed) bug in `avahi_s_entry_group_check_probed()`. Apologies for the noise, and thanks for pushing for the backtrace -- it caught a real mistake.

evverx:
Can you open an issue instead and attach as many details as possible there? I don't think it makes sense to fix issues without figuring out what exactly is going on first. Closing.

PR #974, opened 2026-09-01
core: handle IPv4 prefix length 32 correctly

msekletar:


evverx:
Thank you for the PR.

Can you update the commit message and cite the standard
> If the value of the right operand is negative or is
greater than or equal to the width of the promoted left operand, the behavior is undefined

to make it easier to figure out what is being fixed here?

OpenScanHub is unhappy: https://openscanhub.fedoraproject.org/task/137943/log/added.html (and Coverity is likely to be unhappy too with those assertions) and the NDEBUG builds complain
```
2026-09-01T11:27:01.2051976Z iface-test.c:21:40: warning: unused parameter 'interface_address' [-Wunused-parameter]
2026-09-01T11:27:01.2053570Z    21 | static int address_on_link(const char *interface_address, unsigned prefix_len, const char *address) {
2026-09-01T11:27:01.2067304Z       |                                        ^
2026-09-01T11:27:01.2067849Z iface-test.c:21:92: warning: unused parameter 'address' [-Wunused-parameter]
2026-09-01T11:27:01.2068567Z    21 | static int address_on_link(const char *interface_address, unsigned prefix_len, const char *address) {
2026-09-01T11:27:01.2069153Z       | 
...
2026-09-01T11:27:01.2780630Z iface-test.c:38:57: warning: variable 'candidate' is uninitialized when passed as a const pointer argument here [-Wuninitialized-const-pointer]
2026-09-01T11:27:01.2787328Z    38 |     return avahi_interface_address_on_link(&interface, &candidate);
2026-09-01T11:27:01.2787920Z       |                                                         ^~~~~~~~~
2026-09-01T11:27:01.2788419Z iface-test.c:21:12: warning: unused function 'address_on_link' [-Wunused-function]
2026-09-01T11:27:01.2789102Z    21 | static int address_on_link(const char *interface_address, unsigned prefix_len, const char *address) {
2026-09-01T11:27:01.2789606Z       |            ^~~~~~~~~~~~~~~
2026-09-01T11:27:01.2789959Z 4 warnings generated.
```

Ideally there should be something like `assert_se` that wouldn't be compiled out but until then can you replace the assertions with
```c
r = ...;
assert(r);
```
to get the code at least to run when the assertions are compiled out.

evverx:
It was sent to Coverity and it's happy too. I can no longer see the UBSan backtrace
```
iface.c:794:35: runtime error: shift exponent 32 is too large for 32-bit type 'uint32_t' (aka 'unsigned int')
    #0 0x7f2c38018f39 in avahi_interface_address_on_link /home/vagrant/avahi/avahi-core/iface.c:794:35
    #1 0x7f2c3802f17b in dispatch_packet /home/vagrant/avahi/avahi-core/server.c:993:14
    #2 0x7f2c3802e315 in mcast_socket_event /home/vagrant/avahi/avahi-core/server.c:1093:13
    #3 0x7f2c383b37d9 in avahi_simple_poll_dispatch /home/vagrant/avahi/avahi-common/simple-watch.c:585:13
    #4 0x7f2c383b3d8a in avahi_simple_poll_iterate /home/vagrant/avahi/avahi-common/simple-watch.c:605:14
    #5 0x0000004fb92c in run_server /home/vagrant/avahi/avahi-daemon/main.c:1279:18
    #6 0x0000004ecd94 in main /home/vagrant/avahi/avahi-daemon/main.c:1708:13
    #7 0x7f2c37bba680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #8 0x7f2c37bba797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #9 0x000000401444 in _start (/home/vagrant/avahi/avahi-daemon/.libs/avahi-daemon+0x401444) (BuildId: 4d30d9c7d7b6c22283f3524b0b4a550e942a6dc6)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior iface.c:794:35
```
either so the CI is green. I'll take a closer look later.

evverx:
Is there any reason `>>` was replaced with `<<`? `<<` still triggers UBSan
```sh
./avahi-core/iface-test 
iface.c:797:54: runtime error: left shift of 4294967295 by 8 places cannot be represented in type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior iface.c:797:54
```
The "integer" check is off by default and flags suspicious patterns that aren't necessarily undefined behaviour but it helps to catch stuff so it would probably be better if there were fewer places where it was triggered.

msekletar:
No particular reason, I found this a bit less convoluted than shifting -1 and doing negation. In any case, I thought that only shifting by 32 is undefined. This is now explicitly prevented so we should be fine. Am I missing something?

evverx:
avahi is built with the `integer` check from time to time to catch things like 5e5f6538b3611da58998319d5d0231e6250bb47b for example so if there were fewer places where it was triggered it would make it easier to go over those warnings.

siteshwar:
@msekletar Thanks for the fixes based on OpenScanHub report. I have added your name to the [list of contributors](https://github.com/openscanhub/openscanhub/wiki/Upstream-and-Fedora-Contributors) who have made fixes based on OpenScanHub reports.

@evverx If there are other contributors making fixes based on these reports, please let me know their names and I would give them due credit. Thanks!

evverx:
@siteshwar all the new PRs are expected to be green on OpenScanHub among other things so in a way all the contributors should make sure that there are no true positives and fix them when they pop up. PRs like https://github.com/avahi/avahi/pull/863 fix those things too but in that particular case I don't think it was prompted by OpenScanHub. It just confirmed that the warnings were gone.

siteshwar:
> @siteshwar all the new PRs are expected to be green on OpenScanHub among other things so in a way all the contributors should make sure that there are no true positives and fix them when they pop up. PRs like #863 fix those things too but in that particular case I don't think it was prompted by OpenScanHub. It just confirmed that the warnings were gone.

I have updated the list to mention all Avahi contributors. Thanks!

PR #979, opened 2026-09-15
Explicit enum value for AvahiPublishFlags instead of 0

N1coc4colA:
Hello, I've been using Avahi recently. I noticed that AvahiPublishFlags is set to 0 frequently, but no enum value explicitly represent it. I was using C++ which, emitted a warning, and I also noticed that it was difficult to find why put 0 instead of an enum value.
For this reason, I propose to use an explicit enum value that I called AVAHI_PUBLISH_DEFAULT as the docs stated that it is what to use by default unless some particular behaviour is requested.
I believe that this makes things clearer, and can help in bindings as well as foreign languages which may complain.
It is only a refactoring.


evverx:
Thank you for the PR.

This approach was proposed in https://github.com/avahi/avahi/pull/836 and rejected.

Issue #389, opened 2022-07-28
mdns hostnames (other devices on the network) incrementing at vastly faster rate since installing avahi (and scrypted).


diothar:
Issue: Hostnames (of other devices on the network) are incrementing (to combat perceived conflicts), but incremeting substantially faster than they used to.  Breaks some browsing and home automation that I would like to do.  Many devices on the network, but 2 iPhones and an appletv consistently increment their hostnamesmuch more often than the rest of the Bonjour devices. Example: my HomePod mini is on like name-3.local, but my iPhone is on name-659.local and I had reset it.  It has climbed much higher.  Restarting my router helps.   Renaming hostnames on iPhone helps. But always temporary.

Unfortunately, I lost my history that had my latest set of logs and troubleshooting, but I was able to save my ssh session, which captures my avahi conf, the contents of avahi-browse, iOS log regarding networking (mdns debugging profile enabled- the rest of the logs stripped out because it's a personal device) and output of syslog and dmesg from device avahi is installed on. 

 **Logs are just under 50 mb and I can't upload them.  Do you have a server or share I can drop them to?**

If you can tell me the exact logs you need, I will reproduce all of it and generate appropriate logs from the same time (right now I'm just hobbling to get you some data). 

Hardware:
Qotom box running untangle firewall
3x tplink eas 245v3 Wi-Fi access points
3x tplink smart switches
2x vlans, 1x management
Avahi and scrypted running on a Linux Mint mini pc.
172.25.10.1 is router
172.25.10.132 is avahi host
When trying to test IoT network, I put my iphone on 172.25.69.0/24 subnet, and my testing is printing to my brother laser printer.  I consider it a success when I can print to my laser printer from the 69 subnet (it's on 172.25.10.0/24)
joshua_iphone, katie_iphone are the biggest offenders for the hostname increments.  
Some devices increment hostname a lot. Others not a much, but ideally, I'd love for it to stay with its original hostname.

Rebooting router temporarily fixes. 
Renaming hostname from impacted device temporarily fixes. 
Do not have any devices that share the same hostname that conflict with each other.
Firewall logs don't show much, but do show the traffic on 5353
Wireshark does show malformatted packets with some mdns traffic.  Tons of packets, more than I would expect. 

Let me know a specific test and what logs you'd like and I will send them your way. 
I suspect devices going to sleep and waking up with a stale avahi cache entry can trigger renames.
I suspect the malformed packets could trigger renames.
[...]

PR #982, opened 2026-09-20
core: follow the RFC 6762 section 8 conflict rules for stale probes and responses

### LINKED ISSUE #117: Spurious name conflicts
Hi, hope this is the right place for reporting issues with avahi-daemon as the readme on my system points to the bugtracker on freedesktop.org that does not list avahi as a bug report target.

I am experiencing spurious name conflicts on various systems, all of which have a common trait in having two interfaces, one on the local lan, having a static IP address and the other getting a dhcp address from somewhere (typically an ADSL router).

What happens is the following. Suppose that the host is called "foo". Initially, it is correctly advertised as foo.local. After some time the name conflict occurs and the host starts being advertised as foo-2.local, foo-3.local, etc., even if it is certainly the sole host named foo on the network. In practice there is a spurious name conflict with the host itself, probably due to some race in avahi.  The unfortunate result is that no other system cannot find "foo" no more on the network, since they look for foo.local.

I see the issue on a couple of debian jessie systems (avahi version 0.6.31); on a raspbian jessie system (same); and on an openwrt chaos calmer system (avahi version 0.6.31 again).

I see a lot of reports for this same issue (or possibly something similar) on many distro bugtrackers, applications bugtrackers and question sites:

- https://bugs.launchpad.net/ubuntu/+source/avahi/+bug/831022
- https://serverfault.com/questions/714917/host-name-conflict-with-avahi-daemon-on-ubuntu-14-04-lts
- https://github.com/foosel/netconnectd/issues/8
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=771806
- https://lists.freedesktop.org/archives/avahi/2011-March/001995.html

I wonder if there is something misconfigured on my systems (and in this case some hit at diagnosing would be appreciated) or if this is an issue (possibly a race) with the avahi daemon.

Even if this cannot be fixed rapidly, I'd like to suggest an interim point release of avahi with an option to disable the name conflict analysis when he/she is absolutely sure that it won't be needed on his/her network.

### LINKED ISSUE #654: avahi-core/server: incoming_probe: Erroneous conflict detection if TXT data is changed

stevenschlansker:
Fixes #654. Fixes #117.

This PR is built with help from Claude Fable 5.1, but a human (me! :wave: )
experienced the problem (For years at this point),
drove the test-driven fix, and read everything before submitting.

Given the high visibility of the defect, its long-standing nature, and the
conformance-test-driven approach here, I hope the contribution
clears the "drive by LLM slop" bar. 

Status: Fix submitted (here), unit and conformance tests pass -
now running patched builds locally to verify longer term behavior.

I'm a developer but not a mDNS nor IP-theory expert. Please find Claude's report following:

## The defects

My Fedora desktop, luminol, has wired and wireless interfaces on one
LAN and renames itself up to luminol-50. avahi conflicts with itself
in three ways. I reproduced each in a network namespace with two
interfaces on one link and confirmed it with a packet capture.

1. A lost simultaneous probe tiebreak renamed the host at once.
   RFC 6762 section 8.2 says the loser "defers to the winning host by
   waiting one second, and then begins probing for this record again",
   because the winning probe can be stale, "possibly even stale probe
   packets sent moments ago by this host itself". That is the #117
   sequence: a global IPv6 address arrives, avahi withdraws the
   link-local AAAA record while a probe for it is still queued, the
   probe comes back milliseconds later, and the new record loses the
   tiebreak against the withdrawn one.

2. Conflicting responses received before the first probe was sent
   counted as conflicts. Section 8.1 says they "MUST be silently
   ignored". Two back-to-back responses were enough for a rename: the
   first returned the record to probing, the second withdrew it.

3. Withdrawal announced the withdrawn group. When the last probing
   announcer of a group goes away, `remove_announcer()` completes the
   group and announces its waiting records with a full TTL. That
   announcement then conflicted with the record the second interface
   was probing.

mDNSResponder implements 1 and 2: `ResolveSimultaneousProbe()` and the
"Ignoring response received before we even began probing" branch of
`mDNSCoreReceiveResponse()` in mDNSCore/mDNS.c.

## The changes

One commit per defect, a test, and a doc:

- `incoming_probe()` defers every record of the name that probes on
  the interface for one second, with one shared deadline, instead of
  withdrawing the losing RRset. Deferring only that RRset is not
  enough: a probe asks for the name with type ANY, so a sibling record
  (HINFO, the AAAA record published over IPv4) that keeps probing draws
  a denial for the name before the deferred record re-probed, and the
  denial is ignored as stale. BCT fails PROBING for that variant as it
  does for the immediate rename. Records under other names are not
  touched, so a peer that streams later probes for one name cannot
  stall the rest of the registration. This follows #692 by
  @ceggers-arri, co-author of the commit. The deferral keeps no state
  on the entry, which answers the review question there: a real winner
  answers the new probes and `handle_conflict()` renames as before.
- `handle_conflict()` ignores a conflicting response for an entry that
  has not sent a probe in its current probing cycle, and does not
  cache or reflect the record, as a record that caused a withdrawal
  was not.
- `avahi_goodbye_entry()` removes announcers without completing the
  group. Only `avahi_goodbye_interface()` still completes it, because
  the announcers on the other interfaces continue.
- `conflict-test` runs a scripted mDNS peer in the same process and
  event loop as the server under test, on loopback. Five scenarios: a
  real conflict renames the host; an unanswered later probe does not
  (8.2); two back-to-back conflicting responses do not (8.1); a
  conflicting response that withdraws a group with two addresses, one
  still probing and one waiting for it, announces nothing; a
  lost tiebreak plus one denial of the re-probe renames exactly once
  with no probe during the pause (BCT's PROBING subtest, with a second
  unique record under the host name). On Linux it enters a private
  user and network namespace when available, otherwise it skips with
  77 if the mDNS port is taken or multicast does not loop back. It
  runs in `make check` and in the smoke tests under `unshare -n`.
- `docs/BONJOUR-CONFORMANCE-TEST` describes how to run Apple's test
  against avahi-daemon. The setup took far longer than the test.

## Results

`conflict-test` (18 checks) on the tree with only the test commit:
stale-probe, stale-response and tiebreak-denial rename the host, and
withdraw does not reach its precondition because the stale probe
already withdrew the group. With commits 1 and 2 only, the withdrawal
completes the group and announces the waiting records, and both
withdraw checks fail. With the three fixes all 18 pass. Clean under
valgrind with the CI options. Two interfaces on one link
with address churn, 18 cycles: master renames once per run in three
runs, this branch never.

On the distribution package: Fedora 44 ships avahi 0.9~rc2. A trigger
that reproduces the #117 sequence on the real interface, a
global-scope IPv6 address added and replaced by a lower one within 20
to 250 ms, then removed after three seconds, renamed the host 13 times
in 120 cycles with the stock package (hostname became hostname-14). The
same package rebuilt with the five code commits of this series and
its two follow-ups, and nothing else, renamed it 0 times in 120
cycles on the same machine. Cycling the link itself did not trigger
the stock daemon in 60 cycles on this box, because NetworkManager
brings the global address up 1.5 seconds after the link-local one,
after probing has finished.

A further 60 cycles with `--debug` logged "Deferring probing" 10 times
and no conflict: each of those is a lost tiebreak against the daemon's own
stale probe that the section 8.2 wait absorbed.

**Bonjour Conformance Test 1.5.4, v0.9-rc5 base, wired Mac, device on
IPv4 link-local, run on this code**:

- rc5 fails PROBING in both runs: it renames on the conflicting probe
  instead of deferring (defect 1).
- rc5 with these commits passes INITIAL PROBING, PROBING including the
  rate limit after the fifteenth rename, and WINNING SIMULTANEOUS
  PROBES, then times out in SRV PROBING because the server reaches the
  running state 77 seconds after the host name settles.
- With the two follow-up PRs (registration hold-off, running state
  during withdrawal) it passes every subtest of the multicast DNS test
  through MANUAL NAME CHANGE and HOT-PLUGGING. Both are pre-existing
  bugs that only show after about fifteen renames, hence separate PRs.

The daemon log of that run also shows the section 8.1 ignore working
on the wire: during HOT-PLUGGING two denials arrived before the
deferred re-probe went out and were logged as ignored, and the rename
followed the denial of the re-probe.

Two harness notes. BCT tracks every SRV probe from the device, and
D-Bus clients such as Samba do not rename on a collision, so the runs
used `disable-user-service-publishing=yes` and two static services.
Those need literal names: a static service with `%h` never renames on
conflict because `static-services.c` recomputes the name from the
wildcard after a collision. That is #166, open since 2018, and the
reason the SRV subtest in #2 fails with the stock service files. File
Sharing on the test Mac must be off, or BCT's reply timing subtest
queries the device for the Mac's own `_smb._tcp`; the new doc says so.

## The reports in #117

I read all 261 comments. The issue, as its reporter and @adriancable
scoped it, is avahi renaming itself with no other owner of the name.
Nearly every first-hand report with detail is that, and every one of
those fits the three defects:

- Withdraw of the link-local AAAA record when a global address arrives,
  with a probe still queued: @lathiat's own diagnosis in 2017
  (issuecomment-302849130, issuecomment-309958496) and most reports
  that name IPv6, ULA or prefix changes.
- Link events that withdraw and re-register every record: cable pulls,
  wake from suspend, DHCP renewals, and @mvduin's isolated VLAN with a
  single interface, no reflector, and switch-port toggling, where
  `--debug` shows the conflict against "a very recently transmitted
  probe" (issuecomment-2329365216, issuecomment-2331334421). That is a
  single-homed host hitting defect 1 through multicast loopback.
- An IPv4 fallback address replaced by a DHCP lease at boot
  (issuecomment-842288060), the IPv4 form of the same churn.
- Disabling `publish-aaaa-on-ipv4` and `publish-a-on-ipv6` stops it
  (issuecomment-1651848898). Those options create the sibling
  announcers that the deferral of the whole name is there for.

Three groups of comments are different problems and are already
called out as such in the thread:

- A second mDNS stack on the same host: a container with its own
  responder (issuecomment-1955907861), KDE Connect on port 5353
  (issuecomment-2132064268), librespot (issuecomment-2417815258),
  systemd-resolved's responder (issuecomment-2666840702). Those are
  real conflicts and the rename is correct.
- avahi's reflector renaming other devices, mostly Apple ones. #263
  fixed one cause; the rest was moved to #389
  (issuecomment-1197400042).
- One-offs: two boards with the same name, a retracted report, a
  machine-id reset.

So this PR closes #117 as scoped. If a report of the first kind
survives on a release with these commits, it deserves a new issue with
a `--debug` log, which now names the conflicting record.

## Previous attempts

- #309 treated every packet from a local address as our own. That
  hides a second mDNS stack on the host, and testers lost resolution
  of their own name.
- The 117-spurious-name-conflicts branch kept dead entries for one
  second. Both conflict paths skip dead entries, so it changed nothing
  there, which matches the reports of "rarer, not gone". It also failed
  BCT "PROBING: PROBE DENIALS" (@Altair-Bueno in #117).
- #692 implemented the 8.2 deferral and stalled on the request for a
  test. This PR adopts it and adds the test.

## Why IPv6 and dual-homed hosts

- IPv6 addresses churn. avahi publishes the link-local AAAA record
  only until a global address exists
  (`avahi_interface_address_is_relevant()`), so it withdraws a record
  it just started probing. Temporary address deprecation repeats that.
  IPv4 addresses rarely change after boot.
- The stale link-local record always wins the tiebreak against the
  new global one: fe80::/10 sorts after 2000::/3 and fc00::/7. Between
  two IPv4 addresses it is a coin toss.
- Only IPv6 echoes cross to the other interface. On Linux a frame from
  one of the host's own IPv4 addresses arriving on another interface
  is a martian source (`fib_validate_source()`, `accept_local` 0).
  IPv6 has no such check.

Multicast loopback echoes both families and the daemon handles those
echoes like foreign packets (`from_local_iface` only gates
reflection), so the first two points apply to single-homed hosts too.

Per CONTRIBUTING.md: an LLM agent did much of the investigation, the
prototype and the test under my direction. I reviewed every line and
can answer for each change.


evverx:
Thank you for the PR. Can you expand on how you fixed the CI failure and why you think the fix makes sense?

stevenschlansker:
Hi, thank you for taking a look. The failure was in the two `-DNDEBUG` jobs, and the NetBSD job. The NetBSD job looks like a flake - the job failed in a smoke test in unrelated territory and disappeared on the next push. 

The `-DNDEBUG` jobs were an incorrectly written test packet builder. The problematic code was of the form:
```
assert(avahi_dns_packet_append_record(p, r, 1, 0));
avahi_dns_packet_set_field(p, AVAHI_DNS_FIELD_ANCOUNT, 1);
```
So, when assertions are disabled, the call is compiled out and the fixture does not append records to packets as intended. Therefore, anything that "must" occur has to move out of `assert`.
The fix is a test helper `must` instead of assert, with similar "error and exit" behavior, but not conditioned on build flags.

evverx:
Why do you think it should be bolted on top of that test only?

stevenschlansker:
It looks like other tests have the same hazard, and they skip their checks and pass anyway. I can update the PR to `#undef NDEBUG` and continue to use assert, if that's your preference, or did you mean something else?

evverx:
> It looks like other tests have the same hazard, and they skip their checks and pass anyway

That's correct and that bug in the testsuite should be fixed properly in a separate PR (and take things like Coverity with its "side effects in the assertion" warnings into account along the way).

stevenschlansker:
Ok, I can try to come up with some shared helper and fix the other test cases and Coverity in a separate PR.

evverx:
Let's start with that. That's how the latest PRs usually go. Claude generates a convoluted bogus test and cut corners to get them to pass and when stuff actually gets fixed and backed up by actual tests PRs clear the "drive by LLM slop" bar. If you ever fix that bug please attach links to GH Actions jobs where it was actually tested by analogy with, say, https://github.com/avahi/avahi/pull/976. It's worth mentioning (as mentioned in https://github.com/avahi/avahi/pull/971) the idea behind the LLM rule is to make sure that PRs are tested and reviewed before they are submitted and avoid the scenario where cheaply generated code is offloaded onto the avahi project.

stevenschlansker:
I enabled GitHub Actions on my fork. Latest test results at https://github.com/stevenschlansker/avahi/actions/runs/35563356738

The existing test code does not fit the `assert` contract well.  This means we should move test code away from using it entirely, rather than use the wrong tool. I promoted `must` to a new `test-util.h` in #985 and included justification for the approach.

Since the new PR introduces a new test utility that is used here, I rebased on it and re-drafted this PR until we handle 985 and will come back after it lands, or you recommend a different approach.  Unfortunately the same NetBSD failure again showed in CI - I am not 100% sure if it is something I introduced or a pre-existing flake. I will have to diagnose that further tomorrow.
 
Thank you again for your review.

evverx:
I haven't reviewed it yet. I just took a quick look to see whether there is any human oversight or whether Claude just generates stuff to get the CI green. By the way the PR limit was set for a reason so there is no need to open a gazillion draft PRs and stack them on top of each other (among other things it delays the CI results and PRs aren't reviewed until the basic checks pass). If your agent runs into an issue please open an issue and describe it there precisely along with packet captures, the steps to reproduce it and stuff like that.

PR #985, opened 2026-09-21
tests: check with must() instead of assert()

stevenschlansker:
In review of #982, I tripped on an existing hazard in the codebase: the build matrix includes some cells that define `-DNDEBUG` which disables `assert()` entirely. This caused my first CI runs to fail because they included test cases that had assertions with side effects - compiled out under Ubuntu. From the ensuing discussion, I found that this is not a new problem, see 3730966 and 9b8e798, and per request will try to structurally fix this hazard.

A second commit fixes `ASSERT_NOT_NULL` in two `avahi-compat-howl` tests, which declared `_r` but used `r`. This is a latent compile error or incorrect reference; I fixed it since I edited the same line.

`CONTRIBUTING.md` updated to clarify the `assert` vs `must` distinction.

This PR was built with Claude Fable 5.1, but I selected the approach.

## Designs

Rejected:

* `#undef NDEBUG` - this feels like fighting the build system, and leaves another hazard with regards to its order with `#include <assert.h>`
* Move all side-effects out of assert - this makes the test pass, but leaves a vacuous pass - with `assert` disarmed, many checks are skipped

These two points indicate that we are fighting the contract of `assert`, which is meant to be used as a runtime check, not a unit test harness. Therefore, we should introduce a new abstraction. I chose `must` over `assert_se` : the "se" (side effect) is only one class of issue here, even a side-effect free assertion is removed under `NDEBUG`.
If you prefer to mimic other projects with `assert_se` despite the concern that it invites asking "is this a side effect?", let me know and I can rework this PR.

## Implementation

`must` macro: check the parameter.  if zero: flush stdout, log an error, and abort.
Similar to `assert` but always checks its argument.
I did not see an existing home for test utilities, so I added `avahi-common/test-util.h` as a `noinst_HEADERS` with a macro.
Since `must` is not `assert` and can not be compiled out, this should fix the Coverity warnings.

Every `assert` in test code should be a `must` so that it enforces in every build matrix cell.
Six test programs (e.g. `hashmap-test`) have no asserts and are unchanged. The fuzz harnesses use `assert` but I called that out of scope for now as they do not run with `NDEBUG` today.

## Testing

Tested on
* Develop on Fedora 44 `make check` with and without `-DNDEBUG`
* `make check distcheck install` with `--disable-qt3 --disable-qt4 --disable-qt5 --disable-mono --disable-monodoc` on Rocky EL10.2
* verified mutation testing by flipping a test check: branch fails `make check` even with `-DNDEBUG`, continues to pass on `master`
* `make dist` includes the header, `make install` does not

Removes 30 pre-existing CI warnings about unused variables when run with `-DNDEBUG`.

GHActions: https://github.com/stevenschlansker/avahi/actions/runs/35745920930

Note: While developing, my fork ran into a second CI failure of the same class as I experienced in #982 in NetBSD tests:
`assertion "found_addr" failed: socket.c:793 avahi_recv_dns_packet_ipv4` 

After it hit twice, I wanted to make very sure I was not introducing a new problem. The investigation showed that I was not, and I filed a fix for an existing race condition in https://github.com/avahi/avahi/pull/987


avahi-compat-howl/text-test.c
```diff
@@ -21,18 +21,18 @@
...
 
-#include <assert.h>
+#include <avahi-common/test-util.h>
 #include <stdio.h>
 
 #include <avahi-common/gccmacro.h>
 #include "howl.h"
 
-#define ASSERT_SW_OKAY(t) { sw_result _r; _r = (t); assert(_r == SW_OKAY); }
-#define ASSERT_NOT_NULL(t) { const void* _r; r = (t); assert(_r); }
```
stevenschlansker, line 31: Calling out subtle change here: `r` looks like an existing typo. same change in `browse-domain-test.c`
The macro is unused in this file, otherwise it would cause a compile error or mis-reference

evverx:
I'll convert this to draft and undraft https://github.com/avahi/avahi/pull/987 because that's the only PR I'm planning to take a look at and merge in the foreseeable future.

PR #986, opened 2026-09-21
core: set the receive options before binding the sockets

stevenschlansker:
**Withdrawn**
Sorry, I meant to open this PR against my fork to get a GitHub actions run. I will follow this separately once I have the PR body ready. Apologies for the noise.

<details>

avahi_open_socket_ipv4() and avahi_open_socket_ipv6() bound the socket to port 5353 and set IP_PKTINFO / IP_RECVTTL (IPV6_RECVPKTINFO / IPV6_RECVHOPLIMIT) afterwards. On the BSD kernels that order is a race: udp_input() builds the control messages when it queues a datagram to a socket, with the options the socket has at that moment, and queues a multicast datagram to every socket bound to the port (NetBSD sys/netinet/udp_usrreq.c, udp4_realinput() and udp4_sendup() calling ip_savecontrol(); XNU udp_append() does the same). With another responder on the host keeping it in the mDNS group, the new socket receives copies of all mDNS traffic from bind() on, and a datagram queued before the setsockopt() calls carries no control messages. The first recvmsg() on it then aborted the process:

    assertion "found_addr" failed: file "socket.c", line 793, function "avahi_recv_dns_packet_ipv4"
    assertion "found_ttl" failed: file "socket.c", line 916, function "avahi_recv_dns_packet_ipv6"

Both seen in the NetBSD CI job during the cname-test loop of the smoke tests, where each case opens a fresh server while the daemon answers the previous one's goodbyes. Linux builds the control messages at recvmsg() time and never hits this.

Set the options before bind() in all four open functions. Keep the asserts, which is how this surfaced, and add a drop path after them: with NDEBUG the asserts compile out and a datagram without control messages reached dispatch_packet() with an uninitialized destination address, which decides whether the packet counts as multicast and which interface it belongs to.
</details>


stevenschlansker:
Sorry, I meant to open this PR against my fork to get a GitHub actions run. I will follow this separately once I have the PR body ready. Apologies for the noise.

PR #987, opened 2026-09-21
core: set socket receive options before binding

stevenschlansker:
Fixes an intermittent abort of the NetBSD CI job, seen on #982 and #985. Outcome of a failed CI build investigation.

First #982 hit an assertion failure near `found_ttl`, next attempt succeeds. Then, #985 had a CI build fail (in my fork, while preparing the `GHActions:` citation) for an adjacent reason in `found_addr` - next attempt succeeds again. Given that 985 is test-code-changes only and should not have any observable difference (as long as `-DNDEBUG` is not set), this is a strong indication of a pre-existing flake not something new I introduced in my PR stacks.

In the spirit of "getting things right" rather than just hurrying to the destination by hitting "retry tests", I prompted Claude Fable 5.1 to diagnose the flake and try to determine the cause.

Linux and BSD handle socket options differently. `avahi` currently binds the socket first, then sets socket options. In Linux, this is reliable since the implementation acts on this socket option in the `recvmsg` path (late), so the ip info will always apply. In BSD and XNU (Mac), the option is applied when an incoming datagram is queued to be received (early), so the race allows a packet to arrive before the socket option becomes effective.

I briefly tried to find an authoritative citation regarding socket options before vs after bind, but was not able to locate a RFC, manpage, or other document explicitly stating so. So, I had my agent dig into the BSD and Mac (similar lineage) source trees and confirm by building a standalone empirical test that the current ordering is problematic: run an analogous code path rigged by `sleep` to force the race here. More details in Claude's report below.

With avahi's current approach (bind, set opt), 6 of 8 received datagrams lacked the required packet info, the last 2 included it and were the packets that arrived after the option had been set. In the corrected order (set opt, bind), with the same `sleep` rigged test harness, all 6 datagrams correctly receive the required IP info.

Additionally, the current error checking is hazardous: the `assert` is the only condition protecting `found_addr`, `found_iface`, and `found_ttl`, and as we discussed in 985 this is not appropriate for required checks: a `-DNDEBUG` build will skip it, and in this case proceed with uninitialized data into `dispatch_packet` as if it had valid IP info.

Therefore, the fix is two parts:

* Always set socket options before binding so that the ip packet info is reliably set on all systems
* Handle missing-info case always, even though we request it: the kernel decides whether to attach it. Warn and drop the datagram in production builds; keep `assert` in debug builds to guard against regression.

GHActions: https://github.com/stevenschlansker/avahi/actions/runs/35624494929

## Claude's report

<details open>
<summary>Claude's report</summary>

`avahi_open_socket_ipv4()` and `avahi_open_socket_ipv6()` bind the
socket to port 5353 and set the receive options (`IP_PKTINFO`,
`IP_RECVTTL`; `IPV6_RECVPKTINFO`, `IPV6_RECVHOPLIMIT`) afterwards. On
the BSD kernels that order is a race:

- `udp4_sendup()` builds the control messages when it queues a datagram
  to a socket, from the socket's flags at that moment
  (`ip_savecontrol()`, NetBSD `sys/netinet/udp_usrreq.c`). Nothing is
  recomputed at `recvmsg()`.
- `udp4_realinput()` queues a multicast datagram to every socket bound
  to the port, with no group-membership check, as long as the sockets
  have `SO_REUSEPORT` or `SO_REUSEADDR`.
- `setsockopt(IP_PKTINFO)` sets `INP_RECVPKTINFO` (`ip_output.c`), so
  the option avahi uses is the right one once it is set.

With another responder on the host keeping it in the mDNS group, the
new socket receives copies of all mDNS traffic from `bind()` on, and a
datagram queued before the `setsockopt()` calls carries no control
messages. The first `recvmsg()` on it hit an assert and took the
process down:

```
assertion "found_addr" failed: file "socket.c", line 793, function "avahi_recv_dns_packet_ipv4"
assertion "found_ttl" failed: file "socket.c", line 916, function "avahi_recv_dns_packet_ipv6"
```

Both in the `bsd (netbsd, 11.0, gcc)` job, inside the cname-test loop
of the smoke tests, right after `server state: 1`: each case opens a
fresh server while the daemon answers the previous one's goodbyes.
Linux builds the control messages at `recvmsg()` time from the stored
packet and never hits this. macOS (XNU `udp_append()`) has the BSD
behavior.

The change sets the options before `bind()` in all four open
functions, and handles a datagram without the expected control
messages in every build: warn and drop it, with `assert(0)` inside
that branch so a debug build still aborts, which is how this surfaced.
Before, with `NDEBUG` the asserts compiled out and such a datagram
reached `dispatch_packet()` with an uninitialized destination address,
which decides whether the packet counts as multicast and which
interface it belongs to.

Evidence: a standalone program that binds 5353, joins the group, sends
itself six multicast datagrams over three seconds with the options
unset, then sets them and reads. On macOS, whose XNU kernel shares the
BSD design:

```
late: bound; sending 6 multicast datagrams over 3 s before the options are set
datagram  1 from 192.168.42.43    12 bytes  addr cmsg: NO  ttl cmsg: NO
...
datagram  6 from 192.168.42.43    12 bytes  addr cmsg: NO  ttl cmsg: NO
datagram  7 from 192.168.42.116  197 bytes  addr cmsg: yes  ttl cmsg: yes
datagram  8 from 192.168.42.116  197 bytes  addr cmsg: yes  ttl cmsg: yes
late: 6 of 8 datagrams without the requested control messages
early: 0 of 6 datagrams without the requested control messages
```

The six queued before `setsockopt()` have no control messages; the two
that arrived from another host afterwards do. With the options set
before `bind()` ("early") every datagram has them.

The same program ran in the smoke tests of a throwaway branch on my fork (https://github.com/stevenschlansker/avahi/pull/3). Results on the CI kernels, NetBSD being the platform of the two aborts:

| kernel                            | job                                                                                 | late (bind, then options)       | early (options, then bind) |
|-----------------------------------|-------------------------------------------------------------------------------------|---------------------------------|----------------------------|
| NetBSD 11.0, gcc                  | https://github.com/stevenschlansker/avahi/actions/runs/35621499363/job/106405402578 | 6 of 6 without control messages | 0 of 6                     |
| FreeBSD 14.5 and 15.1, three jobs | https://github.com/stevenschlansker/avahi/actions/runs/35615542564              | 6 of 6 without control messages | 0 of 6                     |
| Linux, all jobs                   | same runs                                                                           | all present                     | all present                |

The two aborts this fixes, both bsd (netbsd, 11.0, gcc), both inside the cname-test loop of the smoke tests:

| assertion                                            | job                                                                                                        |
|------------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
| found_ttl, socket.c:916, avahi_recv_dns_packet_ipv6  | https://github.com/avahi/avahi/actions/runs/35530495026/job/106135511500 (#982, second push)                |
| found_addr, socket.c:793, avahi_recv_dns_packet_ipv4 | https://github.com/stevenschlansker/avahi/actions/runs/35563181489/job/106219804941 (#985 head, on my fork) |

The fixed daemon passed the full smoke suite on NetBSD, FreeBSD, OpenBSD and OmniOS in the run linked under GHActions: above.

</details>

## Appendix: where the behavior is implemented

<details>
<summary>Code references: Linux, NetBSD, XNU, Avahi</summary>

Line numbers are from the branch heads read on 2026-09-21 (NetBSD
`netbsd-11`, XNU `main`, Linux `v6.12`, avahi master ef903fb).

### NetBSD: control messages are built when the datagram is queued

`sys/netinet/udp_usrreq.c`, `udp4_sendup()`, lines 462 to 469. The
control messages are built from `inp->inp_flags` as they are at this
moment, then appended to the receive buffer with the datagram:

```c
	if ((n = m_copypacket(m, M_DONTWAIT)) != NULL) {
		if (inp->inp_flags & INP_CONTROLOPTS ||
		    SOOPT_TIMESTAMP(so->so_options)) {
			struct ip *ip = mtod(n, struct ip *);
			ip_savecontrol(inp, &opts, ip, n);
		}

		m_adj(n, off);
		if (sbappendaddr(&so->so_rcv, src, n, opts) == 0) {
```

https://github.com/NetBSD/src/blob/netbsd-11/sys/netinet/udp_usrreq.c#L462-L469

### NetBSD: multicast is queued to every socket bound to the port

`sys/netinet/udp_usrreq.c`, `udp4_realinput()`, lines 503 to 558. No
group-membership check per socket; the walk continues while the
sockets have `SO_REUSEPORT` or `SO_REUSEADDR`:

```c
	if (IN_MULTICAST(dst4->s_addr) ||
	    in_broadcast(*dst4, m_get_rcvif_NOMPSAFE(m))) {
		...
		TAILQ_FOREACH(inp, &udbtable.inpt_queue, inp_queue) {
			if (inp->inp_af != AF_INET)
				continue;
			...
			udp4_sendup(m, off, (struct sockaddr *)src,
			    inp->inp_socket);
			...
			if ((inp->inp_socket->so_options &
			    (SO_REUSEPORT|SO_REUSEADDR)) == 0)
				break;
```

https://github.com/NetBSD/src/blob/netbsd-11/sys/netinet/udp_usrreq.c#L503-L558

### NetBSD: the option avahi sets is the one the receive path reads

`sys/netinet/ip_output.c`, `ip_ctloutput()`, lines 1191 to 1196:
`IP_PKTINFO` with an integer sets `INP_RECVPKTINFO`.

```c
		case IP_PKTINFO:
			error = sockopt_getint(sopt, &optval);
			if (!error) {
				/* Linux compatibility */
				OPTSET(INP_RECVPKTINFO);
				break;
```

`sys/netinet/ip_input.c`, `ip_savecontrol()`, lines 1525 and 1553 to
1558: the flags are read once, and `IP_PKTINFO` is emitted only when
`INP_RECVPKTINFO` is among them.

```c
	int inpflags = inp->inp_flags;
	...
	if (inpflags & INP_RECVPKTINFO) {
		...
		    sizeof(ipi), IP_PKTINFO, IPPROTO_IP);
```

https://github.com/NetBSD/src/blob/netbsd-11/sys/netinet/ip_output.c#L1191-L1196
https://github.com/NetBSD/src/blob/netbsd-11/sys/netinet/ip_input.c#L1525-L1558

### XNU (macOS): the same design

`bsd/netinet/udp_usrreq.c`, `udp_append()`, lines 870 to 901:

```c
udp_append(struct inpcb *last, struct ip *ip, struct mbuf *n, int off,
    struct sockaddr_in *pudp_in, struct udp_in6 *pudp_in6,
	...
			ret = ip_savecontrol(last, &opts, ip, n);
```

https://github.com/apple-oss-distributions/xnu/blob/main/bsd/netinet/udp_usrreq.c#L870-L901

### Linux: control messages are built at recvmsg()

`net/ipv4/udp.c`, `udp_recvmsg()`, line 1902: the control messages
are produced from the stored packet when the application reads it, so
the options in force at read time apply.

```c
		ip_cmsg_recv_offset(msg, sk, skb, sizeof(struct udphdr), off);
```

https://github.com/torvalds/linux/blob/v6.12/net/ipv4/udp.c#L1902

### avahi: the order this PR changes

`avahi-core/socket.c` on master, `avahi_open_socket_ipv4()`: `bind()`
at line 344 to 347, `ipv4_pktinfo(fd)` at line 351.
`avahi_open_socket_ipv6()`: `bind()` at line 414 to 417,
`ipv6_pktinfo(fd)` at line 421. The asserts on the receive side:
`assert(found_addr)` at line 793, `assert(found_iface)` and
`assert(found_ttl)` at lines 915 and 916. `dest` in
`avahi_recv_dns_packet_ipv4()` is written only inside the control
message loop (lines 742 to 782), so with `NDEBUG` a datagram without
control messages reaches `dispatch_packet()` with it uninitialized.

https://github.com/avahi/avahi/blob/ef903fb/avahi-core/socket.c#L344-L351
https://github.com/avahi/avahi/blob/ef903fb/avahi-core/socket.c#L793
https://github.com/avahi/avahi/blob/ef903fb/avahi-core/socket.c#L915-L916

</details>


evverx:
Thank you for the PR.

I can see that Claude tested it on FreeBSD and NetBSD (and analyzed the code of NetBSD only for some reason). What's about OpenBSD and illumos? I saw some CI runs where avahi-daemon failed to start on illumos. There were no logs so it's hard to tell what happened there but it looks like it can be related to this.

Please don't fiddle with the assertions and leave them as is. NDEBUG builds are a separate issue and kludges Claude keeps adding here and there don't help much. Other than that assuming the bug is really fixed it shouldn't matter whether those if clauses are present or not.

evverx:
> In Linux, this is reliable

It isn't correct. On Linux it doesn't crash but the ifindex can be 0 so the packets get thrown out down the line effectively.

I reproduced actual crashes on NetBSD and OpenBSD but I'm not sure what the impact on FreeBSD and illumos yet. It needs looking into so that it can be added to the commit message.

stevenschlansker:
Ok, thank you for the feedback. I'll do additional testing and get back to you.

